PR #170, opened 2018-09-04
Cleanup Project

danieldaeschle:
Removing .idea and *.iml files and extend gitignore file with gitignore.io presets


danieldaeschle:
If you want i can contribute more. I want to create an note taking app too, but i was working on the UI only at the moment. Not useable yet.

PR #249, opened 2018-12-01
fix #248: crossorigin attribute is needed when including with type="module"

### LINKED ISSUE #248: CORS issues
I'm trying to setup trilium behind an [oauth proxy](https://github.com/thomseddon/traefik-forward-auth), and the application stays blank due to security issues:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://accounts.google.com/o/oauth2/auth? [...] %2Fjavascripts%2Fsetup.js. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).[Learn More]
Module source URI is not allowed in this document: “https:// [...] /javascripts/setup.js”.
```

I believe this is due to the way `setup.js` is imported:
```
<script src="/javascripts/setup.js" type="module"></script>
```

I don't know much about this, but after reading [this blog](https://jakearchibald.com/2017/es-modules-in-browsers/) I believe a `crossorigin` attribute is required `type="module"`. Otherwise the browser tries to load the JS file without the cookie and is redirect to the auth domain.

nicoulaj:
As discussed in #248.

Tested on Linux Mozilla Firefox 63.0.3 and Chromium 70.0.3538.110.

### DIFF
--- a/src/views/index.ejs
+++ b/src/views/index.ejs
@@ -236,7 +236,7 @@
 
     <link href="/stylesheets/style.css" rel="stylesheet">
 
-    <script src="/javascripts/services/bootstrap.js" type="module"></script>
+    <script src="/javascripts/services/bootstrap.js" crossorigin type="module"></script>
 
     <link rel="stylesheet" type="text/css" href="/libraries/jam/css/jam.min.css">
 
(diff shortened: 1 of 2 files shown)


nicoulaj:
OK, seems like everything is working fine now.

PR #289, opened 2019-01-08
use electron-builder for dev install

perissology:
This will build deps as required for electron as well as the os.

This fixes the missing binding error for sqlite3 when trying to run the
electron app after a fresh npm i


tvortsa:
i anyway get this error with sqlite3 (((

Issue #356, opened 2019-01-17
Webhook to submit notes


christophersw:
It would be awesome if the Trilium Server supported a simple Webhook mechanism for note submission. 

I would immediately leverage this to send notes to Trilium from Slack, from IFTTT, from JIRA, etc.

zadam:
Hmm, Trilium actually does have API to do this, but it's currently not documented. It is used in [trilium-sender](https://github.com/zadam/trilium-sender) which is a simple Android app from which you can create notes and send pictures to Trilium.

christophersw:
Looking at that code - I think the main change here is support for an "api key" or other token that would allow a robot to login.  (Otherwise Slack/IFTT/Other isn't going to be able to authenticate).

zadam:
Yes - the flow in the android app is that first user needs to login using credentials using which the android app gets long lived token which is then used for authentication against the server installation.

But I guess for e.g. Slack it would make more sense to generate the token in Trilium and then manually set it in the slack plugin, right?

christophersw:
I dug a little deeper into Slack and IFTTT - just as the two services that I can personally imagine using the most.... There are probably myriad other apps and integrations that could be imagined. 

Slack's integration is not as easy as I had hoped. Slack allows you to easily create "slash commands" that call a specific URL, but the format and content of the POST is fixed by Slack. [See this...](https://api.slack.com/slash-commands#app_command_handling). So it looks like for Slack we would need a dedicated endpoint. 

It looks from first glace like that's more or less how github handles their webhooks as well - [See this](https://developer.github.com/webhooks/configuring/#writing-the-server)

IFTTT allows you to "make a web request" after a trigger, and it is a bit easier to take advantage of than Slack. IFTTT allow you to set the URL, the HTTP call type (GET, POST), as well as edit the message body to fit your needs. Unfortunately, they don't allow you to set headers, only the response body. So I guess the only real limitation here would be the API Token would need to be in the body text... Not really ideal, but maybe workable.

christophersw:
Completely agree on implementing a per-service API... That's just whack-a-mole thinking that could lead to terrible bloat-ware. 

**I love the idea of the catch-all + scripting.** That really leverages the awesome capabilities of the Trilium's scripting! It's also really flexible, so you could implement all kinds of interactivity and interaction beyond simply posting new notes. You leave it entirely up to the user how to deal with the specific response bodies and content - so the main repo remains clear of all that integration-specific code.

zadam:
[...]

PR #368, opened 2019-01-21
New Theme: Rainbow in the Dark

Abourass:
While working on the math plugin for CKEditor I created the following theme, figured I'd open a PR, and it can be merged if you want, if not, no worries feel free to close. 

Sample Theme


zadam:
Hello, that looks interesting :-) I want to include some more built in color themes but using only the "exposed interface" (essentially defined variables) because if a theme uses CSS selectors then it will be too difficult to maintain when HTML structure will change in the future.

But I also plan to add some easier way to develop and use custom themes - e.g. CSS note with `@appTheme` label would automatically show up in the theme select box in options. This might then be provided as an exported note in .tar archive.

So instead of merging this, what about creating a separate repo and I can link to it from [wiki page on themes](https://github.com/zadam/trilium/wiki/Themes)?

Abourass:
Something like [this?](https://github.com/Abourass/RainbowInTheDarkTril/blob/master/theme.css)
Not sure if the variable declaration will work still, and I know I can't import font via @import within the @appTheme file, so codeLens wont use the font I'd declared. 
Maybe something like you did for font size with custom fonts? 
Because I must admit I love using Raleway
Example Code Image

gerroon:
How do I apply this in docker?

zadam:
Yes, even though I don't think that providing modified builds is sustainable in the future. 

As I understand the main reason for this is to provide extra fonts? I think it might be possible to include a generic functionality where you would mark a note (in this case file note with uploaded) with a label which would expose it on the HTTP level so you could link to it from CSS. This is actually planned for 0.29 and comes originally from #356.

Anyway I included a link to your repo to this page: https://github.com/zadam/trilium/wiki/Themes

Abourass:
@zadam I agree. 

I will remove the modified builds shortly, while keeping the installation instructions for the @appCss attribute method, and the replacing files then build on their own method (until v0.29 is released, then I will modify it too only be the @appCss option). 

It is indeed to provide a single extra font. The solution you just outlined seems perfect. I would think that would also allow custom icons which is exciting. I apologize if my request (for fonts) is a tad tedious. 

I will close this issue now. Thanks.

PR #377, opened 2019-01-23
add confirm type change dialog when note not empty

### LINKED ISSUE #259: Chaging note type which has content should produce warning
Something like;

> It is not recommended to change note type when note content is not empty. Do you want to continue?

Related to #258

flurmbo:
Fixes #259. Let me know what you think!


zadam:
Thank you for the contribution!

PR #602, opened 2019-07-29
fix note height in distraction-free-mode

mohamedhagag:
to fix note height in distraction free mode this should be 100% or totally remove height style.


mohamedhagag:
sorry this will remove subnote buttons.

PR #831, opened 2020-01-18
add hotkey to copy contents with line breaks, fixes #349

### LINKED ISSUE #349: Copy and Paste doesnt preserve line breaks
Not sure if this was meant to be fixed in #51 but if I copy multiple lines of bash script in Chrome on Mac (using the web client) and then paste them into Terminal, the line breaks are not honoured.

Heniker:
Probably `Alt+Ctrl+C` is not the best hotkey though.
On breaks inserted with `shift+enter` it copies a single break line for each line.
On breaks inserted with just `enter` it copies double break line, which may be unwanted behavior.

Tested in latest version of Chrome and Firefox.

### DIFF
--- a/src/public/javascripts/services/entrypoints.js
+++ b/src/public/javascripts/services/entrypoints.js
@@ -241,7 +241,7 @@ function registerEntrypoints() {
 
         d.showDialog(selectedOrActiveNodes);
     }));
-    
+
     keyboardActionService.setGlobalActionHandler("CreateNoteIntoDayNote", async () => {
         const todayNote = await dateNoteService.getTodayNote();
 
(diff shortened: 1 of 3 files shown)


zadam:
originally I thought better would be `ctrl+shift+c` since (`ctrl+shift+v` is used for paste without formatting) but that's already used for another action

src/public/javascripts/services/entrypoints.js
```diff
@@ -288,6 +288,9 @@ function registerEntrypoints() {
 
         searchNotesService.searchInSubtree(node.data.noteId);
     });
+
+    $('document').on('copy', utils.copySelectionToClipboard)
```
zadam, line 292: what is the intention behind it? It seems to me like it is trying to change default "copy" behavior to this without formatting ... but a) it does not seem to work and b) default behavior should be copy with formatting IMHO ...

Heniker: You're right. I actually can't remember why I did this. This line clearly should not be there.

zadam: OK, no problem, I removed it.

PR #849, opened 2020-02-01
Fix corner case preventing notes from being created before ckeditor is initialized

jasontan056:
If Trilium is opened or reloaded on a note without any content, ckeditor will not be initialized. Attempting to create a note after or into at this point will cause this error:
`tree.js:639 Uncaught (in promise) TypeError: Cannot read property 'getSelectedHtml' of undefined
    at createNote (tree.js:639)
    at tree.js:808`

Fixed by checking if window.cutToNote is defined before using it.

### DIFF
--- a/src/public/javascripts/services/tree.js
+++ b/src/public/javascripts/services/tree.js
@@ -635,7 +635,7 @@ async function createNote(node, parentNoteId, target, extraOptions = {}) {
         extraOptions.saveSelection = false;
     }
 
-    if (extraOptions.saveSelection && window.cutToNote) {
+    if (extraOptions.saveSelection && utils.isCKEditorInitialized()) {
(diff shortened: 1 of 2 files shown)


src/public/javascripts/services/tree.js
```diff
@@ -635,7 +635,7 @@ async function createNote(node, parentNoteId, target, extraOptions = {}) {
         extraOptions.saveSelection = false;
     }
 
-    if (extraOptions.saveSelection) {
+    if (extraOptions.saveSelection && window.cutToNote) {
```
zadam: As it is now this does not express the intention well (the fact it is trying to check that CKEditor is initialized). Could you encapsulate it into a function like `isCKEditorInitialized` which would be more explanatory ...

jasontan056: Good idea. How's this?

zadam:
Thanks!

PR #982, opened 2020-04-25
Parse date in revisions dialog

erdnaxeli:
Hi,

I just started to use this tools which looks AMAZING but I noticed the dates in the revisions dialog are not shown using the correct timezone. This is an attempt to fix it.

Sadly I could not test it because it would not run. I tried on gitpod, and here is what I get (from you repo, with master):
```
> trilium@0.41.5 start-server /workspace/trilium
> TRILIUM_ENV=dev node ./src/www

Generated session secret
Generated sourceId=6a5PpyRWp0oy
App HTTP server starting up at port 8080
{
  "appVersion": "0.41.5",
  "dbVersion": 158,
  "syncVersion": 14,
  "buildDate": "2020-04-20T22:40:02+02:00",
  "buildRevision": "a86177bb597c752fbc96a24d4be7ab5ae6c0344d",
  "dataDirectory": "/home/gitpod/trilium-data",
  "clipperProtocolVersion": "1.0"
}
DB not initialized, please visit setup page - http://[your-server-host]:8080 to see instructions on how to initialize Trilium.
{ Error: Router not found for request /app-dist/setup.js
    at app.use (/workspace/trilium/src/app.js:84:17)
    at Layer.handle [as handle_request] (/workspace/trilium/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/workspace/trilium/node_modules/express/lib/router/index.js:317:13)
    at /workspace/trilium/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/workspace/trilium/node_modules/express/lib/router/index.js:335:12)
    at next (/workspace/trilium/node_modules/express/lib/router/index.js:275:10)
    at Layer.handle [as handle_request] (/workspace/trilium/node_modules/express/lib/router/layer.js:91:12)
    at trim_prefix (/workspace/trilium/node_modules/express/lib/router/index.js:317:13)
    at /workspace/trilium/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/workspace/trilium/node_modules/express/lib/router/index.js:335:12) status: 404 }
^CCaught interrupt signal. Exiting.
```

### DIFF
--- a/src/public/app/dialogs/note_revisions.js
+++ b/src/public/app/dialogs/note_revisions.js
@@ -42,8 +42,9 @@ async function loadNoteRevisions(noteId, noteRevId) {
     revisionItems = await server.get(`notes/${noteId}/revisions`);
 
     for (const item of revisionItems) {
+        const formattedTime = utils.formatTime(utils.parseDate(item.dateLastEdited));
         $list.append($('<a class="dropdown-item" tabindex="0">')
-            .text(item.dateLastEdited.substr(0, 16) + ` (${item.contentLength} bytes)`)
+            .text(formattedTime + ` (${item.contentLength} bytes)`)
             .attr('data-note-revision-id', item.noteRevisionId));
     }
 


zadam:
Hello @erdnaxeli ,

this is strange since `dateLastEdited` already contains date including local timezone - e.g. `2020-04-24 18:21:58.131+0200` - 18:21:58 is the local time I made the modification while +0200 describes the offset from UTC.

The only problem I can think of is with server instances which run in different time zone than the user - then the server time is used which might be unexpected.

Can you describe your problem in more detail, perhaps also including example?

Thanks.

erdnaxeli:
> Can you describe your problem in more detail, perhaps also including example?

Sure, here is a screenshot:
2020-04-26-105635_1920x1080_scrot

We can see that my timezone is UTC+2 (the result of the date command in the shell), the backend responds with a datetime in UTC (+0000), but the frontend shows the datetime in UTC and don't convert it to my timezone.

It can be noted that the recent changes endpoint also responds with a datetime in UTC, but the frontend shows it in the correct timezone. The only difference I spotted between the two views is the line I added in the revisions dialogs.

zadam:
Ok, thanks, so it looks like the problem is with a server in a different time zone.

I think what should happen is that the client's timezone should be used to record times instead of the server timezone and then this problem would be fixed. Let me think about it ...

zadam:
Hello, I implemented a solution mentioned in my previous comment - to store date time records, code takes client local time instead of server local time. This way we always store the correct local time zone in the database. So now with this different approach, I'm going to close this PR.

This fix will be in 0.42

erdnaxeli:
But what if my client timezone change ? Like if I write a note in Paris, move to New-York and open it again ? IMHO the date should be localized in the client. The timezone used to store it in the database, as long there is a timezone we can localize it in another timezone.

By the way the recent changes view also get as input an datetime in an incorrect timezone (UTC) but manages to show it in the correct timezone of the client.

zadam:
> But what if my client timezone change ? Like if I write a note in Paris, move to New-York and open it again ?

That's exactly my motivation for not wanting to convert local dates in the client. Imagine this scenario:

* I'm in New York, edit a note at 18:00
* I fly to Paris and open the revisions dialog
* Now I see 00:00 - but that does not correspond to my memory since I remember it was afternoon, not midnight when I was writing the note. It's not really useful to know that it was midnight in Paris time since I wasn't in Paris then.

I believe that it's more natural for user to see the original local date time and not "converted to current timezone" local date time.

> By the way the recent changes view also get as input an datetime in an incorrect timezone (UTC) but manages to show it in the correct timezone of the client.

Yes, I changed that as well now with the aforementioned fix. To be clear, your fix (and the former behavior of recent changes) work just fine if you don't switch timezones, but I believe it will product suboptimal behavior when switching timezones ...

erdnaxeli:
> Now I see 00:00 - but that does not correspond to my memory since I remember it was afternoon, not midnight when I was writing the note. It's not really useful to know that it was midnight in Paris time since I wasn't in Paris then. In other words I would still expect to see 18:00 there, independent of my current location.

Well, I think we disagree about what we want :p. I personally totally would not expect that. What is useful is to have a correct ordering of time events. I have to say that's the first time I see anything like the "solution" (not one in my opinion) you've implemented.

Telling the user the revision was done at 18h when the user is in Paris is, well, a lie. What about another example:
* I live in UTC+2, but by mistake my computer is set to UTC+4
* I wrote a note at 18h, in UTC+2, so 20h in UTC+4
* Now I notice the time shown on my laptop is incorrect, and I set timezone correctly to UTC+2
* I change my note, so now I have a new revision at 18h10
* Finally I open the revision list, but surprise! I see a revision at 18h10 and a revision _before_ at 20h!

And I did not even try to find an example with date, but it's not only about hours. We could easily get into a situation where the list shows J+1 / J+2 / J, which looks to be in a wrong order but it's actually just different datetimes with different timezones.

My idea is that a program should always show the real time in the local zone of the user. To get back to your example, the concept of jet lag is, I think, well known. Like I known I can't from Paris at 10 am call someone in New-York and except them to be awake, I known when it was the afternoon there is was not here.

Side-note: I think your "fix" works with SQLite because it just takes the data you give it and store it, but it would fail with a db like PostgreSQL wihch __converts timestamp with timezone to UTC__ before storing them.

If you stick with your solution I think we should at least show the timezone to allow the user to disambiguate corner cases.

Again your project is great and I don't want to tell you what you do, I wrote this with all the benevolence I can :)

sambbaron:
I agree with @erdnaxeli on this. I think the app should show local time in a common timezone. I would have a setting in the app options that allows the user to change the timezone and detect it automatically from the device. This is a feature in all calendaring apps, and I suggest following this pattern.

zadam:
> I live in UTC+2, but by mistake my computer is set to UTC+4
> I wrote a note at 18h, in UTC+2, so 20h in UTC+4
> Now I notice the time shown on my laptop is incorrect, and I set timezone correctly to UTC+2
> I change my note, so now I have a new revision at 18h10
> Finally I open the revision list, but surprise! I see a revision at 18h10 and a revision before at 20h!

This is a result of incorrect setup, so I don't really consider it a valid use case. Garbage In -> Garbage Out principle.

The use case I have in my mind is this: half of my notes have been written on a different continent in a vastly different time zone. With a local date conversion the time would be show that my work notes have been written during the evening or at night which does not make sense. Even worse, the conversion would often result in completely wrong date than when the note has been written. I mean when the note has been written on 26. 4. 2019, then I surely don't want this to show up as 27. 4. 2019 or 25. 4. 2019 depending on where I am currently located.

> it would fail with a db like PostgreSQL wihch converts timestamp with timezone to UTC before storing them.

That's factually incorrect, PostgreSQL has data types which preserve the time zone and data types which do not preserve the time zone. Almost all serious database engines have such support.

> If you stick with your solution I think we should at least show the timezone to allow the user to disambiguate corner cases.

Yes, part of my fix is to add a tooltip to the shortened time which displays the complete value - i.e. local time + time zone offset.

BTW, the fix I committed is mainly about storing correct data - the real local date + time zone offset. If showing the original local date time will prove later to be unsuitable, presentation can simply change to accommodate that since we store the full information. Besides local date time, Trilium also stores the UTC timestamp for easier database sorting for correct sorting.

erdnaxeli:
> This is a result of incorrect setup, so I don't really consider it a valid use case. Garbage In -> Garbage Out principle.

Erf. You really want me to find to time zones close enough together so I have a valid (not incorrect setup) example? :p 

> That's factually incorrect, PostgreSQL has data types which preserve the time zone and data types which do not preserve the time zone. Almost all serious database engines can store dates with timezone info and/or offset.

No, that's not. PostgreSQL does not preserve the _original_ timezone. From [the doc](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-TIMEZONES) :

> All timezone-aware dates and times are stored internally in UTC. They are converted to local time in the zone specified by the TimeZone configuration parameter before being displayed to the client.

The local time here being the one in the server configuration or the one set in the current client connection. So with only a TIMESTAMP WITH TIMEZONE column, it is impossible to know with which timezone was this value inserted. You will need to store the timezone in another column in order to do that.

> With a local date conversion the time would show that my work notes have been written during the evening or at night which does not make sense.

Exactly. And it **totally** makes sens, because that's what happens. Like I know jet lag, I know the night here is not the night there, that's ok.

> I mean when the note has been written on 26. 4. 2019 in e.g. US, then I surely don't want this to show up as 27. 4. 2019 or 25. 4. 2019 depending on where I am currently located.

Well again, we have different point of view :) Your fix will result in an look-like not ordered list, because it will show different times in different time zones side by side. I prefer having the real time in my local timezone.

> BTW, the fix I committed is mainly about storing correct data

The data was already correct, it was just showed in UTC rather than my local timezone.

> Yes, part of my fix is to add a tooltip to the shortened time which displays the complete value - i.e. local time + time zone offset.

That's cool, even if it will not be obvious if a time I see is in my local timezone or not before looking at the tooltip.

I am not sure I understand what setting @sambbaron propose, but maybe we can have a toggle setting like "show times in local timezone" or something like that.

zadam:
> Erf. You really want me to find to time zones close enough together so I have a valid (not incorrect setup) example? :p

Yes, there are probably some cases where this can happen but I consider them to be corner cases since they happen only when crossing timezones (which for most people does not happen too often) and thus won't affect too many notes. And your "body timezone" will be messed up anyway :-)

> The local time here being the one in the server configuration or the one set in the current client connection. So with only a TIMESTAMP WITH TIMEZONE column, it is impossible to know with which timezone was this value inserted. You will need to store the timezone in another column in order to do that.

Oh, ok, I stand corrected. The column name is very misleading then and that's pretty big weakness IMHO. 

> Exactly. And it totally makes sens, because that's what happens. Like I know jet lag, I know the night here is not the night there, that's ok.

If the time is converted to local timezone, then it becomes quite useless since I don't want to calculate in my head the offsets to figure out when did I actually wrote the note to get the context. And again, it gets even worse with dates ...

> Well again, we have different point of view :) Your fix will result in an look-like not ordered list, because it will show different times in different time zones side by side. I prefer having the real time in my local timezone.

Only in corner cases during timezone transitions. I can live with that.

> The data was already correct, it was just showed in UTC rather than my local timezone.

Nope, it lost the valuable information of the original timezone in which the user edited the note. It looked like you were in UTC which was not correct.

> That's cool, even if it will not be obvious if a time I see is in my local timezone or not before looking at the tooltip.

It will be in a time zone you originally used to edit it.

Frankly, I think this discussion is moving in an unproductive direction, so I'd rather pause for a moment. Let's see how this turns out in the real world, and we can revisit the issue later.

erdnaxeli:
> Oh, ok, I stand corrected. The column name is very misleading then and that's pretty big weakness IMHO.

I think the guys at PostgreSQL though a datetime would be presented localized too :D

> Frankly, I think this discussion is moving in an unproductive direction, so I'd rather pause for a moment. Let's see how this turns out in the real world, and we can revisit the issue later.

Yeah I agree on this, we disagree on a product point, not a technical one. The solution of having non-local times and timezone in a tooltip is actually good enought :)

PR #1059, opened 2020-05-30
Add support for Windows/VSCode development

bhank:
This PR uses cross-env to set environment variables in a cross-platform-safe way so that you can run `npm run start-server` on Windows. It also adds a configuration file so you can easily run the equivalent of the `start-server` script in the VSCode debugger.


bhank:
Looks like someone else made the same cross-env change today. I'll open a new PR with just the VSCode change.

PR #1074, opened 2020-06-01
Add import option to preserve line breaks in imported note content

bhank:
The option defaults to disabled, since line breaks are not supposed to be meaningful in markdown. (But they are meaningful in markdown exported from Zim.)


zadam:
Hello, I'm beginning to be concerned with the amount of Zim specific options here. Are more of them coming?

bhank:
Since I made this PR I've found a couple other issues with Zim's exported markdown. I'll close this PR and just write a script to preprocess my markdown before importing.

PR #1110, opened 2020-06-15
Made Docker build running again

piccobit:
- Copy 'package.json' instead 'server-package.json'
- Use 'production' environment

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM node:12.16.3-alpine
 # Create app directory
 WORKDIR /usr/src/app
 
-COPY server-package.json package.json
+COPY package.json package.json
 
 # Install app dependencies
 RUN set -x \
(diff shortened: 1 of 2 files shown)


zadam:
Hi, I don't understand these changes - docker build is running for me OK.

Also why changing npm scripts - those are for development and not for production use.

piccobit:
Hi, apologies for the confusion but I just wanted to test the newest bugfixes and features and didn't find a Dockerfile in your official repository on the Docker hub. Therefore I tried the one here in your GitHub repository and immediately noticed that it didn't work because your repository doesn't contain a file called 'server-package.json' at all. ;-)

So, if the Dockerfile in your repository is only needed for the development, where do I find the one you're using the build the production Docker image?

zadam:
If you want to build docker image, run `bin/build-docker.sh` - that will correctly generate the missing `server-package.json`

Reason for this is that server version doesn't need some package dependencies like electron so we don't include those in the docker build.

piccobit:
OMG, didn't see that!

Thanks for the explanation, I'll close this PR now...

PR #1120, opened 2020-06-20
Add optional support for note title tooltips under note tree widget

shonramamurthy:
This can be useful for previewing note titles which are too long or become hidden beneath the main view area. An example use case where this can be useful is seen below,

image

Currently, this change will be disabled by default and will only be enabled via a setting under the 'Other' menu as below,

image

### DIFF
--- a/db/migrations/0159__add_tree_tooltip.sql
+++ b/db/migrations/0159__add_tree_tooltip.sql
@@ -1,2 +1,0 @@
-INSERT INTO options (name, value, utcDateCreated, utcDateModified, isSynced)
-VALUES ('treeNoteTitleTooltipEnabled', 'false', '2020-06-14T16:24:00.874Z', '2020-06-14T16:24:00.874Z', 0);
--- a/db/migrations/0159__add_tree_tooltip.sql
+++ b/db/migrations/0159__add_tree_tooltip.sql
@@ -1,0 +1,2 @@
+INSERT INTO options (name, value, utcDateCreated, utcDateModified, isSynced)
+VALUES ('treeNoteTitleTooltipEnabled', 'false', '2020-06-14T16:24:00.874Z', '2020-06-14T16:24:00.874Z', 0);
--- a/src/public/app/widgets/note_tree.js
+++ b/src/public/app/widgets/note_tree.js
@@ -1316,3 +1316,45 @@ export default class NoteTreeWidget extends TabAwareWidget {
         noteCreateService.duplicateNote(node.data.noteId, branch.parentNoteId);
     }
 }
+
+export function setupNoteTitleTooltip() {
+    // Source - https://gist.github.com/jtsternberg/c272d7de5b967cec2d3d
+    var is_colliding = function( $div1, $div2 ) {
+        // Div 1 data
+        var d1_offset             = $div1.offset();
+        var d1_height             = $div1.outerHeight( true );
+        var d1_width              = $div1.outerWidth( true );
+        var d1_distance_from_top  = d1_offset.top + d1_height;
+        var d1_distance_from_left = d1_offset.left + d1_width;
+    
+        // Div 2 data
+        var d2_offset             = $div2.offset();
+        var d2_height             = $div2.outerHeight( true );
+        var d2_width              = $div2.outerWidth( true );
+        var d2_distance_from_top  = d2_offset.top + d2_height;
+        var d2_distance_from_left = d2_offset.left + d2_width;
+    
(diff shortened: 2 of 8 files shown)


db/migrations/0159__add_tree_tooltip.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO options (name, value, utcDateCreated, utcDateModified, isSynced)
```
zadam: DB upgrade is not necessary for options - it's enough it's specified in options\_init - it will be created if missing ...

src/public/app/widgets/note_tree.js
```diff
@@ -279,6 +279,7 @@ export default class NoteTreeWidget extends TabAwareWidget {
             source: treeData,
             scrollParent: this.$tree,
             minExpandLevel: 2, // root can't be collapsed
+            tooltip: (options.get('treeNoteTitleTooltipEnabled') === 'true'),
```
zadam: you can just use `options.is('treeNoteTitleTooltipEnabled')`

src/services/app_info.js
```diff
@@ -4,7 +4,7 @@ const build = require('./build');
 const packageJson = require('../../package');
 const {TRILIUM_DATA_DIR} = require('./data_dir');
 
-const APP_DB_VERSION = 158;
+const APP_DB_VERSION = 159;
```
zadam: As mentioned above, DB migration is not necessary here.

shonramamurthy:
> 
> 
> Hi, this looks good although I'm a bit skeptical about features which are disabled by default.
> 
> Ideally the feature would be always active, but in case where the whole item fits into view the tooltip would not be shown. However that would be probably quite complex :-/ (probably would require IntersectionObserver and some integration with fancytree)

Thank you @zadam for the feedback.

Yes, you are correct. Enabling the tooltip depending on how much of the title is actually shown will be quite complex. This was one of my motivations of having this optional atleast at the initial stage.

db/migrations/0159__add_tree_tooltip.sql
shonramamurthy: Thank you for the feedback @zadam 
Sure I will remove it 👍

src/public/app/widgets/note_tree.js
shonramamurthy: Sure, will do 👍

src/services/app_info.js
shonramamurthy: Understood, will update the same 👍

zadam:
Hi, I have alternative proposal - what about adding title only to an icon - this would not be distracting in normal cases so could be activated always and also somewhat discoverable. I just tried it and it's two lines of code ...

shonramamurthy:
> 
> 
> Hi, I have alternative proposal - what about adding title only to an icon - this would not be distracting in normal cases so could be activated always and also somewhat discoverable. I just tried it and it's two lines of code ...
> 
> 

Hi @zadam, thank you for the suggestion. Adding the tooltip to display based on the icon definitely improves our situation with regards to intersection detection. However I am leaning towards your original suggestion of displaying them based on visibility as this makes this feature more apparent to use.

This branch is now updated with changes which reflect a more intuitive solution. A demonstration of this is below,

set-1

As seen above, the tooltip now displays only based on note-title visibility. This change also becomes notably simpler than originally anticipated as we are not relying explicitly on _IntersectionObserver_. For achieving this we now bypass the Fancytree and rely on the DOM to detect element overlap and then decide on displaying tooltips for note-title elements

I am now currently looking at a way to place this event handler at an appropriate place. Let me know what you think.

zadam:
Hi, this looks promising! I would put the handler into the `note_tree.js` which is a component handling the note tree behavior.

shonramamurthy:
> 
> 
> Hi, this looks promising! I would put the handler into the `note_tree.js` which is a component handling the note tree behavior.

Thank you for the feedback @zadam! 

The latest changes on this branch now contains the note-title tooltip handler definition inside `note_tree.js`. The registration takes place under `side_pane_toggles.js` as we need to be able to detect collision after the `center-pane` element has been initialized. With this we now have the behaviour which I shared earlier under https://github.com/zadam/trilium/pull/1120#issuecomment-647117620

Let me know what you think.

zadam:
@shonramamurthy hi, I merged and refactored the code a bit. Instead of detecting collision with center pane, the code now detects whether the tree node is visually enclosed into the parent tree element. The advantage is that it does not rely on the position or even existence of center element.

shonramamurthy:
Thank you @zadam for your time 😀

PR #1135, opened 2020-06-28
WIP: New note type that supports drawing with pen/stylus and hand written notes: `canvas-note` 

thfrei:
Goal: New note type, that allows drawing on an infinite canvas with pen/stylus like the Surface Pen. The UI should look quite minimalistic with a couple of brushes and an eraser. Shapes that are drawn should be selectable, so that they can be rearranged. The canvas should be ininfinte, like in OneNote. It might also be possible to add images and other goodies (since this is supported via fabric.js)

Technologies:
* Usage of fabric.js to handle different elements on the canvas, have an infinite canvas and also allow to use images and other.

Working: store and retrieve hand drawn sketch, basic drawing, some brushes
ToDo: Pressure support, advanced brushes, minimalistic, unobtrusive user interface, infinite canvas, image support

Demo:
canvas-note


zadam:
Hi, some points/questions:

*   what is the storage format? Something specific to fabric or SVG?
*   what about editing history - by that I mean simple CTRL-Z to undo/redo actions - is that perhaps supported out of the box by fabric or would that have to be added explicitly?
*   I understand that you primarily develop this to support writing handwritten notes with a stylus but I sort of see this more generally and would like to add things like text support, drawing boxes and straight lines, adding links to notes - this should also satisfy free-form diagramming. This is not required for first versions but perhaps good to keep in mind.
    *   In the demo above you use the "circle" brush - I understand this is just a demo but given the focus on note taking I don't think this will be the most useful brush :)
*   Probably clear point but in the end there shouldn't be a "save" button, instead standard save using `spacedUpdate` should be used.

thfrei:
Thanks for your feedback.

> what is the storage format? Something specific to fabric or SVG?

Storage format is JSON. One can export it to SVG (the other way round does not seem to be so clear, however), which would be nice I guess, to reference the drawing as an image inside another note? Maybe dual save? JSON for note, SVG for referencing?
Some info on serialization here: http://fabricjs.com/fabric-intro-part-3#serialization

>what about editing history - by that I mean simple CTRL-Z to undo/redo actions - is that perhaps supported out of the box by fabric or would that have to be added explicitly?

Quite straight forward, see commit https://github.com/zadam/trilium/pull/1135/commits/b1d5470f7249b3caefa03765e7c4122b9dc2f245

>I understand that you primarily develop this to support writing handwritten notes with a stylus but I sort of see this more generally and would like to add things like text support, drawing boxes and straight lines, adding links to notes - this should also satisfy free-form diagramming. This is not required for first versions but perhaps good to keep in mind.

I think fabric.js is  quite a sloid foundation here. It allows for all sorts of shapes and texts.

> *  In the demo above you use the "circle" brush - I understand this is just a demo but given the focus on note taking I don't think this will be the most useful brush :)

Totally :-P

> Probably clear point but in the end there shouldn't be a "save" button, instead standard save using spacedUpdate should be used.

Yes. However, I wonder how the performance will be, since the JSON can become quite large quite easily. Maybe some debouncing after finish drawing...

My next issues:
* Canvas in full size of available space
* Drag canvas, move around, infinite canvas

thfrei:
canvas-note-update

thfrei:
BTW: The undo redo with the separate Array is in my opinion even better than classic undo/redo, since it allows to redo a change, after you made another change. Like redoing the big blue strokes, or purple circle-strokes, after the small blue strokes.

zadam:
> Storage format is JSON. One can export it to SVG (the other way round does not seem to be so clear, however), which would be nice I guess, to reference the drawing as an image inside another note? Maybe dual save? JSON for note, SVG for referencing?

Yeah, I sort of assume that fabric uses JSON since it needs to store some metadata to work better. SVG could be generated even on the fly for the export and possibly also for including into other notes (if that's not too slow). Storing SVG on the side as well would have some benefits as well (performance, mitigates potential bit rot) but then I'm not sure how to approach storing 2 "contents" for one note.

> Quite straight forward, see commit [b1d5470](https://github.com/zadam/trilium/commit/b1d5470f7249b3caefa03765e7c4122b9dc2f245)

Hmm, but that is not really true undo/redo - it assumes undo is always removing object and redo always adding it back. It can be the other way round or you could just e.g. undo object rotation.

My other thought regarding this was if undo/redo information is stored in the JSON - that it would "survive" e.g. restarts. But I assume not and for that we still have note revisions ...

> Yes. However, I wonder how the performance will be, since the JSON can become quite large quite easily. Maybe some debouncing after finish drawing...

Yes, this is actually controlled by that `spacedUpdate` component, the name is trying to suggest that the updates are "spaced" in time.

cidious:
@zadam is there any chance to have this PR merged? The feature looks promising.

thfrei:
It's still work in progress.

Even though it might not look like it, I try to work on every free minute on it. Albeit, free minutes are very sparse :-(

I plan to have something in the next month or so. In order to have quicker iterations for the drawing functionality, I created another project which solely focuses on the drawing aspect. As soon as it's ready, I will merge the functionality into this feature branch and fix merge conflicts. 

You  are welcome to check it out: 

https://github.com/thfrei/infinite-drawing-canvas

thfrei:
For your information: I deleted the branch `canvas-note` on my forked trilium branch, and therefore accidentally closed the pull request #1135 

@zadam can you reopen the PR? The given branch exists again.

Current features Cnavas:
* see https://github.com/thfrei/infinite-drawing-canvas

ToDo:
* Proper Undo/Redo
* Insert images
* Shortcuts (ESC, Delete, etc.)

Trilium Integration ToDo:
* handle tabs (see comment about note-detail selector
* Save note as SVG so that "drawn note" can be integrated into other notes
  * Save as SVG, or trasnform to SVG using API call
* Buttons on the side

2020-11-09 23_22_13-Trilium Notes - c2 - fdsa

sigaloid:
@thfrei @zadam any updates on this PR? I would really like this feature.

PR #1310, opened 2020-10-16
support for applying a label from android sender api

elementc:
Hey, tried to keep this one surgical since I'm super not comfortable in node. If there are tests that need updating, or you have other suggestions for how to do this sanely, please don't hesitate to suggest.

Android side of this change coming in the next couple of days.

As usual, if you think this PR is helpful, please consider applying the `hacktoberfest-accepted` label.

### DIFF
--- a/src/routes/api/sender.js
+++ b/src/routes/api/sender.js
@@ -35,6 +35,14 @@ function saveNote(req) {
         mime: 'text/html'
     });
 
+    if (req.body.label && req.body.label.trim()){
+        let value;
+        if (req.body.labelValue && req.body.labelValue.trim()){
+            value = req.body.labelValue;
+        }
+        note.setLabel(req.body.label, value);
+    }
+
     return {
         noteId: note.noteId,
         branchId: branch.branchId


src/routes/api/sender.js
```diff
@@ -35,6 +35,14 @@ function saveNote(req) {
         mime: 'text/html'
     });
 
+    if (req.body.label && req.body.label.trim()){
+        let value;
+        if (req.body.labelValue && req.body.labelValue.trim()){
+            value = req.body.labelValue;
+        }
+        note.setLabel(req.body.label, value);
+    }
```
zadam, line 44: Here I would prefer to make the interface more generic and accept a list of labels.

If you're not comfortable with node I can merge this and fix to array afterwards ...

elementc: Your decision, I'm probably done writing code for the night so you can merge and fix after, or you can commit directly to this branch so this set of changes doesn't actually go in until the android ui to support it is put together. I've turned on edits by maintainers.

I'm not sure how the UI for setting a list of key-value labels in the preferences UI will work / look in android, but I concur that your preference for the generic capability is the better choice.

zadam: To be clear, Android app doesn't need to allow user to enter more labels at once - it can allow only once for now.

But on the server API side it's better to have this more generic since it's cheap and allow easier extension of the Android app in the future if needed.

elementc: That sounds like a great first step- I was just testing with a hardcoded default one, so currently it allows zero :laughing:

PR #1374, opened 2020-11-02
Use user theme on mobile

good-praxis:
A small pet peeve I had when I transitioned from the stand-alone app to the server was that the chosen theme would only show when in desktop mode. From what I have seem theming doesn't seem to impact mobile layout (correct me if I am wrong @zadam ), so this quick change allows us to use glorious dark mode on our phones!

### DIFF
--- a/src/views/mobile.ejs
+++ b/src/views/mobile.ejs
@@ -94,7 +94,7 @@
         }
     </style>
 </head>
-<body class="mobile">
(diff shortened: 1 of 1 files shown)


Engr-AllanG:
For what its worth, the neon dark theme I created shows up correctly on mobile, without switching to desktop version.

PR #1703, opened 2021-03-01
feat(build): Add github image action

jokajak:
This is a proof of concept to demonstrate what's possible. Any interest in this? Github can automagically publish to docker on push to branches or tags. You would have to update the target name, repository, and secrets. I copied the instructions from https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/ to start and then merged in instructions from https://docs.github.com/en/actions/guides/publishing-docker-images

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,45 @@
+name: build our image
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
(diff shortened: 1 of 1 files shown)


zadam:
Hi, github actions are pretty cool and could replace my custom scripts (in `bin/` folder), but not sure what benefit would it bring ...

jokajak:
I'm specifically interested in its ability to perform a cross platform build that is usable on amd64, arm64, and armv7 platforms automagically. It might not be something you're as interested in which is totally understandable.
I'm using (and loving) trilium as a server based notes app and wanted to make it easier for others to deploy it. I'll try and get it so that builds happen automagically and are tagged automagically but think it would be better to come directly from the source repository.
Keep up the awesome work.

zadam:
I see. I'm hesitant to merge this because then these builds will "look official", although I will not test them.I would prefer if you made the builds in a fork and push to a different docker repo. I will be happy to link to it from README/wiki.

I'm intentionally limiting the amount of platforms I personally support, on the other hand I want to support them in the long term. Spreading myself too thin means less fun and less time for actual development.

jokajak:
I'll work on automating my publishing and mapping versions with tags etc so that you can update the wiki to point to my repository once it's ready.

PR #1756, opened 2021-03-17
Mobile view: iOS sizing/styling changes

### LINKED ISSUE #1312: Mobile Interface: Can't see bottom part of a note
I just updated from 0.34 to 0.44.8. Now I can no longer see the bottom part of a note which also means I can no longer see what I am typing except when I am typing something in the middle of the existing text.
See the images below for an example. I have scrolled down as far as possible in the pictures. The browser used is firefox:

image
image

patrickjm:
### Changes
* Fix bug where action buttons were black and hard to see on dark themes due to Safari user agent styles
* Fix bug where you couldn't scroll on Safari (note detail view + tree view)
* Hopefully fix #1312 
* Note title and MobileGlobalButtonsWidget are now fixed at the top of the page even when you scroll; I added some margin so there's no overlap
* Hid floating buttons at bottom of tree view

### Tested on
* Firefox, Safari on iOS
* Windows Chrome + Firefox on responsive mode with Android user agents
* Tested that normal desktop view was unaffected

### DIFF
--- a/src/public/app/layouts/mobile_layout.js
+++ b/src/public/app/layouts/mobile_layout.js
@@ -86,7 +86,6 @@ export default class MobileLayout {
             .setParent(appContext)
             .id('root-widget')
             .css('height', '100vh')
-            .css('display', 'block')
             .child(new ScreenContainer("tree", 'column')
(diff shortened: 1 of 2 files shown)


patrickjm:
Example of how the note title now sits fixed at the top of the screen and cuts off the note below
image

Similar change on the tree view now
image

zadam (approved):

Issue #1749, opened 2021-03-16
Mobile toolbar buttons


patrickjm:
What do you think about adding a fourth button to the mobile toolbar? It could have a drop down with a list of mobile-only buttons.

`api.addButtonToToolbar({ ..., mobile: true })` could indicate the button is mobile friendly. 

Day notes are a great feature. They're a little difficult to use on mobile since they rely on attributes which are only editable on desktop.

Another use-case is #1389

Thanks

patrickjm:
This has been added, I updated the wiki [here](https://github.com/zadam/trilium/wiki/Events)

To add a toolbar button on mobile, add `#run=mobileStartup` to your code note that calls `api.addButtonToToolbar`.

PR #1762, opened 2021-03-18
Add mobile dropdown for plugin buttons

patrickjm:
Re #1749 

## Changes
* `/scripts/startup` takes new optional `?mobile=true` query param that searches for notes where `#run=mobileStartup`
* App context no longer prevents mobile from making the call
* New toolbar dropdown button holds the `#plugin-buttons` container
* `addButtonToToolbar` frontend api fn checks for mobile and makes a few styling changes

image

Happy to rename the `#run=mobileStartup` attribute search or change the toolbar icon. Wasn't sure the best option for those.

### DIFF
--- a/src/public/app/widgets/mobile_widgets/mobile_global_buttons.js
+++ b/src/public/app/widgets/mobile_widgets/mobile_global_buttons.js
@@ -40,7 +40,6 @@ const WIDGET_TPL = `
         <div class="dropdown-menu dropdown-menu-right">
             <a class="dropdown-item" data-trigger-command="switchToDesktopVersion"><span class="bx bx-laptop"></span> Switch to desktop version</a>
             <a class="dropdown-item" data-trigger-command="logout"><span class="bx bx-log-out"></span> Logout</a>
-            <a class="dropdown-item" ><span class="bx bx-log-out"></span> test</a>
         </div>
     </div>
 </div>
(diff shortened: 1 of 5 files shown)


src/routes/api/script.js
```diff
@@ -53,8 +53,11 @@ function getBundlesWithLabel(label, value) {
     return bundles;
 }
 
-function getStartupBundles() {
+function getStartupBundles(req) {
     if (!process.env.TRILIUM_SAFE_MODE) {
+        if (req.query.mobile === "true") {
+            return getBundlesWithLabel("run", "mobileStartup");
+        }
         return getBundlesWithLabel("run", "frontendStartup");
```
zadam: I think here should be `else`

src/public/app/widgets/mobile_widgets/mobile_global_buttons.js
```diff
@@ -21,6 +26,14 @@ const WIDGET_TPL = `
 
     <a data-trigger-command="scrollToActiveNote" title="Scroll to active note" class="icon-action bx bx-crosshair"></a>
 
+    <div class="dropdown">
+        <a title="Plugin buttons" class="icon-action bx bx-chip dropdown-toggle" data-toggle="dropdown"></a>
```
zadam: For consistency, I would prefer `bx-extension` should be used since that's also used for conceptually similar "render note".

patrickjm: Updated

src/routes/api/script.js
patrickjm: Updated

zadam (approved):
Thanks!

PR #1827, opened 2021-04-05
Fix for Evernote workaround and support for en-todo

MatAsmania:
@zadam Great tool, thanks. I am currently in the process of transitioning from Evernote and noticed that the workaround did not trigger correctly. Furthermore, I have added support for en-todo using unicode characters. Using the CKEditor todo lists instead seem to be more involved.


zadam:
Thanks!

PR #2065, opened 2021-07-14
Render math in zip/single export

kees-:
Adds a snippet for KaTeX's Auto-render Extension ([docs](https://katex.org/docs/autorender.html)) to render equations on html note exports.

Adding `<!doctype html>` to `content` becomes necessary or else things fail:
`> Warning: KaTeX doesn't work in quirks mode. Make sure your website has a suitable doctype.`

### Example
ss-210714-13 47 26

#### Before
ss-210714-13 48 22

#### After
ss-210714-13 49 33

---

- Ok to add custom js into static page exports?
- Should snippet declaration be moved to avoid redundancy?
- Other math renderers do work, e.g. [MathJax](https://www.mathjax.org/#gettingstarted), with slightly different supported commands. Since it is KaTeX I assume their own web renderer is best.

### DIFF
--- a/src/services/export/single.js
+++ b/src/services/export/single.js
@@ -23,7 +23,11 @@ function exportSingleNote(taskContext, branch, format, res) {
     if (note.type === 'text') {
         if (format === 'html') {
             if (!content.toLowerCase().includes("<html")) {
-                content = '<html><head><meta charset="utf-8"></head><body>' + content + '</body></html>';
+            	// KaTeX Auto-render Extension — from https://katex.org/docs/autorender.html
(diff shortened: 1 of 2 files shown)


zadam:
Hi. I'm not sure about this. My concerns:

1) this loads an external script. Depends on an internet connection and some specific CDN.
2) it adds these scripts even when note doesn't have math formulas. Probably more than 99% of notes don't have any math formulas ...

kees-:
Understandable and thanks for the notes. Yea this fix would be sloppy.

Maybe something like this one day with better handling, i.e smart detecting notes, opt-in checkbox in settings. I'll just try scripting and extending my own build for now. Feel free to close PR

kees-:
Understandable and thanks for the notes. Yea this fix would be sloppy.

Maybe something like this one day with better handling, i.e smart detecting notes, opt-in checkbox in settings. I'll just try scripting and extending my own build for now.

PR #2084, opened 2021-07-24
Fix math panel overflow - closes #2056

### LINKED ISSUE #2056: The formula bar overflows from the app and is not visible
image

abitofevrything:
The issue was originally caused by the math plugin only allowing the balloon to show up below the target. This pull request allows it to show on top as well, fixing the overflow issue.

### DIFF
--- a/src/public/app/widgets/type_widgets/editable_text.js
+++ b/src/public/app/widgets/type_widgets/editable_text.js
@@ -125,6 +125,8 @@ export default class EditableTextTypeWidget extends AbstractTextTypeWidget {
             }
         });
 
+        this.preventMathInputOverflow();
+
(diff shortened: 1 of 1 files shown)


zadam:
Hi, wouldn't it be better to fix this in the upstream - https://github.com/isaul32/ckeditor5-math ? It would then benefit all projects using this plugin ...

abitofevrything:
It seems the issue was added in [this commit](https://github.com/isaul32/ckeditor5-math/commit/aa0392c12fb22525d2c8b83aab5e2e92edae2bbc).

I will open an issue there and see if this is intentional.

skitou:
Any updates about this? The author of the plugin seem to be active again as he published a new release a month ago.

PR #2134, opened 2021-08-29
Address 2 moderate severity vulnerabilities

jeffs:
This commit is the result of running `npm audit fix`.


zadam:
I had to run the command myself since I did some package upgrades beforehand.

Seems like these vulnerabilities were in the deps of `electron-packager` which is used only in the build process.

PR #2175, opened 2021-09-20
Add shortcut to toggle readonly mode of notes

### LINKED ISSUE #1424: Shortcut for making a note readOnly and vice versa
It would be easier to make notes read only and read using arrow keys. Because rightnow with editing mode, arrow keys moves the note one line at a time. So I need to make a note read only quickly and do the reading.
Basically I need to be able to read the note like I am reading something on a document or website, without cursor showing up.

abitofevrything:
Closes #1424

### DIFF
--- a/src/public/app/services/note_context.js
+++ b/src/public/app/services/note_context.js
@@ -40,7 +40,8 @@ class NoteContext extends Component {
         this.notePath = resolvedNotePath;
         this.noteId = treeService.getNoteIdFromNotePath(resolvedNotePath);
 
-        this.readOnlyTemporarilyDisabled = false;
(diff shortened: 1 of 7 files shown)


abitofevrything:
Any update on this? This has been open for over a week without any response.

zadam:
I'm not yet decided on whether this is worth it or not. Making editable note temporarily read-only just for easier scrolling seems to me like a very niche feature. Also having functionality attached only to a keyboard shortcut (and no UI) means it's not discoverable, which will make it even more niche. Features which almost nobody uses should be avoided since they will still cost time/effort in the maintenance/testing in the future.

Issue #948, opened 2020-04-07
[Feature request] mermaid diagrams


BigGitWorld:
First of all
@zadam
Thank you for this nice note taking app.
I wonder why the number of stars of this app is less than its competitors (like the Notable).
I was comparing the trilium vs the Notable and see the later supports the [mermaid diagrams](https://github.com/mermaid-js/mermaid)  which is very useful for at least developers.
Can we use the mermaid in the trilium now?

zadam:
Hello, yes, I do have plans in this direction, but I'm trying to get "basics" right first and then I will work on adding more note types (besides diagrams I'd also like to add math/latex support).

BigGitWorld:
Thank you very much
Nice job!

aniston:
@zadam  just checking on this mermaid diagrams Feature Request, any update or is this already possible ?

From a user point of use, lovely work with bringing us 'trilium', thanks ! first try and like it, looking forward towards the multiple user and public  interface possibility.

cheers,
Aniston

PR #2187, opened 2021-09-27
Implement mermaid diagrams

abitofevrything:
Mermaid diagrams have already been requested in #948, and I personally think that they would be very useful for a note taking app, allowing for more free-form notes.

This isn't ready for merge yet, as I would like to know your opinion on how certain things are handled:
- How should editing diagrams be done? At the moment you can edit them by changing the note type to  `code/plain text`, making your edit and then switching back to the `Mermaid Diagram` note type, but this is obviously not ideal.

- How should theming support be added? Unlike CKEditor and CodeMirror which provide CSS variables to change the theme, mermaidjs expects a config option to change the theme. It would be possible to have the JS code pull a CSS variable from the document stylesheet (e.g `--mermaid-theme`) and pass that as the config option, but this isn't the most elegant and also restricts the possibilities (`base`, `dark`, `default`, `forest` and `neutral` would be the only theme options), but allowing users to change the theme through JS would mean sacrificing the one-file-per-theme option which, I think, is quite a nice thing to have.

I'd like to know what your opinion on this is @zadam and hopefully it can be sorted out.


zadam:
Hi, thanks for this effort!

1) the way I would do it would be "mermaid" note type being handled as code note - i.e. code editor will be opened. And there will be new mermaid render widget above the code editor which would listen on note changes and re-render them. SQL console note type is done in somewhat similar way.

2) I don't think that advanced theme config of Mermaid would be done by many people. In similar vein the code note component CodeMirror is also themeable but you can't choose its theme in Trilium - it's constant. But nobody even asked for it yet.

So I think having limited amount of themes is a good tradeoff for this "one file" themability.

abitofevrything:
I have left mermaid diagram notes with the default ribbon because switching from code note to mermaid diagram might be something users would like to do.

PR #2189, opened 2021-09-29
enex: Convert OneNote (via Evernote) todo images to Unicode checkboxes

bjj:
When a OneNote todo bullet is imported by Evernote it is converted to an
image showing the checkbox the way it looked in OneNote. This change
takes advantage of the consistency of that image hashing to convert
those images into unicode checkboxes (matching the existing ENEX
<en-todo> behavior) . It also removes the image attachments.


zadam (approved):

bjj:
Something I realized after this pull request:  Currently all of the image linking in ENEX imports is broken because HTML sanitize step was added before the code that replaces `<en-media>` tags. So those tags are all stripped out, which is why I only saw one of each kind of checkbox as an attachment. Prior to the sanitize breaking media links, ENEX import of OneNote checkboxes probably drew the OneNote image checkboxes.

I will try to fix the non-checkbox media in another pull request.

PR #2190, opened 2021-09-30
html_sanitizer: Demote H1 tags

bjj:
Generalize clipper.js support for removing H1 tags so it also applies to
imports. This will move all heading levels down to make room if
needed, only leaving duplicates at H6.

### DIFF
--- a/src/routes/api/clipper.js
+++ b/src/routes/api/clipper.js
@@ -104,10 +104,7 @@ function createNote(req) {
 }
 
 function processContent(images, note, content) {
-    let rewrittenContent = htmlSanitizer.sanitize(content)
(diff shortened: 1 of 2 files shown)


tredondo:
I found the link to this issue at https://github.com/TriliumNext/Notes/releases/tag/v0.94.1, which says "@zadam was kind enough to give us the original name. See [#2190](https://github.com/orgs/TriliumNext/discussions/2190) for more info."

Was that a wrong issue number @zadam ?

eliandoran:
@tredondo , the issue number was correct but since we've transferred the issue numbers from older version no longer correspond, unfortunately.

Here's the right one: https://github.com/orgs/TriliumNext/discussions/5867

PR #2191, opened 2021-09-30
enex: Turn "Created with OneNote comment" into attribute

bjj:
A boilerplate line is added by Evernote import. This converts the text
into an attribute.

### DIFF
--- a/src/services/import/enex.js
+++ b/src/services/import/enex.js
@@ -221,6 +221,17 @@ function importEnex(taskContext, file, parentNote) {
 
         content = extractContent(content);
 
+        // This boilerplate is added by the Evernote import of OneNote notes (and preserved in ENEX export)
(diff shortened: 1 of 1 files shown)


src/services/import/enex.js
```diff
@@ -221,6 +221,17 @@ function importEnex(taskContext, file, parentNote) {
 
         content = extractContent(content);
 
+        // This boilerplate is added by the Evernote import of OneNote notes (and preserved in ENEX export)
+        const replaced = content.replace(/<div>\s*<p>\s*<\/p>\s*<p>\s*Created with OneNote.\s*<\/p>\s*<\/div>\s*$/m, "");
+        if (replaced != content) {
```
sigaloid, line 226: I think this could simply check `content` to see if it contains the string, then if it does, then replace it and add the attribute. No need to clone the entire note into a new variable then compare them.

bjj:
I think I will close this for now. The changes to get good import are much more extensive than I expected, so I don't think submitting them as a nice series of small, atomic changes is going to work.

PR #2193, opened 2021-10-01
Add option to copy note ID in context menu

abitofevrything:
This can be useful e.g when copying an image note to get the URL to insert it into a mermaid diagram.


zadam:
Frankly I don't believe this is important enough to warrant an entry in the context menu. NoteId is mostly an internal thing and user should not really care about it ...

abitofevrything:
The use case i came across that would justify this is inserting images in mermaid diagrams: a link to the image is required, and that link contains the note with the image data's ID:`/api/images/<note id>/<file name>` (on a side note, `<file name>` isn't used at all and is just a useless parameter). Unless we implement a different way to insert images into mermaid diagrams, this seems like the best option...

PR #2197, opened 2021-10-03
Fix "Theme defined" option not working in font override

abitofevrything:
Previously choosing the "Theme defined" option would set font to `theme`, not leaving it as defined by the current theme.
The fonts route now checks for this specific value, and if it is set to `theme`, does not send the override as part of the response.


zadam:
Thx.

PR #2235, opened 2021-10-15
add heroku deployment support

feilongfl:
Hi, 
I create this PR for setup trilium in `heroku` service.

The reason for me to use this is I need to sync my data between sevrial computers in different place.
and i do not have an public IP address can be visited everywhere.
so I found heroku service.

### usage:

1. fork the repo in github
2. Create New App with heroku
3. in heroku app deploy tabs, link to gitlab & select your repo.
4. select `deploy branch`and wait it done.
5. finally, click `Open App` button, then you'll see the trilium web page.

### Shortcomings
1. when restart heroku dynos nothing can be keep, so this plan can only sync data temporary at this time.
I've try to solve this problem by trying mount webdav in docker, but It need more permisson to use `mount` command in docker.
not support by heroku.
then I try to modify trilium to use webdav to save every thing, then I found https://github.com/zadam/trilium/issues/242#issuecomment-442897744 seems like a difficult way too.
so I don't have any other idear to solve this problem.maybe we can solve it later.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,6 @@
 /docs
 /npm-debug.log
 node_modules
+/heroku.dockerfile
+/app.json
+/heroku.yml
--- a/app.json
+++ b/app.json
@@ -1,0 +1,7 @@
+{
+    "name": "trilium",
+    "description": "A barebones Node.js app using Express",
+    "repository": "https://github.com/zadam/trilium",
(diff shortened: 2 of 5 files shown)


feilongfl:
Thank you for reply so fast.
I understand your concerns, and will close this PR.

feilongfl:
@zadam 
I've updated a `heroku` deploy button in https://github.com/zadam/trilium/wiki page.

dafmdev:
I have an error when I use the heroku button, please help me

PR #2236, opened 2021-10-15
Add origin parameter

sigaloid:
Requires: https://github.com/zadam/trilium-web-clipper/pull/32

This allows for an "origin" attribute, so I can search all of my web saved notes by the domain. Example: 
image
So I would be able to search by this attribute to find any saved stories in this domain.

image
image
This then allows for searching by domain: 
image

Question: Should the attribute name be changed to `clipOrigin` or equivalent? Just a thought :p

### DIFF
--- a/src/routes/api/clipper.js
+++ b/src/routes/api/clipper.js
@@ -43,7 +43,7 @@ function getClipperInboxNote() {
 }
 
 function addClipping(req) {
-    const {title, content, pageUrl, images} = req.body;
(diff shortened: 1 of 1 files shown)


zadam:
Hi, there's no need for a special label for domain, you can just use `=*` [search operator](https://github.com/zadam/trilium/wiki/Search) which matches prefix:

```
#pageUrl =* "https://www.theglobeandmail.com"
```

sigaloid:
Oof, that's right. :face_with_head_bandage:

PR #2246, opened 2021-10-19
Set default icon for clipped notes

sigaloid:
I like being able to differentiate between clipped notes and regular notes, picked the first one that made sense to me


zadam:
Hi, what about [globe](https://boxicons.com/?query=globe) - sometimes used for internet, as in it's a content from internet.

Clip seems to me a bit misleading - the first thing which comes to my mind is some email attachment which is not the case.

PR #2271, opened 2021-10-25
Update "Sync from Desktop" setup instructions.

meichthys:


src/views/setup.ejs
```diff
@@ -134,12 +134,12 @@
             <p>This setup needs to be initiated from the desktop instance:</p>
 
             <ol>
-                <li>please open your desktop instance of Trilium Notes</li>
-                <li>click on Options button in the top right</li>
```
meichthys, line 138: The options menu hasn't been on the top right for a while now.

PR #2310, opened 2021-11-05
fixed button background

### LINKED ISSUE #2285: (Feature request) Add opacity/background to floating buttons
### Preflight Checklist

### Describe feature

The floating buttons on the bottom of the node list ("id=left-pane") can get lost when 'floating' above node text:
image

I would recommend that we add some opacity to the buttons or move them into a bottom 'toolbar' div (like the quick search bar).

### Additional Information

_No response_

Myzel394:
Improved the background of the tree settings buttons

Screenshot:
image

closes #2285

### DIFF
--- a/src/public/app/widgets/note_tree.js
+++ b/src/public/app/widgets/note_tree.js
@@ -39,7 +39,6 @@ const TPL = `
     
     .tree-actions {
         padding: 4px 0;
-        width: 100%;
         background-color: var(--launcher-pane-background-color);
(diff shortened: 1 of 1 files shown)


zadam:
Hi, I don't like that it uses background of the whole row, which wastes space especially if the left pane is wide enough that the buttons don't overlay the tree content. It would be better to give background only to the buttons themselves.

Myzel394:
What about this?

image

PR #2329, opened 2021-11-11
Add update available box

### LINKED ISSUE #2319: (Feature request) Add a "Update available" box
### Preflight Checklist

### Describe feature

I would like to be alerted whenever there's a new version available.

### Additional Information

We could use a simple `alert` in the web version, send a notification in the electron version and it any of them fails show a dialog box (that gray box at the top) otherwise.

I can implement this feature if you'd like :)

Myzel394:
closes #2319

### DIFF
--- a/src/public/app/dialogs/options/appearance.js
+++ b/src/public/app/dialogs/options/appearance.js
@@ -281,7 +281,7 @@ export default class ApperanceOptions {
         })
 
         this.$newVersionAvailable = $("#new-version-available");
-        const onNewVersionAvailableChange = async () => {
+        this.onNewVersionAvailableChange = async () => {
             const isChecked = this.$newVersionAvailable.prop("checked");
 
             // Enable / Disable new-minor-version-available input
@@ -289,18 +289,16 @@ export default class ApperanceOptions {
 
             $newMinorVersionAvailableWrapper.prop('disabled', !isChecked);
 
-            await server.put('options/showNewVersionAvailable/' + isChecked);
+            await server.put('options/newVersionAvailable/' + isChecked);
         }
 
-        this.$newVersionAvailable.on('change', onNewVersionAvailableChange);
-        // Initial call to disable other input if checkbox is disabled
-        onNewVersionAvailableChange();
+        this.$newVersionAvailable.on('change', this.onNewVersionAvailableChange);
 
         this.$newMinorVersionAvailable = $("#new-minor-version-available");
         this.$newMinorVersionAvailable.on('change', async () => {
             const isChecked = this.$newMinorVersionAvailable.prop("checked");
 
-            await server.put('options/showNewMinorVersionAvailable/' + isChecked);
+            await server.put('options/newMinorVersionAvailable/' + isChecked);
         });
     }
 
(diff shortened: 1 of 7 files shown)


src/public/app/dialogs/options/appearance.js
```diff
@@ -181,6 +188,23 @@ const TPL = `
...
+                    <input type="checkbox" name="new-version-available" id="new-version-available">
+                    <label for="new-version-available">Inform about new versions</label>
+                </div>
+                <div class="new-minor-version-available">
+                    <input type="checkbox" name="new-minor-version-available" id="new-minor-version-available">
+                    <label for="new-minor-version-available">Also inform about minor versions</label>
+                </div>
+            </div>
+        </div>
+    </section>
```
zadam: I think this is rather overcomplicated - if the check is not intrusive then there's no need to turn it off. I don't think this needs to be configurable.

src/public/app/widgets/buttons/global_menu.js
```diff
@@ -89,15 +108,35 @@ const TPL = `
...
+        const [newMajor, newSemiMajor, newMinor] = newVersion.split(".").map(Number);
+
+        if (newMajor !== oldMajor) {
+            return "major";
+        } else if (newSemiMajor !== oldSemiMajor) {
+            return "semi-major";
+        } else if (newMinor !== oldMinor) {
+            return "minor";
+        }
+    }
```
zadam, line 135: Standard terminology is MAJOR.MINOR.PATCH (see e.g. [this](https://semver.org/) even though Trilium does not follow semantic versioning strictly). But I don't think there's a need to differentiate between versions ...

src/public/app/widgets/buttons/global_menu.js
```diff
@@ -114,7 +153,52 @@ export default class GlobalMenuWidget extends BasicWidget {
...
+
+        this.$widget.find(".global-menu-button-update-available").append(
+            new UpdateAvailableWidget()
+                .withVersionChange(versionChange)
+                .render()
+        )
+    }
+
+    async checkIfShowNewVersionAvailable(versionChange) {
+        const data = await server.get("options");
```
zadam: There's `services/options.js` which already contains cached options, no need to fetch them from server.

zadam (changes requested):

src/public/app/widgets/buttons/global_menu.js
Myzel394: I would pledge to differentiate between versions because some users will probably not update a patch (because a patch often happens and most people find them annoying) but are welcome to minor are major versions (because they often contain new features)

src/public/app/dialogs/options/appearance.js
Myzel394: The second input only makes sense if the first one is checked; So I think it's better UX to disable it to make it clear that the second input (if the first one is disabled) doesn't change anything

src/public/app/widgets/buttons/global_menu.js
Myzel394: Which `options` do you mean? `app/services/options` or `services/options`?

zadam: `service/options.js`

src/public/app/dialogs/options/appearance.js
zadam: I think you misunderstand me. What I'm saying is that the feature should be non-intrusive so that there's no point in turning it off. Which means there shouldn't be any configuration at all.

Myzel394: I think the feature is non-intrusive but just like [this commenter](https://github.com/zadam/trilium/issues/2319#issuecomment-966206716) just said, there are people who don't want any notification for new updates at all

zadam: Perhaps, although I'm not sure why this would bother people. Anyway, I would start simple - MVP style. It's very common that features are over-engineered with many options with very little value but high maintenance costs.

Let's make it minimal style with no config, then gather feedback after this is released and then we can improve it iteratively (if it proves necessary).

Myzel394: fixed

zadam (approved):

PR #2381, opened 2021-11-25
Add top banner

### LINKED ISSUE #2366: (Feature request) Add non-intrusive internet connection lost message
### Preflight Checklist

### Describe feature

When you have no internet connection and try to open notes, you will always get these notification boxes. They are not really understandable by people who are not versatile with computers.

I would suggest creating a non-intrusive banner at the top that shows that the internet is missing (see screenshot example from protonmail.com)

image

### Additional Information

_No response_

Myzel394:
closes #2366 

This is not the full PR yet. @zadam what do you think so far? If you are ok with this approach so far, I will add the internet-lost-message.


src/public/stylesheets/theme-light.css
```diff
@@ -68,4 +68,11 @@ html {
     --link-color: blue;
 
     --mermaid-theme: default;
+
+    --banner-background-color-error: #DA321B;
+    --banner-background-color-info: #1bb7da;
+    --banner-background-color-warning: #daad1b;
+    --banner-background-color-success: #67da1b;
+    --banner-background-color-plain: var(--more-accented-background-color);
+    --banner-color: var(--main-text-color);
```
zadam, line 73: I don't think it's needed to have this themable since it's for exceptional cases only and user shouldn't see it often.

src/public/stylesheets/style.css
```diff
@@ -55,10 +55,6 @@ table td, table th {
     background-color: var(--modal-backdrop-color) !important;
 }
 
-.component {
-    contain: size;
-}
```
zadam: This PR shouldn't be changing such fundamental classes.

src/public/app/widgets/containers/flex_container.js
```diff
@@ -10,4 +10,10 @@ export default class FlexContainer extends Container {
 
         this.attrs.style = `display: flex; flex-direction: ${direction};`;
     }
+
+    withFullSize() {
+        this.attrs.style += `width: 100%; height: 100%;`;
```
zadam: This does not look correct.

The root container is already 100% height, banner has certain fixed height and then the rest (the main app) should be filled with flex-grow (`filling()` method)

Alternatively, wouldn't it be better to have this floating with absolute coordinates above (in the z-index sense) the main content? Since with your approach the main app will jump up and down as connection is lost/renewed.

In such case bootstrap toasts might be used as an existing solution.

src/public/app/widgets/banner_message.js
```diff
@@ -0,0 +1,174 @@
...
+        super();
+    }
+
+    doRender() {
+        this.$widget = $(TLP);
+        this.$bannerParagraph = this.$widget.find("p");
+        this.$timer = this.$widget.find(".timer");
+
+        this.$widget.on("mouseenter", this.pauseTimer.bind(this));
+        this.$widget.on("mouseleave", this.resumeTimer.bind(this));
```
zadam, line 87: Can you clarify what's going on here?

In my mind it's really simple - (non-pausable) timer, when it detects that the network is down it will show the banner, once it's up again, banner disappears. But here I see info, warning, success, plain, what is all that for?

zadam (changes requested):

src/public/app/widgets/containers/flex_container.js
Myzel394: > Alternatively, wouldn't it be better to have this floating with absolute coordinates above (in the z-index sense) the main content? Since with your approach the main app will jump up and down as connection is lost/renewed.

I don't think this is a good idea. This is what it looks like:

image

src/public/app/widgets/banner_message.js
Myzel394: In case we want to show other notifications, they should be pauseable

src/public/app/widgets/containers/flex_container.js
zadam: What about this:

image

can be done with few lines of code:

```javascript
toastService.showPersistent({
    id: "network-down",
    title: "Network is down!",
    message: "Trilium backend is currently not accessible.",
    icon: "wifi-off"
})
```

Myzel394: That's intrusive. You can't see the full title anymoe and some buttons are harder to reach. If you're on a plane, editing your notes with only partial internet, this would drive you crazy since you wouldn't be able to do your workflow properly.

zadam: When you're on a plane and don't have internet, basically no functionality will work correctly. You should in fact stop editing until you have an internet connection, otherwise you risk data loss.

src/public/app/widgets/banner_message.js
zadam: To be honest I don't understand what "pausable notification" is, but in general YAGNI principle applies here.

src/public/app/widgets/containers/flex_container.js
Myzel394: Wait - just for clarification. Does Trilium Notes save the progress offline and then uploads it once it has a connection back?

zadam: Web version - no, it has no offline storage.

Desktop version works fully offline and then synchronizes with the server periodically.

Myzel394: Okay in that case your solution would fit as well for the web version (only). However, we should probably change the location to the top right (or something less intrusive).

lovitus:
@Myzel394 I do really think to have a place for showing the connect status and sync status for WEB version is a great idea. 
I use Web version more than desktop version on my many devices. 
each time I see the pop-up dialog shows connection problem , I have to copy entire content , then refresh the page to paste it again .
I do not sure if my all changes are saved on server correctly , or do I still connected to the server .

PR #2395, opened 2021-11-29
Improve tabs background coloring

Myzel394:
Add some feedback when you hover over the tabs


Myzel394:
Tab close button shouldn't show a cursor pointer

zadam:
I don't get it - when I apply your changes I don't see any difference besides the pointer.

Also, all buttons have pointer cursor, why should close button differ in this?

Myzel394:
Strange, I also changed the background color - gonna take a look at this later. 

> Also, all buttons have pointer cursor, why should close button differ in this?

Tabs usually don't use pointers (see Chrome, Firefox, etc.)

zadam:
> Tabs usually don't use pointers (see Chrome, Firefox, etc.)

Because they don't use point for buttons, so they are consistent with themselves.

But in Trilium buttons have cursor pointer so again this should be consistent.

Myzel394:
Okay that makes sense. Reverted that change - Could you merge the background color hovering feedback please?

zadam:
I don't see any changes after applying the CSS. It's not clear what are you trying to achieve.

src/public/app/widgets/tab_row.js
```diff
@@ -187,6 +187,14 @@ const TAB_ROW_TPL = `
...
     }
     
+    .tab-row-widget .note-tab:hover .note-tab-wrapper {
+        background-color: var(--accented-background-color);
+    }
+    
+    .tab-row-widget .note-tab[active]:hover .note-tab-wrapper {
+        background-color: var(--more-accented-background-color);
+    }
+
```
Myzel394: This has changed

zadam: Yes, I can see the code change, but no visual change. It literally looks exactly the same.

Myzel394: Oh I see, it's only in light mode (I guess because the css variables are probably the same). In dark mode it looks like this (maybe we should update the light mode variables?):

https://user-images.githubusercontent.com/50424412/144488765-eee8d28f-a7cf-493e-8c8f-2f7734be8079.mp4

zadam: I see. Tab coloring is controlled by separate CSS variables (e.g. `--active-tab-background-color`) so I guess better way would be to define new "hover background" variable for tab which would solve this theme inconsistency.

tobealive:
Nice good feature.

The themes I've created for Trilium already had it implemented.

src/public/app/widgets/tab_row.js
zadam: Don't forget the light theme also please.

PR #2398, opened 2021-11-30
Add loading indicator

Myzel394:
I think it would be useful if there is some kind of loading indicator when clicking on a note

### DIFF
--- a/src/public/app/services/note_context.js
+++ b/src/public/app/services/note_context.js
@@ -33,7 +33,6 @@ class NoteContext extends Component {
     }
 
     setLoading() {
-        console.log("setLoading")
         this.notePath = null;
         this.noteId = null;
         this.parentNoteId = null;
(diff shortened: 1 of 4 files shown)


Myzel394:
Hi @zadam, this PR is not done yet. I'm currently figuring out how to replace the page. Could you help me out please?

zadam:
Why do you mean by "replace a page"?

Myzel394:
Replace the note type (See `EmptyTypeWidget`, `SearchWidget`, etc.)

zadam:
These widgets are controlled by note type, but "loading" is no note type.

So I guess I would probably try to solve that in another way. Maybe some completely separate widget somewhere [here](https://github.com/zadam/trilium/blob/89322c4b03748f2904ea6c0a108e4996e2cf512c/src/public/app/layouts/desktop_layout.js#L150)

But in general I'm not sure how to do this right. It's important that when the networks is fast there shouldn't be any flashing caused by this loading appearing and quickly disappearing.

PR #2466, opened 2021-12-24
Display PDF in shared notes 

sigaloid:
Render PDF in share page. I had to add an API endpoint to remove the content-disposition header, otherwise the file would just be opened as a download. `api/notes/${note.noteId}/view` 

I wonder what else an iframe could embed. Maybe other mimetypes could be rendered like this. PDF was the obvious first target, though.

### DIFF
--- a/src/share/routes.js
+++ b/src/share/routes.js
@@ -78,6 +78,25 @@ function register(router) {
 
         res.send(note.getContent());
     });
+    router.get('/share/api/notes/:noteId/view', (req, res, next) => {
+        const {noteId} = req.params;
+        const note = shaca.getNote(noteId);
+
+        if (!note) {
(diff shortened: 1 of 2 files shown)


src/share/routes.js
```diff
@@ -78,6 +78,26 @@ function register(router) {
...
+
+        if (!note) {
+            return res.status(404).send(`Not found`);
+        }
+
+        const utils = require("../services/utils");
+
+        const filename = utils.formatDownloadTitle(note.title, note.type, note.mime);
+
+        // res.setHeader('Content-Disposition', utils.getContentDisposition(filename));
```
sigaloid, line 94: I left this commented out to better show the difference between the two API endpoints, and to clearly show that this does not send that header.

Issue #2327, opened 2021-11-11
(Bug report) Can't change TRILIUM_DATA_DIR with the OneDrive directory location


skitou:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/zadam/trilium/issues) for a bug report that matches the one I want to file, without success.

### Trilium Version

v0.48.5

### What operating system are you using?

Windows

### What is your setup?

Local (no sync)

### Operating System Version

Windows 11 / Windows 10

### Expected Behavior

I got a computer for my studies in highschool/college, the computer is set up in organisation mode with a OneDrive storage for each student with 1TB.

I tried using the portable version of Trillium by modifying the TRILIUM_DATA_DIR variable in the .bat file:
```
SET DIR=%~dp0
SET TRILIUM_DATA_DIR=D:\Users\gally\OneDrive\OneDrive - Conseil Régional des Pays de La Loire - Direction des Lycées\Lycee\trilium-data
cd %DIR%
start trilium.exe
```

### Actual Behavior

When launching the .bat I got that error message:
![image](https://user-images.githubusercontent.com/38993289/141333246-0d3ef185-b825-4c93-beab-242526eff845.png)

### Additional Information

- I have already done that with Google Drive and got no error, Trillium launched like a charm.
- The computer is offered by the regional academic of my country -> I have no administration permissions on it and I can't change the OneDrive directory name.
- The computer I use at home is on Windows 11 and the other computer is on Windows 10.

zadam:
Might be some problem with spaces and/or accents. Not sure.

It's not recommended to sync the database file with tools like OneDrive or Google Drive - they are known to corrupt the DB file. See e.g. #995 or #913.

abitofevrything:
Note that all of the folders in the path must already exist (excluding `trilium-data` which is auto-generated). So the folder `D:\Users\gally\...\Lycee` must already exist.

Did you make a typo? I'm surprised the last `Lycee` is spelt without an accent (`Lycée`)

skitou:
> Note that all of the folders in the path must already exist (excluding `trilium-data` which is auto-generated). So the folder `D:\Users\gally\...\Lycee` must already exist.

All the folders already exists.

> Did you make a typo? I'm surprised the last `Lycee` is spelt without an accent (`Lycée`)

I named it without the accent to avoir any problems with the path.

skitou:
> Might be some problem with spaces and/or accents. Not sure.

I think it is a problem with path with accents.

[...]

Issue #1578, opened 2021-01-27
Trilium-Notes - Internal Server error, if rebooted while docker image is running


zDEFz:
Ver 0.45.8
[trilium-2021-01-27.log](https://github.com/zadam/trilium/files/5879635/trilium-2021-01-27.log)

What doesn't help: Deleting Browser Cache. 

What helps: Stop docker container before reboot and then reboot. 
Drove me crazy. 

Luckily, we have daily backups.

Log afterwards:
[trilium-2021-01-27.log](https://github.com/zadam/trilium/files/5879701/trilium-2021-01-27.log)

Also, database grew from 1.9 GB to 5.8 GB over night magically. Have not changed anything.

zadam:
Hi, you're right with the `-wal` and `-shm` files, those need to be deleted as well. For now I fixed at least the docs on restoring the backup: https://github.com/zadam/trilium/wiki/Backup#restoring-backup

zDEFz:
Any update on this?

PR #2561, opened 2022-01-15
Add powershell equivalents to existing batch scripts

U-Bren:
Hi, as to avoid startup scripts crashing when paths (especially TRILIUM_DATA_DIR) with accents and other non-ASCII characters are provided (part of issue #2327), I have made powershell equivalents to them. I have searched for a solution in pure batch, but unfortunately nothing I could find worked.

This shouldn't break compatibility with systems where powershell is unavailable (Vista and previous versions), as each .bat will try to invoke its .ps1 version if powershell is available, or else will fallback on the current behavior (with obviously the same issue with non-ASCII).

Also, as I will create for my own usage a script that open the automatically generated backups (perhaps with a menu selection) in a new folder and with a new port so that I can ~~copy~~ compare notes from the backup to the current document, and provided that it could be merged, should I create another PR or would it be simpler to add it to this one, as it is to do with scripts anyway? 
It isn't exactly what #1578 requested, but it would at least be useful for a local usage.


zadam:
Hi,

> as each .bat will try to invoke its .ps1 version if powershell is available

I don't see this behavior in the PR...

> Also, as I will create for my own usage a script that open the automatically generated backups (perhaps with a menu selection) in a new folder and with a new port so that I can copy compare notes from the backup to the current document, and provided that it could be merged, should I create another PR or would it be simpler to add it to this one, as it is to do with scripts anyway?

I would not provide this in the menu. Backups are there for catastrophic situations like corrupted DB etc., user shouldn't need to touch them in normal usage (and should be let to believe the through UI).

I'm also somewhat skeptical towards trying to automate this process, since it's not something people should actually have to do (often). You also have to provide that for all platforms, there are some potential problems (sync) ...

U-Bren:
> I don't see this behavior in the PR...

My bad, I forgot to include the .bat in the commit... Will fix that ASAP.

> I would not provide this in the menu. Backups are there for catastrophic situations like corrupted DB etc., user shouldn't need to touch them in normal usage (and should be let to believe the through UI).

To be clear, I wasn't talking about the Trilum menu. I was talking about a [TUI] to open (a copy of) either the daily, weekly or monthly backup, and launch another instance of trilium. Basically a trilium-portable.sh but for recovery from backups.
This should hopefully make it more easy to recover from backups (no need to replace the current document.db with a backup, which run the obvious risk of accidentally overwriting it) and also ensure that only a copy of the backup is opened, so as to never modify the original.

> You also have to provide that for all platforms

I don't think that this is an issue, as it would be a simple shell script. 
> there are some potential problems (sync) ...

The only goal is to simplify launching a trilium instance using a copy of a backup and another port. I guess that this shouldn't cause a pushing old data issue with sync, unless one of the note were to be modified, and since it would operate on a copy of the backup pulling new data shouldn't matter.


> I'm also somewhat skeptical towards trying to automate this process, since it's not something people should actually have to do (often).

Quite understandable. Let's refocus on the PR itself, I'll write the script later and perhaps it will be clearer then 🙂.

PR #2629, opened 2022-02-11
fix broken Dockerfile

velmoga:
I want to try trilium. 
I fetched repo, run `docker built .` and it failed. 
After my fix it started.


zadam:
Hi, the correct way to build docker image is via `bin/build-docker.sh` script which pre-processes the package.json.

Issue #2239, opened 2021-10-16
(Bug report) Docker Compose file needs to be updated


Unrepentant-Atheist:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/zadam/trilium/issues) for a bug report that matches the one I want to file, without success.

### Trilium Version

latest

### What operating system are you using?

Other Linux

### What is your setup?

Local (no sync)

### Operating System Version

Ubuntu live server 20.10

### Expected Behavior

To have data in data-Folder. 

```
version: '2.1'
services:
  trilium:
    build:
      context: .
    image: zadam/trilium
    restart: always
    environment:
      - TRILIUM_DATA_DIR=/data
    ports:
      - "8080:8080"
    volumes:
      - trilium:/data

volumes:
  trilium:
```

Neither `- TRILIUM_DATA_DIR=/data` nor `volumes: - trilium:/data` does anything meaningful. The data folder is empty and recreated everytime the docker image is recreated -> It stores the data inside the container itself. 

### Actual Behavior

Empty data folder

Unrepentant-Atheist:
Solution: `chown 1000:1000` the trilium volume.

zadam:
Hi, I'm not sure what exactly needs to be fixed. Would you mind creating a PR?

Thanks.

zadam:
I see. This file was contributed externally as a sort of example, maybe it's missing e.g. the exact directory mapping which needs to be filled by the end-user.

If the file is more confusing than helpful I might just delete it.

Mist-Hunter:
Thank you for Trilium, it is one the best note taking apps I have ever seen, really awesome. 

I just want to add in here, that ```TRILIUM_DATA_DIR``` as seen in the [docker-compose.yml](https://github.com/zadam/trilium/blob/master/docker-compose.yml) doesn't seem to be supported by the script logic anywhere outside of /home/node

Looking at [start-docker.sh](https://github.com/zadam/trilium/blob/master/start-docker.sh)

```bash
#!/bin/sh

[[ ! -z "${USER_UID}" ]] && usermod -u ${USER_UID} node || echo "No USER_UID specified, leaving 1000"
[[ ! -z "${USER_GID}" ]] && groupmod -g ${USER_GID} node || echo "No USER_GID specified, leaving 1000"

chown -R node:node /home/node
exec su-exec node node ./src/www
```

[...]

Issue #1971, opened 2021-05-21
Docker Error: EACCES: permission denied, mkdir '/home/node/trilium-data/log'


beeyev:
Hello, 
I am trying to run trilium under docker, i use this command from the official documentation:
`docker run -d -p 0.0.0.0:8080:8080 -v ~/trilium-data:/home/node/trilium-data zadam/trilium:0.47-latest`

And after this i got this error, how can i fix that?
```
trilium_1  | internal/fs/utils.js:269
trilium_1  |     throw err;
trilium_1  |     ^
trilium_1  |
trilium_1  | Error: EACCES: permission denied, mkdir '/home/node/trilium-data/log'
trilium_1  |     at Object.mkdirSync (fs.js:921:3)
trilium_1  |     at Object.<anonymous> (/usr/src/app/src/services/log.js:7:8)
trilium_1  |     at Module._compile (internal/modules/cjs/loader.js:1015:30)
trilium_1  |     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1035:10)
trilium_1  |     at Module.load (internal/modules/cjs/loader.js:879:32)
trilium_1  |     at Function.Module._load (internal/modules/cjs/loader.js:724:14)
trilium_1  |     at Module.require (internal/modules/cjs/loader.js:903:19)
trilium_1  |     at require (internal/modules/cjs/helpers.js:74:18)
trilium_1  |     at Object.<anonymous> (/usr/src/app/src/app.js:1:13)
trilium_1  |     at Module._compile (internal/modules/cjs/loader.js:1015:30) {
trilium_1  |   errno: -13,
trilium_1  |   syscall: 'mkdir',
trilium_1  |   code: 'EACCES',
trilium_1  |   path: '/home/node/trilium-data/log'
trilium_1  | }```

zadam:
Hi, same issue as in https://github.com/zadam/trilium/issues/1960

ivosm:
I've have same/similar issue. The problem is the user "node" 
`/usr/src/app $ getent passwd node`
`node:x:1000:1000:Linux User,,,:/home/node:/bin/sh`
Which means, that  dir `~/trilium-data` needs permission for user with uid=1000  but  it has permission of the user the daemon runs. In my case root.
`drwxr-xr-x  2 root root     4096 kvě 24 19:42 trilium-data`
Quick workaround is setting permission for uid=1000 user but we are then in situation, when user x with uid=1001 has his data accessible by user y with uid=1000.
[...]

PR #2632, opened 2022-02-11
Fixed the docker compose configuration

teodor440:
In regards to the issue #1971, and using the solution presented in issue #2239 by the user wwwald, here's a configuration for docker-compose that will work and persist data across container restarts.


teodor440:
Just read the documentation and saw that the persistent data should be in /data directory and it can be configured to another directory via config.ini

PR #2697, opened 2022-03-01
Add x-trilium-version header

patrickjm:
Helpful for writing http clients for trilium

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -9,6 +9,7 @@ const FileStore = require('session-file-store')(session);
 const sessionSecret = require('./services/session_secret');
 const dataDir = require('./services/data_dir');
 const utils = require('./services/utils');
+const { version } = require('../package.json');
 require('./services/handlers');
 require('./becca/becca_loader');
 
@@ -22,6 +23,11 @@ app.use(helmet({
     hidePoweredBy: false, // errors out in electron
     contentSecurityPolicy: false
 }));
+
+app.use((req, res, next) => {
+    res.setHeader('x-trilium-version', version);
+    next();
+});
 
 app.use(express.text({limit: '500mb'}));
 app.use(express.json({limit: '500mb'}));


patrickjm:
At least in my case it was helpful outside of ETAPI. However I can add it there too if you like, assuming these changes don't already do that... I haven't checked

zadam:
Hi, TBH I don't really understand the use case. 

Also, the problem here is that version information should not be leaking to the public part, in this case login and shared pages, since that makes life easier for potential attackers.

patrickjm:
Hi @zadam, I can share my use case. I'm building a trilium client that, at least until ETAPI is fully featured, won't use ETAPI but instead the internal endpoints.

My understanding is that ETAPI is the only part of the trilium api that cares about backwards compatibility. So to ensure the client works for different versions of trilium, this header could hint at which contract to expect.

It's not unusual to put server version in headers. For example, Apache shares a header like this: `Server: Apache/2.4.10 (Debian)`

Nriver:
In production environment, you should avoid expose sensitive information like version number directly. Cause, if there happens to be a vulnerability on certain of the software version, the hacker will know you are vulnerable immediately.

Apache has a `ServerTokens Prod` config to hide info in production environment, and nginx has `server_tokens off;`.

So, if there is a need to get the version number, it would be better go through an authenticated method like ETAPI.

zadam:
Hi, I just added a new ETAPI method `/etapi/app-info` which returns the following:

```json
{
  "appVersion": "0.50.2",
  "dbVersion": 194,
  "syncVersion": 25,
  "buildDate": "2022-02-09T22:52:36+01:00",
  "buildRevision": "23daaa2387a0655685377f0a541d154aeec2aae8",
  "dataDirectory": "/home/user/trilium-data",
  "clipperProtocolVersion": "1.0",
  "utcDateTime": "2022-03-07T21:56:32.904Z"
}
```

patrickjm:
Awesome, thank you!

PR #2762, opened 2022-03-31
Solved the problem of multiple scroll bars in js frontend/backend

### LINKED ISSUE #2760: (Bug report) Multiple scroll bars in long js backend/js frontend notes
### Trilium Version

0.50.3

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 11 Pro 21H2

### Description

Long code notes with type "js backend" or "js frontend" have multile scroll bars. I think that's because of the additional buttons below text editor. There was no such problem in 0.49.5.
image

KhadishaKudys:
closes #2760

This PR solves the problem of multiple scroll bars in js code.

image


src/public/app/widgets/note_detail.js
```diff
@@ -32,7 +32,7 @@ const TPL = `
     }
     
     .note-detail.full-height {
-        height: 100%;
+        height: calc(100% - 3em);
```
zadam, line 35: This is too general since it affects also non-script type notes.

PR #2849, opened 2022-05-13
Bump base docker image for ARM stability

### LINKED ISSUE #2848: Latest docker images do not work on Raspberry Pi (and potentially other ARM-based devices) due to base image bug
This is not technically a bug with Trilium, but fixing will probably require updating the dockerfile.

I'm unclear on what the actual root cause of this issue is, but images based off of some versions of the alpine image fail to run on ARM-based devices (especially Raspberry Pi), with a fairly unhelpful "Unreachable Code" error message, probably due to bad/inconsistent architecture tags.

@axelstudios put together a useful list of alpine versions that do and do not exhibit this behavior here: https://github.com/nodejs/docker-node/issues/1589#issuecomment-970581391

chucklay:
Fixes #2848. Haven't thoroughly tested, but the image this creates is at least able to run, which the existing images can't.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # !!! Don't try to build this Dockerfile directly, run it through bin/build-docker.sh script !!!
-FROM node:16.14.2-alpine
+FROM node:17.1.0-alpine3.12
 
 # Create app directory
 WORKDIR /usr/src/app


zadam:
Hi, my concern is that 17.1 is actually quite old non-LTS release, thus probably containing security (and maybe other) issues. Isn't there some up to date version based on alpine 3.12?

chucklay:
Ah, true. I'll confess I'm not particularly familiar with Node development. Seems like the latest version to use Alpine 3.12 is 17.3.0, and the latest LTS version is `node:16.13-alpine3.12`. Not sure which (if either) of those would be better.

zadam:
Hmm, that's unfortunate. It looks like node.js is not supported anymore on alpine 3.12 and downgrading to an unsupported release seems quite unfortunate :-/

chucklay:
Disregard this, didn't read the node bug report closely enough. Manually updating libseccomp on my end was able to fix this.

PR #2864, opened 2022-05-21
Fix docker data permissions

dousha:
This change addresses the issue mentioned in #2856 where manual tweaking of managed docker volumes is discouraged.

By setting the data directory owner with root permission and running the application as user `node`, we no longer need to create a globally accessible directory without compromising application security.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,13 +18,19 @@ RUN set -x \
         nasm \
         libpng-dev \
         python3 \
+        wget \
     && npm install --production \
     && apk del .build-dependencies
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,13 +18,19 @@ RUN set -x \
         nasm \
         libpng-dev \
         python3 \
+        wget \
```
zadam: Is wget needed?

dousha: I think it's not needed. It's a residue of testing different "execute as another user" solutions where I have to fetch a binary from another repository.

PR #2928, opened 2022-06-18
Add a very basic docker health check

bill88t:
What the title says.
It checks if the webserver returns HTTP 200 OK on `/login`.
It will return unhealthy is the webserver is waiting for setup though..

### DIFF
--- a/DockerHealthcheck.sh
+++ b/DockerHealthcheck.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-if wget --spider -S "127.0.0.1:8080/login" 2>&1 | grep -w "HTTP/1.1 200 OK" ; then
+if wget --spider -S "127.0.0.1:8080/login" 2>&1 | awk 'NR==2' | grep -w "HTTP/1.1 200 OK" ; then
     exit 0
 else
     exit 1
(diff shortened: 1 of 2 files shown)


DockerHealthcheck.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+if wget --spider -S "127.0.0.1:8080/login" 2>&1 | grep -w "200" ; then
```
zadam: `wget` prints out all the headers and some other info. I think checking for just "200" word is not specific enough ...

zadam:
Hi, can you elaborate on this:

> It will return unhealthy is the webserver is waiting for setup though..

I tried a clean installation and the `/login` endpoint returns 200 even before the setup is finished ...

DockerHealthcheck.sh
bill88t: alright making it `HTTP/1.1 200 OK`

bill88t:
> I tried a clean installation and the /login endpoint returns 200 even before the setup is finished ...

running it on a new container for `127.0.0.1:8080/login` goes as follows:
image

echo $? returns the previous app's exit code.

Meanwhile for `127.0.0.1:8080/setup`:
image

The true outputs of the wget are:
image

As you can see, when the container has not finished setup, `/login` returns 302 instead of 200 OK.

DockerHealthcheck.sh
bill88t: Now it also checks for the `HTTP/1.1 200 OK` to be in the 2nd line.
It should always be at that line.

PR #2994, opened 2022-07-17
wip: Excalidraw Upgrade v0.11-v0.12

thfrei:
What: Upgrade of excalidraw from v0.11-v0.12

Main features:
- Eraser
- Double pinch while drawing?

Main fixes in regard to trilium:
- Render SVG free path backgrounds
- Should fix an issue with electron svg/image export?!

Problems:
- At the beginning I had the issue, that somehow canvas-notes were not longer of type canvas, but instead of type file. Does anyone else have this? When upgrading?

Next steps:
- I will do some more testing, would be glad if someone can also test.
- In the next couple of days, remove wip state.


Nriver:
Yes, there is one time my canvas note become a json file.

To recover, I create a new canvas note and export, then replace the json and reimport to get my canvas note back.

Do not know how to reproduce it.

thfrei:
I think it might actually have to do something with the sql migrations or something? maybe it is an "undfined" note type and there gets transformed?

Also: Is it a bug or a feature, that one cannot change the type for files?!

thfrei:
Maybe something here? 
https://github.com/thfrei/trilium/blob/e82fa5ee8279137ad64a7884b6c00d042dd9eb5c/src/services/consistency_checks.js#L320

Even though 'canvas' is part of the `noteTypes`: https://github.com/thfrei/trilium/blob/e82fa5ee8279137ad64a7884b6c00d042dd9eb5c/src/services/note_types.js#L1

thfrei:
Also, one can change it during runtime with SQL Console:

```
update notes set `type` = 'canvas' where noteId = 'xyz'
```

However, it seems that trilium server instance needs to be reloaded.

zadam:
@thfrei trilium sets note to "file" type if the type is not recognized. This could happen if you e.g. have server still running an older version of Trilium before canvas has been released. Or perhaps you ran the older version once on desktop by mistake. Something like that.

zadam (approved):

zadam:
Thanks!

PR #3000, opened 2022-07-19
Add config setting to disable update check

sigaloid:
Fixes https://github.com/zadam/trilium/discussions/2999

I was going to add this to the config.ini instead of options but I didn't find a way to access the config from the UI.

During initial setup, the user will be prompted at the password page to set whether Trilium Notes will automatically check for updates.
image
This way, if a user wants zero external connections, they can disable it from the beginning rather than having to open the UI (making that connection) to disable it.


zadam:
Hi, thanks for the contribution, but I removed the setting from the login page, I don't think it belongs there.

sigaloid:
I put it there to allow someone to preemptively disable network connections (otherwise it will check for updates by default when you open the app for the first time). But it might be out of place there, I agree

PR #3031, opened 2022-08-01
Run zip import in a transaction.

losfair:
On SQLite VFSes where transaction start and commit are high-latency operations, this improves ZIP import performance a lot. Should help under the default settings too, as transaction commit involves at least one `fsync`.

With [mvsqlite](https://github.com/losfair/mvsqlite) this reduces the time for importing `demo.zip` from tens of seconds to ~0.6sec.

Related: https://github.com/zadam/trilium/discussions/3029

### DIFF
--- a/src/services/import/zip.js
+++ b/src/services/import/zip.js
@@ -5,6 +5,7 @@ const utils = require('../../services/utils');
 const log = require('../../services/log');
 const noteService = require('../../services/notes');
 const attributeService = require('../../services/attributes');
+const sql = require("../../services/sql");
 const Branch = require('../../becca/entities/branch');
 const path = require('path');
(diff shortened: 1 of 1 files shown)


zadam:
Hi, in general having transactional import would be beneficial. But one significant problem is that this approach forces the import mechanism to first read the whole ZIP into memory and removes the streaming aspect from the equation. This will cause problems when importing large ZIPs.

zDEFz:
> Hi, in general having transactional import would be beneficial. But one significant problem is that this approach forces the import mechanism to first read the whole ZIP into memory and removes the streaming aspect from the equation. This will cause problems when importing large ZIPs.

Can't you just add more swap then or memory?

PR #3079, opened 2022-08-21
Add search result count to search page

### LINKED ISSUE #3076: Provide an easy way to count notes (Feature request)
### Describe feature

I've searched the wiki but did not seem to figure out a way to count the number of notes using a search query, either in the search result page or as part of another note. Is there a way to do this directly, or do I have to use a programmatic way (which I'm not too sure how either)?

### Additional Information

_No response_

sigaloid:
Fixes #3076 
---
image

I'm not sure that inlining the CSS is the most ideal way to do it but here's how it looked non-centered: 
image

Regardless this is just a simple change; the styling is open to changes.

### DIFF
--- a/src/public/app/services/note_list_renderer.js
+++ b/src/public/app/services/note_list_renderer.js
@@ -147,7 +147,7 @@ class NoteListRenderer {
     /*
      * We're using noteIds so that it's not necessary to load all notes at once when paging
      */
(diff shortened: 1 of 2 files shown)


zadam:
IIRC my original problem with this was that it consumed extra line. I [integrated the count into pager](https://github.com/zadam/trilium/commit/78f8a375873c708262e4981e0baf45e26c1e5b03) which is more compact:

PR #3084, opened 2022-08-23
Add an option to disable WAL.

losfair:
Related: https://github.com/zadam/trilium/discussions/3029

SQLite variants/extensions like [mvSQLite](https://github.com/losfair/mvsqlite) and [LiteFS](https://github.com/superfly/litefs) do not support WAL. By adding a environment variable option to disable WAL we can get Trilium working on these systems.


zadam:
Hi, I'm hesitant to add config options for unsupported (and thus untested) use cases as Trilium officially supports only vanilla SQLite.

PR #3085, opened 2022-08-23
Make database path configurable.

losfair:
This makes it possible to configure the database path separately via the `TRILIUM_DB_PATH` environment variable.


zadam:
Hi, what's the point of this?

sharpeagle2022:
> 

I guess it will be sort of a multiuser mode under the same profile or the ability to split workspaces by each with its own database....

zadam:
@sharpeagle2022 that can be achieved already by changing `TRILIUM_DATA_DIR`

Issue #1426, opened 2020-11-16
request：unlinked references


sdlytcl:
“Unlinked References”  is a great feature in some applications。Hope trilium have. 
sorry for my english.

sdlytcl:
这个要求太高了吗？
Is this requirement too high?

Issue #1157, opened 2020-07-19
Split Screen And Unlinked References


marinasflorin:
I absolutely love trilium. Been using most of the note taking apps (I have quite a vast experience here).

There are two big competitors coming: Obsidian and Roam research. But I believe trilium can beat them both.

I would suggest two big features that will bring trilium to another league:

1. Split View -> In obsidian, you can add multiple screens, like you can split and rearrange tabs how you want. They are in the same window (trilium just creates new tabs)
2. Unlinked references -> this way it can help us find contextual notes easier so we can link them.

If you are interested, I can create a website for you and help you get more people to use trilium. It's such a game changer for me.

I thank you for doing this software 👍

zadam:
Hi, thanks for the nice words.

For 1. trilium also supports opening notes in a new window which you can then arrange as you wish.

2\. can you describe what you mean by unlinked references?

Engr-AllanG:
I too love Trilium and have also been checking out Obsidian. I just checked, and I have tested about 20 different note apps over the last 3-6 months. Trilium is my favorite and the most powerful imo, and Obsidian is second. I think there is a paid market for your product. 

Unlinked references in Obsidian show up in a sidebar and lists places the note title appears elsewhere, but isn't actually linked. It is basically an automatic search function on the note title.

zadam:
Oh, ok, in that case Trilium already sort of have this functionality - there's a widget in the sidebar called "Similar notes" which lists notes with fuzzy-similar note title ...

marinasflorin:
Hi Zadam.

Indeed, you cam split the screen in a new draggable window. But this feature should be INSIDE the trilium window: If you know Notepad++, you can split the text editor in 2. Two windows inside the same window container. This enables you and us to easily read and write notes.

P.S. sent you a mail 👍

fivestones:
[...]

PR #3105, opened 2022-09-01
Unlinked reference

charlesdagenais:
Hi,
This is a first attemp to respond to the request of _Unlinked references_ feature and nowhere to be complete.
feature mentionned in issue #1426 and #1157 
Goal is to help find contextual notes easier to create internal links.

Maybe using the search api isn't the right way to do it (too limited to filter the results ?


zadam:
Hi, I'm not sure about the feature. It's probably not scalable enough to do it with normal search since it can take a long time if you have gigabytes of data.

PR #3139, opened 2022-09-17
Add another ETAPI method to the api spec file.

jph:
Hi @zadam,

I've added another method that's present in the API to the API spec file.

There may be more additions like this trickling in as I go - would you prefer me to make an issue and work on a bigger squashed commit with these changes later?

Another thing I'd like your opinion about is if you think it's worth having separate API endpoints for updating things like the note content.

IMO, it would be better to specify the Note schema to have a content property and then have the note patch method receive it as optional, and update if present. If this sounds reasonable to you I'll probably give it a go myself but I'll open an issue for that one instead of just sending PRs

Regards,
@jph

### DIFF
--- a/src/etapi/etapi.openapi.yaml
+++ b/src/etapi/etapi.openapi.yaml
@@ -245,6 +245,19 @@ paths:
             text/html:
               schema:
                 type: string
+    put:
+      description: Updates note content idenfied by its ID
+      operationId: putNoteContentById
+      requestBody:
+        description: html content of note
+        required: true
+        content:
+          text/plain:
+            schema:
+              type: string
+      responses:
+        '204':
+          description: note content updated
   /notes/{noteId}/export:
     parameters:
       - name: noteId


zadam:
> There may be more additions like this trickling in as I go - would you prefer me to make an issue and work on a bigger squashed commit with these changes later?

That depends, and I can't really answer in general. Follow your instinct, I'm usually not very strict on the form.

> Another thing I'd like your opinion about is if you think it's worth having separate API endpoints for updating things like the note content.

`content` is treated in a special way, it's not a regular property of the note entity. This is consistent with the way e.g. the script API is designed and also in how it's treated on a database level or in sync (content is stored separately from the rest of Note entity).

> If this sounds reasonable to you I'll probably give it a go myself but I'll open an issue for that one instead of just sending PRs

Besides what I mentioned above, the ETAPI is already published API and it is generally a bad idea to make such breaking changed without having a very good reason to do so since either you break existing clients or you need to support both new way and old way.

jph:
I understand, thanks for your input!

PR #3152, opened 2022-09-21
use same variable as Firefox for scrollbar thumb

spasche:
Hi,

I noticed that the scrollbar on WebKit browsers is rather hard to see since it's the same color as the text background (and the scrollbar border is very thin). This is especially noticeable with the light theme.

On Firefox, the thumb uses the --active-item-background-color, which makes it much more visible.
I'm not sure if that's intentional for WebKit? I would suggest to also use --active-item-background-color for WebKit, which I think greatly improves user experience by making the scrollbar more visible. Here's a screenshot comparing the existing style (left) with the proposed change (right).

trilium-scrollbar


zadam:
Thanks, I seem to have missed this PR.

PR #3213, opened 2022-10-21
Check both http and https in DockerHealthcheck

agentydragon:
I've been getting Docker reporting my Trilium container as unhealthy because wget was trying to talk HTTP to it while it was expecting HTTPS.


zadam:
Hmm, this is less than ideal, but I also don't see how to fix this easily. Probably still an improvement over the current state.

PR #3299, opened 2022-11-08
Added option to toggle line wrapping for Code notes

### LINKED ISSUE #813: Disable text wrapping
Hi

It seems like there is some inherint text wrapping in the text block area, evident in markdown and text. It would be nice if one could disable this. because it messes up indented blocks if the width of the text is wider than the text area, which makes it look like broken text blocks.

thanks

Banbury:
Fixes  #813
I've got frustrated with the line wrapping in code editors, as it makes code less readable (in particular for languages, where whitespace is meaningful like Python). So I've added a checkox to allow disabling line wrapping in the options dialog (Options/Code notes/Enable Linewrap).
The option is enabled by default, to keep the default behaviour.


zadam:
Thanks for this contribution!

Issue #3370, opened 2022-12-02
(Bug Report) Router not found for request /assets/v0.57.2/app-dist/setup.js


sp00ky:
### Trilium Version

Latest Docker Image

### What operating system are you using?

Other Linux

### What is your setup?

Server access only

### Operating System Version

Docker on Fedora 37 Server

### Description

Using docker and the latest docker-compose.yml.  Navigating to the app for the first time a blank screen loads.  The dev tools console gives a 404 for setup.js and navigating to setup.js gives me the following:

{"message":"Router not found for request /assets/v0.57.2/app-dist/setup.js"}

cyrsis:
+1 in here

aashish108:
Yeah me too, blank screen. Should not have upgraded lol.

st3iny:
Seems to be a duplicate of https://github.com/zadam/trilium/issues/3369.

sp00ky:
> Seems to be a duplicate of #3369.

Probably related.  However, I don't get the error referenced there.

5AMsan:
Same here

yllekz:
Same. Phew thought I screwed something up

Issue #3372, opened 2022-12-02
(Bug report) 0.57.2 Handhsake failed 


gerroon:
### Trilium Version

0.57.2

### What operating system are you using?

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 11, Linux

### Description

Hi

I have been using trilium with my server for couple years, I did many updates and uprgades without any issues. Today I upgraded my windows client and the linux trilium server and now I cant sync. 

I am getting etting the error below. I cant imagine this being a setup issue with mine since it was working well until this point.

I can also login to the server ui usig a browser, so definetely the server and the server url works. My passwork works in the browser

![Be2QPdlI7r](https://user-images.githubusercontent.com/8519469/205209803-3fcc9869-150c-4d1d-a2cd-b296336b581a.png)

gerroon:
Putting the direct Ip:Port in the client does not work either, so like I said I doubt that this is an issue with my setup.

ztankoz:
Same with docker.
Fresh install -> blank page on browser but no errors in logs
Downgrade to 56.2 works fine

DaveDavenport:
I get the following after updating:
```
Error: Router not found for request /assets/v0.57.2/app-dist/desktop.js

    at /usr/src/app/src/app.js:92:17

    at Layer.handle [as handle_request] (/usr/src/app/node_modules/express/lib/router/layer.js:95:5)

    at trim_prefix (/usr/src/app/node_modules/express/lib/router/index.js:328:13)

    at /usr/src/app/node_modules/express/lib/router/index.js:286:9

    at Function.process_params (/usr/src/app/node_modules/express/lib/router/index.js:346:12)

    at next (/usr/src/app/node_modules/express/lib/router/index.js:280:10)

    at Layer.handle [as handle_request] (/usr/src/app/node_modules/express/lib/router/layer.js:91:12)

    at trim_prefix (/usr/src/app/node_modules/express/lib/router/index.js:328:13)

    at /usr/src/app/node_modules/express/lib/router/index.js:286:9

    at Function.process_params (/usr/src/app/node_modules/express/lib/router/index.js:346:12) {

  status: 404

}
```

IPPrem:
[...]

Issue #3369, opened 2022-12-02
(Bug report) /assets/v0.57.2/app-dist/desktop.js net::ERR_ABORTED 404 (Not Found)


wangxinpier:
### Trilium Version

0.57.2

### What operating system are you using?

Other Linux

### What is your setup?

Server access only

### Operating System Version

synology nas

### Description

<img width="1366" alt="image" src="https://user-images.githubusercontent.com/1505835/205194211-b2fe0765-ca82-4a91-bf47-72ff6d99ab6d.png">
<img width="1177" alt="image" src="https://user-images.githubusercontent.com/1505835/205194281-2a1db031-3939-4686-9982-8e593fd65b9e.png">
<img width="1320" alt="image" src="https://user-images.githubusercontent.com/1505835/205194316-0ca5175d-fe44-46f7-8719-d5de13b3dec8.png">

yanickxia:
the same problem ……

derbirch:
same here, the macOS client works and can trigger a sync, but can not access it from the browsers
<img width="1159" alt="Screenshot 2022-12-02 at 11 35 17" src="https://user-images.githubusercontent.com/33411000/205209484-abe37b52-9433-479d-8002-b2ffb1379794.png">

zhang-stephen:
same here... could it be resolved with rolling back to 0.56.2?

yanickxia:
@zhang-stephen  rollback to 0.56.2, everything is OK

derbirch:
> @zhang-stephen rollback to 0.56.2, everything is OK

Rolled back too and it worked like a charm XD

st3iny:
It seems that bundled app code is missing entirely from the container:

```sh
11:04:19 [~/Code/trilium] git:(v0.57.2):dirty
richard@arch 130 $ docker run --rm -ti --entrypoint=sh zadam/trilium:0.57.2
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
/usr/src/app # ls -l src
total 20
-rw-r--r--    1 root     root           616 Dec  1 21:56 anonymize.js
-rw-r--r--    1 root     root          5007 Dec  1 21:56 app.js
drwxr-xr-x    1 root     root           162 Dec  1 21:57 becca
drwxr-xr-x    1 root     root           256 Dec  1 21:57 etapi
drwxr-xr-x    1 root     root            98 Dec  1 21:57 public
[...]

PR #3375, opened 2022-12-02
Fix missing webpack bundle generation in Docker container build

DynamoFox:
Some recent commit fixed missing webpack delivery in the Docker images, however the Dockerfile wasn't updated to generate the webpack bundle aswell and as a result no bundle could be found and shipped by Express.

Should fix: #3370 #3372 #3369


zadam:
Thanks!

PR #3406, opened 2022-12-11
Fix powershell exec policy

deajan:
Since Powershell won't allow script execution by default, let's inline Powershell commands with specified ExecutionPolicy.
Also, make command interpreter unicode aware using `chcp 65001` when Powershell is not available

Fixes https://github.com/zadam/trilium/issues/3398
Also updates https://github.com/zadam/trilium/issues/2327


zadam:
Thanks!

Issue #3409, opened 2022-12-12
(Bug report) Router not found for request /assets/v0.57.4/app-dist/share.js


sigaloid:
### Trilium Version

0.57.4

### What operating system are you using?

Other Linux

### What is your setup?

Local + server sync

### Operating System Version

Fedora 37

### Description

Sharing notes (specifically in this example, an excalidraw canvas note) does not work on v0.57.4 because of this 404.

DynamoFox:
Mhm, I think I understand the issue... But I have no idea why it worked in the first place with v0.57.3.
Also note that this should only affect only the Docker build.

PR #3410, opened 2022-12-12
Fix missing share.js in public/app-dist folder; Delete unused public/app folder (Docker build)

DynamoFox:
Hi, when I enabled the webpack bundle generation in the Docker build I wrongly assumed that share.js would have been bundled as well, well, it's not the case. Looking at copy-trilium.sh it seems like the file is just 'manually' copied to app-dist and then the app folder is wiped.
This PR introduces the same behavior as the copy-trilium.sh script in the Dockerfile.
Should fix: #3409

(Little side-note: Ultimately I think the Docker build should run the same scripts in the bin/ folder to generate Trilium's server edition, instead of replicating the same behavior manually in the Dockerfile which is easily error-prone... This would probably avoid more annoyances like the one this PR solves in the future.)


zadam:
Thanks for the fix.

> (Little side-note: Ultimately I think the Docker build should run the same scripts in the bin/ folder to generate Trilium's server edition, instead of replicating the same behavior manually in the Dockerfile which is easily error-prone... This would probably avoid more annoyances like the one this PR solves in the future.)

I agree, it is currently quite messy and error-prone.

PR #3467, opened 2022-12-28
Evernote import - use pageUrl attribute for enex source-url

contributor:
Currently, notes from Evernote enex file are imported with `sourceUrl` attribute for Evernote's `source-url` tag. It will make more sense to leverage Trilium's `pageUrl` attribute instead. `pageUrl` and `source-url` are direct equivalents in these two apps. Both allow clicking a link on UI to go to a note source web page. 

Of course, I can run sql query after import and rename attribute after each import. But I see no reason why no to use `pageUrl` attribute directly during enex import.

### DIFF
--- a/src/services/import/enex.js
+++ b/src/services/import/enex.js
@@ -115,7 +115,7 @@ function importEnex(taskContext, file, parentNote) {
             let labelName = currentTag;
 
             if (labelName === 'source-url') {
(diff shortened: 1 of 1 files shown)


zadam:
Thanks. The concept of `pageUrl` was added later than ENEX and this was not revisited.

PR #3479, opened 2023-01-03
Fix clipping selection can create multiple notes for the same Url

contributor:
Hi Zadam

This fixes a bug when you make multiple clippings on the page containing '&' char in url query string.
Fix just calls html_sanitizer.sanitize earlier (before finding clipping note)


contributor:
Alternative PR is https://github.com/zadam/trilium/pull/3480, but it has a npm dependency. Let me know which path do you prefer

PR #3480, opened 2023-01-03
Fix pageUrl and clipping selection can create multiple notes for the same Url

contributor:
This is (more correct) alternative of https://github.com/zadam/trilium/pull/3479 - it uses sanitize-url (intead not sanitize-html) for pageUrl)

But it requires additional npm package.


zadam:
Thanks!

PR #3481, opened 2023-01-03
Safe Vs Code debug launch

contributor:
vscode launch can break release app data, because it uses default data dir.

I got this problem by pressing F5 in vscode, then db migration started, then it couldn't work because of version conflicts in cluster. Electron app couldn't sync to server anymore, creating endless syncing with errors in log like here https://github.com/zadam/trilium/issues/1163

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,7 +9,8 @@
                 "<node_internals>/**"
             ],
             "env": {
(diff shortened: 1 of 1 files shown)


zadam:
Thanks, this file was user contributed, I've never actually used it.

contributor:
The feature to auto backup data before migration saved me :)

PR #3502, opened 2023-01-08
add senderInbox attribute handling

kooshi:


zadam:
Hi, instead of introducing new label, I think we can reuse the existing `#inbox`. I did that in https://github.com/zadam/trilium/commit/4ae1b6e07b810c7625222a02a155481a6e416e40

kooshi:
Sounds good, thanks!

PR #3544, opened 2023-01-20
Add #calendarType attribute to support monthly or weekly notes

AllanZyne:
Add new attribute `#calendarType` on calendar root note, its value can be `monthly`(default) or `weekly`.
If `#calendarType=monthly`, then the calendar tree will be organized as "year/month/day", this is default option just like before;
if `#calendarType=weekly`, then the calendar tree will be organized as "year/week/day".

Beside, year/month/week/day notes have their own title pattern: `yearPattern`(new), `monthPattern`, `weekPattern`(new) and `datePattern`, and they all support these macros:

- `{year}`: "YYYY"
- `{month}`: month name
- `{monthNumberPadded}`: "MM"
- `{weekNumber}`(new): ISO week number
- `{weekNumberPadded}`(new): "WW"
- `{weekDay}`: week name
- `{weekDay2}` the first two characters of week name
- `{weekDay3}`: the first three characters of week name
- `{dayInMonthPadded}`: "DD"
- `{isoDate}`: "YYYY-MM-DD"
(Add custom macros support?)

---

To implement the design of https://github.com/zadam/trilium/issues/3461, I added a new attribute #calendarPattern which can control the note tree of calendar.

The default value of #calendarPattern is "year/month/day", this will keep the default calendar tree like before.

The label in #calendarPattern can be one of "day", "week", "month" or "year", and there must be at least one "day" in it.

Thanks for any suggestions.

### DIFF
--- a/src/services/date_notes.js
+++ b/src/services/date_notes.js
@@ -64,24 +64,11 @@ function getRootCalendarNote() {
 }
 
 function getParentNote(label, dateStr, rootNote) {
-    const pattern = rootNote.getOwnedLabelValue("calendarPattern") || "year/month/day";
-    const labels = pattern.split("/")
-
-    {
-        let labelSet = new Set();
-        for (const label of labels) {
-            if (labelSet.has(label)) {
-                throw new Error(`"${label}" in #calendarPattern must be at most one`);
-            }
-            if (label != 'day' && label != 'week' && label != 'month' && label != 'year') {
-                throw new Error(`"${label}" in #calendarPattern should be one of "day", "week", "month" or "year"`)
-            }
-            labelSet.add(label);
-        }
-        if (!labelSet.has('day')) {
-            throw new Error('#calendarPattern must have "day"');
-        }
+    const type = rootNote.getOwnedLabelValue("calendarType") || "monthly";
+    if (type != "monthly" && type != "weekly") {
+        throw new Error('#calendarType should be "monthly" or "weekly"');
     }
+    const pattern = type == "monthly" ? "year/month/day" : "year/week/day";
 
     let parentLabel = "root";
     for (const currLabel of pattern.split("/")) {
@@ -112,87 +99,6 @@ function getNoteTitle(pattern, dateStr, options = {}) {
     return pattern.replace(/{(\w+)}/g, function(_, match) {
         switch(match) {
             case 'year':
-                console.log('year!!!!!', dateStr.substr(0, 4));
-                return dateStr.substr(0, 4);
-            case 'month':
-                return MONTHS[dateObj.getMonth()];
-            case "monthNumberPadded":
-                return dateStr.substr(5, 2);
-            case "weekNumber":
-                return dateUtils.getWeek(dateObj, startOfTheWeek);
-            case "weekNumberPadded":
-            {
-                let weekNum = dateUtils.getWeek(dateObj, startOfTheWeek);
-                if (weekNum <= 9)
-                    return "0" + weekNum;
-                return weekNum;
-            }
-            case "weekDay":
-                return weekDay;
-            case "weekDay3":
-                return weekDay.substr(0, 3);
-            case "weekDay2":
-                return weekDay.substr(0, 2);
-            case "dayInMonthPadded":
-                return dateStr.substr(8, 2);
-            case "isoDate":
-                return dateUtils.utcDateStr(dateObj);
-            default:
-                throw new Error(`unknown pattern {${match}}`);
-        }
-    });
-}
-
-function getParentNote(label, dateStr, rootNote) {
-    const pattern = rootNote.getOwnedLabelValue("calendarPattern") || "year/month/day";
-    const labels = pattern.split("/")
-
-    {
-        let labelSet = new Set();
-        for (const label of labels) {
-            if (labelSet.has(label)) {
-                throw new Error(`"${label}" in #calendarPattern must be at most one`);
-            }
-            if (label != 'day' && label != 'week' && label != 'month' && label != 'year') {
-                throw new Error(`"${label}" in #calendarPattern should be one of "day", "week", "month" or "year"`)
-            }
-            labelSet.add(label);
-        }
-        if (!labelSet.has('day')) {
-            throw new Error('#calendarPattern must have "day"');
-        }
-    }
-
-    let parentLabel = "root";
-    for (const currLabel of pattern.split("/")) {
-        if (currLabel === label) {
-            break;
-        }
-        parentLabel = currLabel;
-    }
-    switch (parentLabel) {
-        case "root":
-            return rootNote;
-        case "year":
-            return getYearNote(dateStr, rootNote);
-        case "month":
-            return getMonthNote(dateStr, rootNote);
-        case "week":
-            return getWeekNote(dateStr, rootNote);
-        case "day":
-            return getDayNote(dateStr, rootNote);
-    }
-    return null;
-}
-
-function getNoteTitle(pattern, dateStr, options = {}) {
-    const dateObj = dateUtils.parseLocalDate(dateStr);
-    const weekDay = DAYS[dateObj.getDay()];
-    const startOfTheWeek = options.startOfTheWeek || "monday";
-    return pattern.replace(/{(\w+)}/g, function(_, match) {
-        switch(match) {
-            case 'year':
-                console.log('year!!!!!', dateStr.substr(0, 4));
                 return dateStr.substr(0, 4);
             case 'month':
                 return MONTHS[dateObj.getMonth()];
@@ -239,12 +145,14 @@ function getYearNote(dateStr, rootNote = null) {
     }
 
     const pattern = rootNote.getOwnedLabelValue("yearPattern") || "{year}";
-    const noteTitle = getNoteTitle(pattern, dateStr);
     const parentNote = getParentNote("year", dateStr, rootNote);
+
     if (!parentNote) {
         let dateObj = dateUtils.parseLocalDate(dateStr);
         return getDayNote(dateUtils.utcDateTimeStr(dateObj.setMonth(0).setDate(1)), rootNote);
     }
+
+    const noteTitle = getNoteTitle(pattern, dateStr);
 
     sql.transactional(() => {
         yearNote = createNote(parentNote, noteTitle);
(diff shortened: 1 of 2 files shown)


src/services/date_notes.js
```diff
@@ -62,6 +63,86 @@ function getRootCalendarNote() {
     return rootNote;
 }
 
+function getParentNote(label, dateStr, rootNote) {
+    const pattern = rootNote.getOwnedLabelValue("calendarPattern") || "year/month/day";
```
zadam: This `calendarPattern` suggests a great deal of flexibility, but what are actually reasonable structures? I can think of just two:

* `year/month/day` (existing)
* `year/week/day` - new

Are there any other? You could imagine e.g. `year/month/week/day`, but I'm not sure how useful that would be. The syntax also allows weird things like `month/year/day` which probably make no sense at all.

In my experience it's often better to provide well-defined options, in this case e.g. `#calendarType=monthly` and `#calendarType=weekly` which force a specific structure.

What do you think?

src/services/date_utils.js
```diff
@@ -68,6 +68,43 @@ function getDateTimeForFile() {
     return new Date().toISOString().substr(0, 19).replace(/:/g, '');
 }
 
+/**
+ * Modify from https://stackoverflow.com/a/9047794
+ * Returns the week number for this date
+ */
+function getWeek(date, startOfTheWeek) {
```
zadam, line 75: This returns ISO week, but I wonder if this is the best choice for "human consumption". For me, it's kinda surprising to see 1. 1. marked as "week 53 of the previous year".

src/services/date_notes.js
AllanZyne: Yes, you're right, using `#calendarType=monthly|weekly` will give a more reasonable fixed structure.
Sometimes, I want to omit "year" notes to make journal structure more flatten, and when new year is arrived, I'll move and archive them in another directory and create a new one.

In fact, if I have a chance to redesign the journal function, I would like to make its structure invisible, and the day notes can only be accessible via the calendar widget or apis. Because in the case of contained weekly notes, it's difficult to represent the journal in a tree structure.
But maybe Trilium doesn't want to make journal opaque. So I'll organize the journal by this way: all year notes in "years" folder, all month notes in "month" folder, ... and all day notes in "days" folder, and uses "note-detail-pane" or "right-pane" to display its children. 

What do you think of this idea?

src/services/date_utils.js
AllanZyne: Yes, it's really odd but have to...

src/services/date_notes.js
zadam: > So I'll organize the journal by this way: all year notes in "years" folder, all month notes in "month" folder, ... and all day notes in "days" folder, and uses "note-detail-pane" or "right-pane" to display its children.

But what benefit would this bring? Year-month-day or Year-week-day is a natural hierarchy so it seems fitting to represent it as such in a tree structure.

src/services/date_utils.js
zadam: But we don't have to use ISO weeks. There are other week numbering systems.

For me, it makes sense to say that January 1st is always the first week, no matter which day of the week it is. But I'm open to discussion on this point.

src/services/date_notes.js
AllanZyne: In my opinion, Year-week-day is not as natural as Year-month-day structure.

Weekly note is a little embarrassing when it comes across the end of month or year, but it'd be really useful to be an independent note, rather than the note of the first day of week.

That idea comes from Obsidian's [Periodic Notes](https://github.com/liamcain/obsidian-periodic-notes), which can set the note folder of daily/weekly/monthly/quarterly/yearly notes separately. (BTW, in some companies, they use two weeks as a [sprint](https://www.atlassian.com/agile/scrum/sprints).)

Therefore, I think we can treat the calendar as the database of all kinds of periodic notes, and the hierarchical structure is managed by the calendar widgets or other panels. And this will maximize the flexibility of calendar notes.

src/services/date_utils.js
AllanZyne: Then you will feel strange that the Monday of 2023-W1 is 2022-12-26 >_<

src/services/date_notes.js
zadam: > In my opinion, Year-week-day is not as natural as Year-month-day structure.

That's why you will be able to choose.

> Weekly note is a little embarrassing when it comes across the end of month or year, but it'd be really useful to be an independent note, rather than the note of the first day of week.

I'm confused. Week should be an independent note, just created in a year-week-day hierarchy. It's a separate note which has (typically 7) children notes.

> Therefore, I think we can treat the calendar as the database of all kinds of periodic notes, and the hierarchical structure is managed by the calendar widgets or other panels. 

I don't think this is a good idea. Besides, since this is a major change, it would break expectations of existing users.

src/services/date_utils.js
zadam: Not necessarily:

* January 1st is the 1. week (no matter which weekday it falls on)
* December 31st is always 53rd week

It has the disadvantage that first and last week in the year may not have 7 days (may be as short as 1 day), but that's IMHO acceptable.

src/services/date_notes.js
AllanZyne: Yes, this will be a breaking change.

In summary, you prefer to use `#calendarType=monthly|weekly`, and for the number of weeks, you also prefer to give an option in ISO week or other numbering system. Am I right?

The reason why I want to use that breaking structure is that, I really hope the weekly notes can coexist with monthly notes. Even for `#calendarPattern`, users can use `#calendarPattern=year/month/week/day` to achieve that.

But `#calendarType` is really interesting, it means this attribute can be other values in the future. Anyway, I'm glad to make contribution on the calendar feature.

src/services/date_utils.js
AllanZyne: IMHO, I don't think a week which has less than 7 days is a good idea. This maybe need a discussion with more people?

src/services/date_notes.js
zadam: > you prefer to use #calendarType=monthly|weekly

Yes.

> and for the number of weeks, you also prefer to give an option in ISO week or other numbering system. Am I right?

Perhaps in the future, but in the beginning I believe one numbering system is enough. I'm just questioning whether ISO is the one which should be implemented in the beginning.

src/services/date_utils.js
zadam: > IMHO, I don't think a week which has less than 7 days is a good idea. This maybe need a discussion with more people?

It's certainly not great, but personally, I don't think it's that bad if it affects only the days around the new year.

I agree that input from other people would be good. I guess it heavily depends on how people use it - specifically if they need some alignment with numbered weeks from perhaps their work or some other applications etc. ISO numbering is a de-iure standard, but that doesn't necessarily mean that it's in widespread use.

AllanZyne: Personally speaking, I don't care about the week numbering(or week note id), because I can always change the title of the week notes. In our daily life, we most likely use the term "this week"/"next week"/..., and don't combine "week" with "year". Therefore, I don't think use ISO week is a problem.

But I still feel that "year/week/day" is not a good structure for weekly.

AllanZyne: How about this idea? 

When the first day of year is not the start of week, for example 2023, we create a week note titled "W53 | W1" which will appear under both "2022" and "2023", and the "2023"'s note is a cloned note.

If we want to align with ISO week at the same time, its title can be "W53 | W0".

zadam: @AllanZyne leveraging cloning actually sounds like a good idea. Reflecting on the previous discussion, I think the biggest problem with the ISO numbering was that the first days of the year would hierarchically appear under the preceding year, but this solution would fix this. So I would say, yes, let's do it like this.

AllanZyne: @zadam Great! I'll implement it ASAP.

AllanZyne:
Hi, @zadam. In backend api document, BNote's `cloneTo(parentNoteId) → {Object}`, what's object?

AllanZyne:
I think my logic might be right, but I encounter the following error message, would you like to give me some helps? Thank you very much!

"Encountered error "Process exceeded time limit 30000", check out the console."

zadam:
> Hi, @zadam. In backend api document, BNote's `cloneTo(parentNoteId) → {Object}`, what's object?

Unfortunately JSDoc does not support object declaration in `@returns`, the value is better defined in the [source](https://github.com/zadam/trilium/blob/master/src/becca/entities/bnote.js#L1365).

zadam:
> "Encountered error "Process exceeded time limit 30000", check out the console."

This usually means that the WebSocket message (with entity updates) has crashed some component in the frontend. Usually you will find the source of the problem above in the console.

AllanZyne:
> > "Encountered error "Process exceeded time limit 30000", check out the console."
> 
> This usually means that the WebSocket message (with entity updates) has crashed some component in the frontend. Usually you will find the source of the problem above in the console.

Backend logs:

```
getParentNote day 2022-12-28
getWeekNote: find weekNote
getWeekNote: not year
getParentNote week 2022-12-28
getParentNote year 2022-12-28
Created new note 'szPqG5zh7kkj', branch 'u7dTta9kSl90_szPqG5zh7kkj' of type 'text', mime 'text/html'
getWeekNote: clone
Cloned note 'y9mnzt2gwt5E' to new parent note 'szPqG5zh7kkj' with prefix 'undefined'
getWeekNote: return
Created new note 'HMx7UmZwNJV9', branch 'y9mnzt2gwt5E_HMx7UmZwNJV9' of type 'text', mime 'text/html'
Slow 200 GET /api/special-notes/days/2022-12-28 with 301 bytes took 23ms
JS Error: Encountered error Process exceeded time limit 30000: timeLimit@http://127.0.0.1:8081/assets/v0.59.1/app/services/utils.js:293:28
@http://127.0.0.1:8081/assets/v0.59.1/app/services/ws.js:169:34
asyncFunctionResume@[native code]
@http://127.0.0.1:8081/assets/v0.59.1/app/services/ws.js:90:60
asyncFunctionResume@[native code]
@http://127.0.0.1:8081/assets/v0.59.1/app/services/ws.js:114:36
asyncFunctionResume@[native code], reloading frontend.
Stack: logError@http://127.0.0.1:8081/assets/v0.59.1/app/services/ws.js:24:29
@http://127.0.0.1:8081/assets/v0.59.1/app/services/ws.js:172:21
asyncFunctionResume@[native code]
@[native code]
promiseReactionJobWithoutPromise@[native code]
promiseReactionJob@[native code]
200 POST /api/tree/load with 7811 bytes took 1ms
```

Reproducer:

1. Delete 2022 & 2023
2. Create 2023-01-01 from calendar widget
3. Create 2022-12-28 from calendar widget

Then you will timeout at `frocaUpdater.processEntityChanges(nonProcessedEntityChanges)`.
But if you reload the web page, then "2022/W0*/2022-12-28" will be displayed.

AllanZyne:
I don't need this feature anymore. If someone needs it, glad to let me know.

PR #3590, opened 2023-02-04
Fix broken reference consistency check

eliandoran:
I have five notes which keep showing up under the root note.

```
Consistency issue fixed: Branch '8yc5SDQRrhss_4aKhk3oLjfnx' was set to root parent since it was referencing missing parent
note '8yc5SDQRrhss'
Consistency issue fixed: Branch 'g0UdmDhnHNiH_16FHE0bZt3pX' was set to root parent since it was referencing missing parent
note 'g0UdmDhnHNiH'
Consistency issue fixed: Branch 'g0UdmDhnHNiH_1PhkrImWtTcA' was set to root parent since it was referencing missing parent
note 'g0UdmDhnHNiH'
Consistency issue fixed: Branch 'g0UdmDhnHNiH_6QnjJ4hWGqRo' was set to root parent since it was referencing missing parent
note 'g0UdmDhnHNiH'
Consistency issue fixed: Branch 'sLHEyPuAseQD_vW4MhGZe1phE' was set to root parent since it was referencing missing parent
note 'sLHEyPuAseQD'
```

**Solution:**

If a branch was identified with a missing parent, the branch would be moved to root. However, the ID of the branch would be changed as a result of that, and this resulted in the creation of a new branch instead of updating the old one. Deleting the old one first solves the issue.

**Root cause analysis:**

Most likely broken by 0758c82983 which "forces" the ID to `parent_note` format. As we were changing the parent branch, the ID was changed. This resulted in the old branch remaining indefinitely. 

**Full analysis:**

[A few notes keep appearing in the root at start-up.pdf](https://github.com/zadam/trilium/files/10609863/A.few.notes.keep.appearing.in.the.root.at.start-up.pdf)

### DIFF
--- a/src/services/consistency_checks.js
+++ b/src/services/consistency_checks.js
@@ -148,13 +148,20 @@ class ConsistencyChecks {
                       AND notes.noteId IS NULL`,
             ({branchId, parentNoteId}) => {
                 if (this.autoFix) {
-                    const branch = becca.getBranch(branchId);
-                    branch.parentNoteId = 'root';
-                    branch.save();
+                    // Delete the old branch and recreate it with root as parent.
+                    const oldBranch = becca.getBranch(branchId);
+                    const noteId = oldBranch.noteId;
+                    oldBranch.markAsDeleted("missing-parent");
+
+                    const newBranch = new Branch({
+                        parentNoteId: 'root',
+                        noteId: noteId,
+                        prefix: 'recovered'
+                    }).save();
 
                     this.reloadNeeded = true;
 
-                    logFix(`Branch '${branchId}' was set to root parent since it was referencing missing parent note '${parentNoteId}'`);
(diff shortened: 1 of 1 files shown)


zadam:
Hello, thank you for your detailed analysis. 

> The only downside with this consistency check ( findBrokenReferenceIssues ) is that it does not seem to treat the situation where it finds a branch
with a bad parent, but that note has some other correct branches.
> This is what happened here, the parent was removed by the sync at some point, but the branch (which was local to one of the devices) got stuck. In
this situation, removing the branch without recreating it at all would have worked, since the note was already placed in the right location.

That's correct and that was the case also before changes in 0.58.

I fixed that in https://github.com/zadam/trilium/commit/ade22ea8259591461f3a8f319fc225eabd636bce

Issue #2875, opened 2022-05-25
(Bug report) 


gzxiexl:
### Trilium Version

0.51.2

### What operating system are you using?

Ubuntu

### What is your setup?

Local + server sync

### Operating System Version

win11 and win10

### Description

Cell Properties - Align cell text to the center This setting seems to be invalid

XXXJetfireXXX:
Also seems to be a duplicate of #2424

Issue #2424, opened 2021-12-08
(Bug report) Table cell - vertical align to "middle" does not work


wwwald:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/zadam/trilium/issues) for a bug report that matches the one I want to file, without success.

### Trilium Version

0.48.7

### What operating system are you using?

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 10 Enterprise, 19042.1348 / Server OS: Ubuntu 20

### Expected Behavior

In a random table, perform these steps:
* select one or more table cells
* go to "Cell properties" via the contextual toolbar
* click the "middle" option for table cell vertical text alignment:
![image](https://user-images.githubusercontent.com/39531/145241078-76aedee8-771b-4054-a2e3-a873101696e4.png)

I'd expect the text to be vertically centered in the table cell.

### Actual Behavior

Instead of vertically centering, the text remains vertically top-aligned.

gdbb:
I met the same problem.

In trilium desktop client (macOS) and web client, the text is vertically top-aligned even the property is set as "middle".

<img width="433" alt="图片" src="https://user-images.githubusercontent.com/7486825/145524702-16675172-333f-44cc-a6f2-43bfb6afbdec.png">

When I open note externally, it shows the same.

<img width="297" alt="图片" src="https://user-images.githubusercontent.com/7486825/145524294-b75d286e-eb1b-4f40-9988-20b49e7a9f3a.png">

zadam:
Hi, this seems to be a problem in CKEditor, I described the details in https://github.com/ckeditor/ckeditor5/issues/10998

I could work around it by changing the default alignment to middle as CKEditor expects, but that's a (minor) breaking change.

eliandoran:
Solved in https://github.com/TriliumNext/Trilium/pull/3650.

PR #3650, opened 2023-02-26
Centered vertical align does not seem to work on merged table cells

eliandoran:
It appears that the “center” vertical align for cells in tables is not working, as the text is bottom-aligned:

Before:

image

After:

image

The fix was to override a style introduced by `bootstrap.min.css` in both the application style, as well as the print style. 

See attached for more information:

https://github.com/zadam/trilium/files/10834709/Centered.vertical.align.does.not.seem.to.work.on.merged.table.cells.pdf

### DIFF
--- a/src/public/stylesheets/print.css
+++ b/src/public/stylesheets/print.css
@@ -13,4 +13,11 @@
     .relation-map-wrapper {
         height: 100vh !important;
     }
+
+    .table thead th,
+    .table td, .table th {
+        /* Fix center vertical alignment of table cells */
+        vertical-align: middle;
+    }
+
 }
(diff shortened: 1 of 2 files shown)


eliandoran:
@zadam , I've just noticed that this issue was already discussed in #2424 and #2875 and the issue was also raised upstream in https://github.com/ckeditor/ckeditor5/issues/10998 for slightly more than a year.

In any case, although I agree that perhaps it's not best that `ckeditor` is based on the assumption that `vertical-align` is middle by default; it does seem to be the case:

image

In this case, adding `vertical-align` for the default value would mean that the `vertical-align` style is added to all the `<th>`, `<td>` elements, which would increase the HTML size.

Since `vertical-align: middle` seems to be a browser default, I would say it is the responsibility of the one who is changing the defaults to set them back in (in this case it got broken by bootstrap).

Let me know your opinion.

PR #3652, opened 2023-02-26
Infinite loop when searching for '' in Jump to note

eliandoran:
By typing two apostrophes (`'`) in “Jump to note”, the server freezes.

Reproduced in both 0.58.8 and latest master branch.

See [Analysis](https://github.com/zadam/trilium/files/10834981/Analysis.pdf).


zadam:
Somehow I missed this, thanks.

PR #3676, opened 2023-03-06
Allow arbitrary labels to be added via web clipper

jrao1:
This change adds handling of a `labels` property in the json body of `POST /api/clipper/notes` API, the value of the `labels` property is expected to be key/value pairs.

### DIFF
--- a/src/routes/api/clipper.js
+++ b/src/routes/api/clipper.js
@@ -75,7 +75,7 @@ function addClipping(req) {
 }
 
 function createNote(req) {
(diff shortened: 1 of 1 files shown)


zadam:
Hello, do you plan to add this feature to web clipper as well? On its own this PR won't do much.

jrao1:
Yes, I added the corresponding feature in web clipper in PR https://github.com/zadam/trilium-web-clipper/pull/49

Issue #3677, opened 2023-03-07
Every day generate a root log?(Bug report) 


holoword:
### Trilium Version

0.59.1

### What operating system are you using?

Windows

### What is your setup?

Local (no sync)

### Operating System Version

Ubuntu 20.04.4 LTS

### Description

After update to 0.59.1, trilium server generate a daily log belong to root, cause the server failing and exit, the error is same as #1971 , after chown the file belong, the server can restart and work correctly.

the configure of my trilium below:
```
version: '3'
services:
  trilium:
    container_name: otrilium
    image: zadam/trilium
    restart: always
    environment:
      USER_UID: 1001
      USER_GID: 1001
      LOG_FILE: ./trilium.log
      TZ: 'Asia/Shanghai'
      TRILIUM_DATA_DIR: /data
    volumes:
      - ./trilium-data/:/data
```

the daliy log of 2023.03.06 last min is ok 
```
23:57:56.169 200 GET /api/health-check with 15 bytes took 1ms
23:58:26.307 200 GET /api/health-check with 15 bytes took 0ms
23:58:56.448 200 GET /api/health-check with 15 bytes took 0ms
23:59:26.584 200 GET /api/health-check with 15 bytes took 1ms
23:59:56.722 200 GET /api/health-check with 15 bytes took 1ms
```
but after 00.00 throw the the permession error of new log file.

### Error logs

[trilium-2023-03-06.log](https://github.com/zadam/trilium/files/10905221/trilium-2023-03-06.log)

holoword:
I am try to set docker user_uid and gid to 0 as a root user, but cannot running, it's pretty wired the docker need root permission to run but cannot set running user as root, at the same time, the docker container create a root permission log sounds like the user_id gid not correct work as we expect!
but if i delete the log file, just restart the docker, the container create the right permission log file as the set uid gid, so the problem maybe in the time everyday create new log.

Matapii:
Same issue here. Likely the health_check is creating the log file with root as owner and insufficient permissions for the node user to update. Workaround: Disable the health check in the docker-compose file as documented here: https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

holoword:
> Same issue here. Likely the health_check is creating the log file with root as owner and insufficient permissions for the node user to update. Workaround: Disable the health check in the docker-compose file as documented here: https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

[...]

PR #3685, opened 2023-03-08
“Docker non-root healthcheck”

holoword:
fix the bug "Every day generate a root log?(Bug report)" #3677
by set the healthy check run as user node.


zadam:
Thanks!

PR #3695, opened 2023-03-10
add natural sort for notes, introduce new label #sortNatural

Nriver:
related issue https://github.com/zadam/trilium/issues/3690

### DIFF
--- a/src/public/app/widgets/dialogs/sort_child_notes.js
+++ b/src/public/app/widgets/dialogs/sort_child_notes.js
@@ -65,6 +65,17 @@ const TPL = `<div class="sort-child-notes-dialog modal mx-auto" tabindex="-1" ro
                             sort folders at the top
                         </label>
                     </div>
+                    
+                    <br />
+                    
+                    <h5>Natural Sort</h5>
+
+                    <div class="form-check">
+                        <label class="form-check-label">
+                            <input class="form-check-input" type="checkbox" name="sort-natural" value="1">
+                            sort by alphanumeric order
+                        </label>
+                    </div>
                 </div>
                 <div class="modal-footer">
                     <button type="submit" class="btn btn-primary">Sort <kbd>enter</kbd></button>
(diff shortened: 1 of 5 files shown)


src/public/app/widgets/dialogs/sort_child_notes.js
```diff
@@ -65,6 +65,17 @@ const TPL = `<div class="sort-child-notes-dialog modal mx-auto" tabindex="-1" ro
...
                     </div>
+                    
+                    <br />
+                    
+                    <h5>Natural Sort</h5>
+
+                    <div class="form-check">
+                        <label class="form-check-label">
+                            <input class="form-check-input" type="checkbox" name="sort-natural" value="1">
+                            sort by alphanumeric order
```
zadam: I think this needs some more explanation. The default sorting is also technically "alphanumeric". Something like "language rule respecting sorting". Not sure how to phrase it.

src/services/tree.js
```diff
@@ -153,7 +153,14 @@ function sortNotes(parentNoteId, customSortBy = 'title', reverse = false, folder
             }
 
             function compare(a, b) {
-                return b === null || b === undefined || a < b ? -1 : 1;
+                if (!sortNatural){
+                    // alphabetical sort
+                    return b === null || b === undefined || a < b ? -1 : 1;
+                } else {
+                    // alphanumeric sort, or natural sort
+                    return a.localeCompare(b, undefined, {numeric: true, sensitivity: 'base'});
```
zadam: I see a problem with not defining locale here - apparently in that case the system locale will be used, but that can be different than desired. Especially the problem can arise when you have a web version running on some linux host with locale you can't change - then your local desktop instance will "fight" with the server version over the sort order.

I'm thinking that for "permanent sorting" - by label - you should be able to define the locale. Maybe then rename the label to something like `#sortLocale=de`. For one time sorting I think it's acceptable without explicit locale, because then it makes sense it's done according to locale where you're executing the action.

zadam (changes requested):

src/public/app/widgets/dialogs/sort_child_notes.js
Nriver: I changed it into "sort with respect to different character sorting and collation rules in different languages or regions."

src/services/tree.js
Nriver: You concern makes sense. I've added the `#sortLocale` label.

PR #3736, opened 2023-03-18
Fix typo in backend API createOrUpdateLauncher()

dymani:
Fixed two typos in `createOrUpdateLauncher()` of `backend_script_api.js`.

1. **Line 478:** `becca.getNote(opts.id)` should be `becca.getNote(noteId)`

The parameter note ID should include the `"al_"` prefix defined a few lines before, otherwise the API would repeatedly create new launcher notes even when there exists one with matching launcher ID.  

The existing problem is even severer when a particular launcher is to be updated from visible to invisible (or vice versa). Then, the API essentially creates two notes with the same "al_XXXX" ID under the Available Launchers and the Visible Launchers branches.  

2. **Line 517:** `launcherNote.removeLabel('keyboardShortcut');` should be `launcherNote.removeLabel('iconClass');`

For properly updating the launcher icon.

### DIFF
--- a/src/services/backend_script_api.js
+++ b/src/services/backend_script_api.js
@@ -475,7 +475,7 @@ function BackendScriptApi(currentNote, apiParams) {
         const noteId = 'al_' + opts.id;
 
         const launcherNote =
-            becca.getNote(opts.id) ||
+            becca.getNote(noteId) ||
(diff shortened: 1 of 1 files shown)


zadam:
Thanks!

PR #3773, opened 2023-03-26
Add svg icons

mm21:
Hello,

I have a larger monitor so I often zoom the page to around 150%. I've found everything in the UI scales nicely except the icon, so figured I'd try cropping and coloring the icon.svg and I'm pretty happy with the result.

Before:

image

After:

image

It's nice being able to see the details in the leaves!


Nyar233:
image
Perhaps the gray logo also needs to do the same thing.

Issue #3689, opened 2023-03-08
(Bug report) Menu has no border


meichthys:
### Trilium Version

0.59.1

### What operating system are you using?

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 10

### Description

The menu (in dark mode) has a very dark boarder. In dark mode it can be hard to differentiate the menu with nodes in the note tree - especially on screens with lower contrast ratios:
![image](https://user-images.githubusercontent.com/10717998/223822233-2bc399b9-1531-44e5-b98a-429027ba996b.png)

eliandoran:
Should have been fixed with https://github.com/TriliumNext/Trilium/pull/3777.

PR #3777, opened 2023-03-28
change downdown border to light border

PhilippOesch:
I tried out a light-border color. This would fix the problem in #3689


zadam:
I think I missed this.

PR #3805, opened 2023-04-06
fix: disable left click when context menu opening

soulsands:
GIF 2023-4-6 13-08-20
In desktop setup, the main context note will change following an click of the context menu occasionally, especially in an low-performance compute. You can use a cpu slowdown simulation in performance panel to reproduce it.
It works totally fine on webpage version.
The context menu has stopped propagation, but force-graph plugin still captured the click event. 
Currently an workaround is moving mouse pointer away to stop hovering a note before clicking an context menu item.
This is quite rough solution, just trying to provide some infomation to make trilium better.🙂

### DIFF
--- a/src/public/app/menus/context_menu.js
+++ b/src/public/app/menus/context_menu.js
@@ -5,6 +5,8 @@ class ContextMenu {
         this.$widget = $("#context-menu-container");
         this.dateContextMenuOpenedMs = 0;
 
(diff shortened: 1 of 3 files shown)


zadam:
Hello, thanks for the report. I've fixed in a shorter hack here: https://github.com/zadam/trilium/commit/e871edc8f347e6add552abe67d94a68793558802

PR #3808, opened 2023-04-06
Fix: correct note-map resizing

soulsands:


zadam:
Thanks. I removed the fallback handler, `ResizeObserver` is supported in all modern browsers.

PR #3812, opened 2023-04-07
Fix: disable attribute updating when composing input

soulsands:
GIF 2023-4-7 12-01-33

Keyup events should be ignored in some languages using IME like Chinese, Japanese.

BTW, it's better to share vscode formatting config in github repo that we can avoid formatting change


zadam:
Hi, I tried a different approach and I think this is a simpler fix: https://github.com/zadam/trilium/commit/293573a0cd759f43966f892c178b76245f7d3456

soulsands:
Sure. I didn’t know the isComposing property of input event and confused about using keyup event.

PR #3816, opened 2023-04-08
fix:don't break words for CJK text

soulsands:
Snipaste_2023-04-08_14-53-39

Hi, please consider sharing an format config.


zadam:
I'm using the default IntelliJ formatting.

PR #3848, opened 2023-04-15
chore: dev auto reload

soulsands:
It leverages internel websocket to support frontend auto reloading and uses nodemon to enable backend auto reloading, which ignores frontend files(src/public), generally only reloads when necessarily.
It works fine in server developing, but server reloading not available in electron.


zadam:
I'm not sure esp. on the backend reloading, but let's give it a try.

PR #3862, opened 2023-04-21
perf: minify css

soulsands:
Of little benefit now though


zadam:
Honestly not sure if it's worth it to complicate the build process here.

soulsands:
> Honestly not sure if it's worth it to complicate the build process / expand tooling here.

Me too actualy. Maybe we can leave it open until trilium can benefit from it obviously. It may be useful in the future.

PR #3870, opened 2023-04-23
perf: improve descendant count performance

soulsands:
Hi, I tested it gives the same result as before.

### DIFF
--- a/src/routes/api/note_map.js
+++ b/src/routes/api/note_map.js
@@ -27,8 +27,8 @@ function buildDescendantCountMap(noteIdsToCount) {
 
         return noteIdToCountMap[noteId];
     }
-    noteIdsToCount.forEach((nodeId) => {
-        getCount(nodeId);
+    noteIdsToCount.forEach((noteId) => {
+        getCount(noteId);
     });
 
     return noteIdToCountMap;
--- a/src/routes/api/note_map.js
+++ b/src/routes/api/note_map.js
@@ -3,9 +3,12 @@
 const becca = require("../../becca/becca");
 const { JSDOM } = require("jsdom");
 const NotFoundError = require("../../errors/not_found_error");
+function buildDescendantCountMap(noteIdsToCount) {
+    if (!Array.isArray(noteIdsToCount)) {
+        throw new Error('noteIdsToCount: type error');
+    }
 
-function buildDescendantCountMap() {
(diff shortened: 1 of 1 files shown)


src/routes/api/note_map.js
```diff
@@ -24,12 +27,12 @@ function buildDescendantCountMap() {
 
         return noteIdToCountMap[noteId];
     }
-
-    getCount('root');
+    noteIdsToCount.forEach((nodeId) => {
```
zadam: It's `noteId`, note `nodeId`.

src/routes/api/note_map.js
```diff
@@ -3,9 +3,12 @@
...
 const { JSDOM } = require("jsdom");
 const NotFoundError = require("../../errors/not_found_error");
+function buildDescendantCountMap(noteIdsToCount) {
+    if (!Array.isArray(noteIdsToCount)) {
+        throw new Error('noteIdsToCount: type error');
+    }
 
-function buildDescendantCountMap() {
-    const noteIdToCountMap = {};
+    const noteIdToCountMap = Object.create(null);
```
zadam, line 11: Is there some particular reason to use this instead of `{}`?

soulsands: It just might be a better choice for map semanticly as it's pure empty object, and also it imporves performance in theory.  https://stackoverflow.com/questions/26961544/use-cases-of-object-createnull 

Using new Map maybe even better, But not sure what the cost of being converted to a object, and it would make more code changes.

src/routes/api/note_map.js
soulsands: fixed

src/routes/api/note_map.js
zadam: IMHO the theoretical improvement in performance does not justify that it's very non-idiomatic construct in JavaScript.

soulsands: It’s not that usual, but used by some fundamental frameworks like vue, https://github.com/vuejs/vue/blob/49b6bd4264c25ea41408f066a1835f38bf6fe9f1/src/shared/util.ts#L114.  I thought it’s fine.

zadam: In frameworks it often makes sense to do micro optimizations, the priorities are often different from application code.

PR #3897, opened 2023-05-03
zip export/import improvements

contributor:


src/services/import/zip.js
```diff
@@ -240,12 +240,10 @@ async function importZip(taskContext, fileBuffer, importRootNote) {
             return /^(?:[a-z]+:)?\/\//i.test(url);
         }
 
+        content = removeTrilumTags(content);
+
         content = content.replace(/<h1>([^<]*)<\/h1>/gi, (match, text) => {
-            if (noteTitle.trim() === text.trim()) {
```
zadam, line 244: I would still leave this as a fallback for cases where the HTML files are not coming from Trilium. In such cases this heuristic can filter useless H1s, even if it's not perfect ...

contributor: Ok, brought the line back and updated my PR. This will also make export format backward compatible with previous version (without data-trilium-* atrributes).

I think in the future there should be a warning after import process finished if export version changed, that says something like: hey, you imported zip from older version of trilium, look at imported notes and make sure that everything is imported as expected.

PR #3900, opened 2023-05-04
Allow update of attribute fields isInheritable, position; enable generated etapi client

mm21:
This PR allows etapi clients to update attribute fields isInheritable and position. I'm developing a Python library for working with Trilium and this will enable idiomatic management of attributes: the user simply maintains a Python list of attributes, and positions are automatically calculated. Therefore this is the most important part of the PR; the others I can work around, even if I have to hack up generated code.

The other changes enable etapi clients to be generated using openapi-generator. I've generated the python-nextgen client and verified all CRUD operations for notes, attributes, and branches. 

The most sensitive change is adding attributeId/branchId path parameters to postAttribute/postBranch. I realize this may unfortunately break existing clients, but this is a requirement for OpenAPI - all methods need to take the same path parameters. See [https://swagger.io/docs/specification/describing-parameters/#path-parameters](https://swagger.io/docs/specification/describing-parameters/#path-parameters):

> Also remember to add required: true, because path parameters are always required.

The etapi.openapi.yaml changes are to fix validation errors:

- Required fields of Attribute and Branch removed: these schemas are also used for patch wherein those fields can't be updated; generated model then fails validation when it's instantiated
- Pass charset in response content-type: the content-type in Trilium's response is used as a dictionary key, so it needs to match exactly what's provided in the yaml
- Python's built-in regex module doesn't recognize unicode property escapes (\p{L}, \p{N}), so replaced them with more standard character classes (this wouldn't be a big deal for me to just change locally)
- - Wondering if this should be less strict to be safe, say [^\s]+
- Allow underscore in EntityId for branches
- Allow negative timezone offsets

For reference, here's the command I used to generate a Python client: (many other languages are supported)

`openapi-generator-cli generate -i etapi.openapi.yaml -g python-nextgen -o ./etapi-client --additional-properties packageName=etapi_client --additional-properties projectName=etapi-client`

### DIFF
--- a/src/etapi/attributes.js
+++ b/src/etapi/attributes.js
@@ -17,7 +17,8 @@ function register(router) {
         'type': [v.mandatory, v.notNull, v.isAttributeType],
         'name': [v.mandatory, v.notNull, v.isString],
         'value': [v.notNull, v.isString],
-        'isInheritable': [v.notNull, v.isBoolean]
+        'isInheritable': [v.notNull, v.isBoolean],
+        'position': [v.notNull, v.isInteger]
     };
 
     eu.route(router, 'post' ,'/etapi/attributes', (req, res, next) => {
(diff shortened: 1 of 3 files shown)


mm21:
I'll go ahead and split this into 2 PRs and then close this one.

Issue #3882, opened 2023-04-28
(Feature request) customize search engine


Nriver:
### Describe feature

Today, a user complained to me that the default search engine in Trilium is duckduckgo, which is blocked in China. It would be beneficial if we could customize the search engine to use alternatives such as bing or baidu for searching messages.

Also, I believe it would be intriguing to send the message to my personal ChatGPT bot. Given its impressive capabilities, connecting Trilium and ChatGPT in some manner would be a fascinating prospect.

zadam:
A configurable search engine is possible, but perhaps not the highest priority. Should be relatively easy as a contribution, though.

Basically there would be a new "options" row and UI in Options where you can define a template like `https://duckduckgo.com/?q={keywords}` with perhaps few predefined options.

Regarding ChatGPT, such integration seems very vendor specific. It looks more like a use case for custom widgets / scripting.

sigaloid:
> I'm confused. We can search the internet from within Trilium? How do we do that?

In the desktop app, you can select text and it suggests to search in a search engine. 

![image](https://user-images.githubusercontent.com/69441971/235509279-b29a8e9f-bc3e-48ce-aa0a-6f9c58e8cf28.png)

Nriver:
@meichthys The search engine url is currently hard-coded to duckduckgo.

The user I said is just one of the Chinese users who is using the Chinese version of Trillium that I made with the trilium-translation project.

Nriver:
Closing for https://github.com/zadam/trilium/pull/3901

PR #3901, opened 2023-05-04
customize search engine

Nriver:
related issue #3882

### DIFF
--- a/src/public/app/widgets/type_widgets/options/other/search_engine.js
+++ b/src/public/app/widgets/type_widgets/options/other/search_engine.js
@@ -1,0 +1,73 @@
+import OptionsWidget from "../options_widget.js";
+
+const TPL = `
+<div class="options-section">
+
+
+    <h4>Search Engine</h4>
+    
+    <p>Custom search engine requires both a name and a URL to be set. If either of these is not set, DuckDuckGo will be used as the default search engine.</p>
+    
+    <form class="sync-setup-form">
+        <div class="form-group">
+            <label>Predefined search engine templates</label>
+            <select class="predefined-search-engine-select form-control">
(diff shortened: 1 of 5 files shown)


src/public/app/widgets/type_widgets/options/other/search_engine.js
```diff
@@ -0,0 +1,73 @@
...
+</div>`;
+
+const SEARCH_ENGINES = {
+    "Bing": "https://www.bing.com/search?q={keyword}",
+    "Baidu": "https://www.baidu.com/s?wd={keyword}",
+    "Duckduckgo": "https://duckduckgo.com/?q={keyword}",
+    "Google": "https://www.google.com/search?q={keyword}",
+}
+
+export default class SearchEngineOptions extends OptionsWidget {
```
soulsands, line 46: this feature only works in desktop build. Runtime should be checked here.
```
 isEnabled() {
        return super.isEnabled() && utils.isElectron();
    }
```

Nriver: You are right. The check has been added.

soulsands:
These two options can be add to defaultOptions, similar to this commit    [  src\services\options_init.js  ](https://github.com/zadam/trilium/commit/6f238f5f1aec7dac83ca1e69441c66e06e220928)

Nriver:
> These two options can be add to defaultOptions, similar to this commit [ src\services\options_init.js ](https://github.com/zadam/trilium/commit/6f238f5f1aec7dac83ca1e69441c66e06e220928)

Thanks! This prevents the bug with undefined values.

zadam:
Thanks!

PR #3908, opened 2023-05-05
Allow etapi to update attribute fields isInheritable, position

mm21:
This PR allows etapi clients to update attribute fields isInheritable and position. I'm developing a Python library for working with Trilium and this will enable idiomatic management of attributes: the user simply maintains a Python list of attributes, and positions are automatically calculated. 

Branches already support updating position, and isInheritable is allowed for attribute creation, so this seems to be a mere oversight.

### DIFF
--- a/src/etapi/attributes.js
+++ b/src/etapi/attributes.js
@@ -17,7 +17,8 @@ function register(router) {
         'type': [v.mandatory, v.notNull, v.isAttributeType],
         'name': [v.mandatory, v.notNull, v.isString],
         'value': [v.notNull, v.isString],
-        'isInheritable': [v.notNull, v.isBoolean]
+        'isInheritable': [v.notNull, v.isBoolean],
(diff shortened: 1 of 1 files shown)


src/etapi/attributes.js
```diff
@@ -40,7 +41,9 @@ function register(router) {
     });
 
     const ALLOWED_PROPERTIES_FOR_PATCH = {
-        'value': [v.notNull, v.isString]
+        'value': [v.notNull, v.isString],
+        'isInheritable': [v.notNull, v.isBoolean],
```
zadam: `isInheritable` should be removed from PATCH. If you need to change this flag, you need to delete the attribute and create a new one with the desired value.

attribute type, name and isInheritable are considered immutable after creation and a lot of code relies on that.

mm21: Fixed. Thanks for the background, that requirement will be pretty easy to accommodate.

PR #3923, opened 2023-05-11
fix included note display

soulsands:
The display of a included note has been cut off if it's the first element. Maybe it would be better to exclude included notes from the margin rule.

from
image

to
image


soulsands:
@zadam    is this one acceptable? if not, we can close it.

zadam:
sorry, missed this.

soulsands:
> sorry, missed this.

it's okay.  Have a good day!

PR #3936, opened 2023-05-13
Add option to set the default behavior of openNote

tevenfeng:
Add an option in settings to let the users choose whether to open a new note in a new tab or in the current tab.


zadam:
Hi, can you please give some rationale for this config option?

It looks like you want to be able to open notes from the tree in new tab by default?

tevenfeng:
> Hi, can you please give some rationale for this config option?
> 
> It looks like you want to be able to open notes from the tree in new tab by default?

Yes, sometimes i will open multiple notes at the same time. And opening new note in current tab would be kinda annoying in this occasion, so i add this option

zadam:
But can't you just right click - open in new tab? Or just hold CTRL when clicking, it will open in a new tab ...

tevenfeng:
> But can't you just right click - open in new tab? Or just hold CTRL when clicking, it will open in a new tab ...

Well actually I'm not quite used to those ways. It would be fine if we don't merge this pr, just a patch for myself LOL

PR #3941, opened 2023-05-16
 toc scrolls smoothly

SiriusXT:
Notes (including read-only and edit mode) scroll smoothly when clicking toc

### DIFF
--- a/src/public/app/components/root_command_executor.js
+++ b/src/public/app/components/root_command_executor.js
@@ -54,6 +54,15 @@ export default class RootCommandExecutor extends Component {
             openService.openNoteExternally(noteId, mime);
         }
     }
+    
+    openNoteCustomCommand() {
+        const noteId = appContext.tabManager.getActiveContextNoteId();
+        const mime = appContext.tabManager.getActiveContextNoteMime()
+
+        if (noteId) {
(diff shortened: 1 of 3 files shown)


src/public/app/widgets/toc.js
```diff
@@ -227,15 +232,15 @@ export default class TocWidget extends RightPanelWidget {
...
-                textEditor.editing.view.scrollToTheSelection();
+                // model.change(writer => {
+                //     writer.setSelection(root.getChild(root.childCount - 1), 0);
+                // });
+                // textEditor.editing.view.scrollToTheSelection();
+                // // Backwards scroll to the heading
+                // model.change(writer => {
+                //     writer.setSelection(headingNode, 0);
+                // });
+                // textEditor.editing.view.scrollToTheSelection();
```
zadam: If this is indeed not needed then better to simply delete the code.

SiriusXT: I found that adding css: 
div.component.scrolling-container{
    scroll-behavior: smooth;
}
can also achieve smooth scrolling, and at the same time make the search function scroll smoothly. I withdraw this pr and modify the css.

PR #3963, opened 2023-05-23
keep this window on top

SiriusXT:
图片


SiriusXT:
> @SiriusXT This looks useful to me👍 Does this pin the Trilium window to the top of other open windows on the machine? If so, is it limited to the desktop version?

He pins the window to the top of other windows, only on the desktop

src/public/app/widgets/title_bar_buttons.js
```diff
@@ -30,9 +30,13 @@ const TPL = `
         height: 40px;
         width: 40px;
     }
+    .title-bar-buttons .top-btn.active{
+        background-color:var(--accented-background-color);
+    }
     </style>
 
     <!-- divs act as a hitbox for the buttons, making them clickable on corners -->
+    <div class="top-btn" title="Bring this window to the top. "><button class="btn bx bx-pin"></button></div>
```
zadam: It doesn't just bring it on top, it keeps it always on top, right?

SiriusXT: yes

zadam: That was a hint that the title should reflect this ;)

PR #3984, opened 2023-05-29
Add buttons to reorder split panes

dymani:
### Background
I use the split-pane function quite often to view other reference/literature notes when doing my writing. Sometimes I need to compare difference versions of my work, and I find it quite clumsy to swap the notes in the two panes, i.e. focus into the left pane, jump to another note, focus into the right pane, and then jump to another note.  

It would be nice if we can somehow reorder the panes to better organize the notes. The reordering/swapping buttons would also allow "closing the left pane" without using the Jump to Note function, that would become a simple swap followed by closing the right pane.

### Proposed funtion
Introduce "Move left" and "Move right" buttons next to "Create new split" and "Close this pane" buttons for users to reorder split panes. When split panes are used, these buttons will appear given that the moving direction is valid in relation to other panes on the screen. 

In case there are more than two panes, both buttons will be shown in the middle panes. Also, when one of the panes involved in the move operation is empty, it has the same effect as copying the non-empty pane to the empty one.  

In this implementation, the note contexts will switch directly to their new note paths. A small catch is that the scrolling position in the panes will not be preserved. I think it might be possible to utilize the `noteContextReorder` event, but I found it quite complicated since we also need to update the main context ID of all the subcontexts before refreshing the UI. Anyway, I think the current proof of concept should be good enough for most use cases.

### DIFF
--- a/src/public/app/widgets/buttons/move_pane_button.js
+++ b/src/public/app/widgets/buttons/move_pane_button.js
@@ -2,42 +2,6 @@ import OnClickButtonWidget from "./onclick_button.js";
 import appContext from "../../components/app_context.js";
 
 export default class MovePaneButton extends OnClickButtonWidget {
-    isEnabled() {
-        if (!super.isEnabled())
-            return false;
-        
-        if (this.isMovingLeft) {
-            // movable if the current context is not a main context, i.e. non-null mainNtxId
-            return !!this.noteContext?.mainNtxId;
-        } else {
-            const currentIndex = appContext.tabManager.noteContexts.findIndex(c => c.ntxId === this.ntxId);
-            const nextContext = appContext.tabManager.noteContexts[currentIndex + 1];
-            // movable if the next context is not null and not a main context, i.e. non-null mainNtxId
-            return !!nextContext?.mainNtxId;
-        }
-    }
-
-    initialRenderCompleteEvent() {
-        this.refresh();
-        super.initialRenderCompleteEvent();
-    }
-
-    async noteContextRemovedEvent({ntxIds}) {
-        this.refresh();
-    }
-
-    async newNoteContextCreatedEvent({noteContext}) {
-        this.refresh();
-    }
-
-    async noteContextSwitchEvent() {
-        this.refresh();
-    }
-
-    async noteContextReorderEvent({ntxIdsInOrder}) {
-        this.refresh();
-    }
-
     constructor(isMovingLeft) {
         super();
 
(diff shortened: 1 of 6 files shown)


zadam:
Thanks, that looks useful! I'll take a closer look tomorrow.

src/public/app/widgets/buttons/move_pane_button.js
```diff
@@ -0,0 +1,55 @@
...
+
+    async noteContextSwitchEvent() {
+        this.refresh();
+    }
+
+    async noteContextReorderEvent({ntxIdsInOrder}) {
+        this.refresh();
+    }
+
+    constructor(isMovingLeft) {
```
zadam: constructor is usually at the top

src/public/app/widgets/containers/split_note_container.js
```diff
@@ -74,6 +74,47 @@ export default class SplitNoteContainer extends FlexContainer {
...
+        const leftIndex = isMovingLeft ? currentIndex - 1 : currentIndex;
+
+        if (currentIndex === -1 || leftIndex < 0 || leftIndex + 1 >= contexts.length) {
+            logError("invalid context!");
+            return;
+        }
+
+        if (contexts[leftIndex].isEmpty() && contexts[leftIndex + 1].isEmpty())
+            // no op
+            return;
```
zadam, line 95: Please use curly braces also for single statement blocks.

src/public/app/widgets/containers/split_note_container.js
```diff
@@ -74,6 +74,47 @@ export default class SplitNoteContainer extends FlexContainer {
...
+        }
+
+        const contexts = appContext.tabManager.noteContexts;
+
+        const currentIndex = contexts.findIndex(c => c.ntxId === ntxId);
+        const leftIndex = isMovingLeft ? currentIndex - 1 : currentIndex;
+
+        if (currentIndex === -1 || leftIndex < 0 || leftIndex + 1 >= contexts.length) {
+            logError("invalid context!");
```
zadam: It would be useful to dump the `currentIndex`, `leftIndex` and `contexts.length` to the error message for debugging.

src/public/app/widgets/containers/split_note_container.js
```diff
@@ -74,6 +74,47 @@ export default class SplitNoteContainer extends FlexContainer {
...
+
+        this.triggerCommand("noteContextReorder", {
+            ntxIdsInOrder: [
+                ...ntxIds.slice(0, leftIndex),
+                ntxIds[leftIndex + 1],
+                ntxIds[leftIndex],
+                ...ntxIds.slice(leftIndex + 2),
+            ],
+            oldNtxIdsInOrder: ntxIds,
+            mainNtxIdsInOrder: mainNtxIds.map(id => id === ntxIds[leftIndex] ? ntxIds[leftIndex + 1] : id)
```
zadam: I find these 3 arrays pretty complicated to understand - e.g. the fact that mainNtxIdsInOrder contains duplicated items. I wonder if there's a simpler way to do this.

zadam (commented):
Looks good with some minor comments.

dymani:
I have simplified the logic by passing the new and old main contexts IDs instead of passing arrays around, since we can assume there will be at most 1 main context change per reorder event.

zadam:
Thanks for the contribution!

PR #3988, opened 2023-05-31
Show highlighted text in the right pane

SiriusXT:
Display highlighted text in the left pane, so you can quickly jump to important content.
As stated in issue [3973](https://github.com/zadam/trilium/issues/3973), I found that Trillium lacks the bookmark function within the notes. For a long article, it may not be convenient to jump to the key points in the article. However, highlighted text can be used as a bookmark to facilitate quick jumps within the notes.
preview


src/public/app/layouts/desktop_layout.js
```diff
@@ -181,6 +182,7 @@ export default class DesktopLayout {
                     )
                     .child(new RightPaneContainer()
                         .child(new TocWidget())
+                        .child(new HltWidget())
```
zadam: "HLT" is pretty unclear, I think it would be better to expand the name to full.

src/public/app/widgets/highlighted_text.js
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    async doRenderBody() {
+        this.$body.empty().append($(TPL));
+        this.$hlt = this.$body.find('.hlt');
+        this.$body.find('.hlt-widget').append(this.closeHltButton.render());
+    }
+
+    async refreshWithNote(note) {
+        const hltLabel = note.getLabel('hlt');
```
zadam: I would expand again the HLT here. Also, there's a pattern for these labels - name e.g. `hideHighlightWidget` and check with `FNote.isLabelTruthy()`. In most cases you don't need to define value, but the method handles also overriding with e.g. `hideHighlightWidget=false`.

src/public/app/widgets/type_widgets/options/text_notes/highlighted_text.js
```diff
@@ -0,0 +1,90 @@
...
+            display: inline-block;
+            min-width: 8em;
+        }
+        .options-section{
+            max-width: 46em;
+        }
+    </style>
+    <h4>Highlighted Text</h4>
+    
+    Displays highlighted text in the left pane. You can customize the highlighted text displayed in the left pane:
```
zadam: The widget is displayed in the right panel.

src/public/app/widgets/type_widgets/options/text_notes/highlighted_text.js
```diff
@@ -0,0 +1,90 @@
...
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#e6994c"> Orange&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#e6e64c"> Yellow&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#99e64c"> Light green&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#4ce64c"> Green&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#4ce699"> Aquamarine&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#4ce6e6"> Turquoise&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#4c99e6"> Light blue&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#4c4ce6"> Blue&nbsp;</label>
+<label class='hlt-checkbox-label'><input type="checkbox" class="hlt-background-color" value="#994ce6"> Purple&nbsp;</label>
```
zadam: I don't fully understand the need for this configuration. Why not display all highlighted snippets?

Also, I assume these values match the CKEditor color palette for highlighting?

SiriusXT: Because not all highlighted text needs to be displayed on the right pane, sometimes, users may not want some highlighted text to be displayed on the right pane when setting the highlight.
These values match the CKEditor color palette for highlighting. I will change value to name later for better understanding

src/public/app/widgets/highlighted_text.js
```diff
@@ -0,0 +1,267 @@
...
+            ["", "show"].includes(hltLabel?.value)
+            || hltColors!="" 
+            || hltBgColors!=""
+        );
+
+        this.triggerCommand("reEvaluateRightPaneVisibility");
+    }
+    //Converts color values in RGB, RGBA, or HSL format to hexadecimal format, removing transparency
+    colorToHex(color) {
+        function rgbToHex(rgb) {
```
soulsands: It would be better to extract these pure functions to a utils file or another location so that they can be shared and reduce the cost of redefining them. It is also desirable for this conversion to be configurable since it is not necessary for users who do not import notes.

src/public/app/widgets/highlighted_text.js
```diff
@@ -0,0 +1,267 @@
...
+            if (Math.abs(parseInt(optionColors[i].slice(5, 7), 16) - parseInt(targetColor.slice(5, 7), 16)) > 2) { continue; }
+            return true;
+        }
+        return false;
+    }
+    /**
+     * Builds a jquery table of helight text.      
+     */
+    getHlt(html, optionsHltColors, optionsHltBgColors) {
+        const hltBCs = $(html).find(`span[style*="background-color"],span[style*="color"]`)
```
soulsands: The $ function creates a lot of elements, which may cause an increase in memory usage and calculation costs. I'm not sure, but perhaps using Regex would perform better.

src/public/app/widgets/highlighted_text.js
```diff
@@ -0,0 +1,267 @@
...
+        this.closeHltButton = new CloseHltButton();
+        this.child(this.closeHltButton);
+    }
+
+    get widgetTitle() {
+        return "Highlighted Text";
+    }
+
+    isEnabled() {
+        return super.isEnabled()
```
soulsands, line 62: This is a promising future, but it may also cause performance issues, especially on a poor machine like mine. Additionally, some users may not be accustomed to using the highlight feature. Personally I prefer to use bold to emphasize a snippet. So it would be beneficial to make the feature configurable for certain users.

soulsands (commented):
You are doing great, keep it up

src/public/app/widgets/highlighted_text.js
SiriusXT: has been modified to a regular expression

src/public/app/widgets/highlighted_text.js
SiriusXT: Added hideHighlightWidget

src/public/app/widgets/highlighted_text.js
SiriusXT: This function has been removed

src/public/app/widgets/highlighted_text.js
SiriusXT: When you do not check any items in the options, refreshWithNote will return quickly
```
if (hltLabel?.value == "" || hltLabel?.value === "true" || (optionsHltColors=="" && optionsHltBgColors=="")) {
            this.toggleInt(false);
            this.triggerCommand("reEvaluateRightPaneVisibility");
            return;
        }
```
I think bold is often used to emphasize some keywords instead of sentences. I feel that it is more appropriate to only display highlights. I am not sure whether it is necessary to configure bold display

src/public/app/widgets/type_widgets/options/text_notes/highlighted_text.js
SiriusXT: > Because not all highlighted text needs to be displayed on the right pane, sometimes, users may not want some highlighted text to be displayed on the right pane when setting the highlight. These values match the CKEditor color palette for highlighting. I will change value to name later for better understanding

```
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Black"> Black &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Dim grey"> Dim grey &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Grey"> Grey &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Light grey"> Light grey &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="White"> White &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Red"> Red &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Orange"> Orange &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Yellow"> Yellow &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Light green"> Light green &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Green"> Green &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Aquamarine"> Aquamarine &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Turquoise"> Turquoise &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Light blue"> Light blue &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Blue"> Blue &nbsp;</label>
<label class='highlighted-text-label'><input type="checkbox" class="highlighted-text-color" value="Purple"> Purple &nbsp;</label>
```

src/public/app/widgets/highlighted_text.js
soulsands: oh I didnt see that. it's already pretty good. I dont think it's necessary to include bolded texts. 
Actually I prefer bolding just because it has a shortcut and is supported by markdown. In my patten,  bold, italic, quote texts are colored via custom styles, and these colors resonate with the trilium logo :)

src/public/app/widgets/type_widgets/options/text_notes/highlighted_text.js
SiriusXT: I thought about it, it seems that the function of selecting the color is really unnecessary, I will remove it later, and then add the function of bold underline italic

src/public/app/widgets/highlighted_text.js
SiriusXT: Modified to support bold italic underline
图片

src/public/app/widgets/type_widgets/options/text_notes/highlighted_text.js
SiriusXT: Color options removed, modified to support bold italic underline

PR #3997, opened 2023-06-02
better size format

Nriver:


src/public/app/services/utils.js
```diff
@@ -19,6 +19,18 @@ function padNum(num) {
...
+function formatSize(bytes, decimals = 2) {
+    if (!+bytes) return '0 Bytes'
+
+    const k = 1024
+    const dm = decimals < 0 ? 0 : decimals
+    const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+
+    const i = Math.floor(Math.log(bytes) / Math.log(k))
+
+    return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`
```
zadam, line 31: There's no need to go above MiB since Trilium can't store gigabyte sized files. Honestly I think the current method is good enough.

Nriver:
Ok, I've seen you made some changes. Closing this pr.

PR #4001, opened 2023-06-02
Reopen recently closed tab/split in place

dymani:
As suggested by @adityaverma415 in #3993, when closing a tab (one or more splits) or a single split, the position will be stored in `tabManager.recentlyClosedTabs` so that when the tab/split is reopened, they can be reordered back to their original position.


dymani:
@zadam What do you think? Would this be a useful addition?

zadam:
Sorry, looks like I missed this ...

dymani:
No problem!

PR #4017, opened 2023-06-09
Fix SQL Console error messages

### LINKED ISSUE #4015: SQL console errors are unhelpful (always 'undefined')
Using the SQL console to execute a query with an error gives the following toast dialog in the frontend:
> Error occurred while executing SQL query: undefined

The code for this message is here:
https://github.com/zadam/trilium/blob/c177aaa901a9388bbce5341b3e6fdb41195dcbe2/src/public/app/components/entrypoints.js#L176

But here we see the backend returns an `error` key, not a `message` key, which is why it is always undefined:
https://github.com/zadam/trilium/blob/c177aaa901a9388bbce5341b3e6fdb41195dcbe2/src/routes/api/sql.js#L54-L57

I can create a PR for this in the morning.

mechanarchy:
Fixes #4015 

Server replies with an `error` key but the frontend is expecting a `message` key, this unifies both sides to use `error`. Details in #4015

### DIFF
--- a/src/public/app/components/entrypoints.js
+++ b/src/public/app/components/entrypoints.js
@@ -173,7 +173,7 @@ export default class Entrypoints extends Component {
             const resp = await server.post(`sql/execute/${note.noteId}`);
 
             if (!resp.success) {
-                toastService.showError(`Error occurred while executing SQL query: ${resp.message}`);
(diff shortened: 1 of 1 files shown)


zadam:
Thanks!

PR #4019, opened 2023-06-10
Fix attempted ETAPI token rename after cancel

### LINKED ISSUE #4018: Closing ETAPI token rename model causes server error
In v0.59.4, server: closing the pop-up "rename ETAPI token" modal still sends a `PATCH` request with payload `{name: null}`. The server then responds with `HTTP 500` error about `NOT NULL constraint failed: etapi_tokens.name`

Problem origin:
https://github.com/zadam/trilium/blob/c177aaa901a9388bbce5341b3e6fdb41195dcbe2/src/public/app/widgets/type_widgets/options/etapi.js#L106-L116

Probably the answer is to check if `tokenName !== null` before sending the request.

mechanarchy:
Fixes #4018


zadam:
Thanks.

PR #4032, opened 2023-06-17
Ignore SQL comments in query console

mechanarchy:
Currently the SQL console doesn't return rows if a query starts with a comment. Consider the following:

```sql
SELECT * FROM notes WHERE title = 'root';
---
-- SELECT * FROM branches WHERE parentNoteId = 'root';
SELECT * FROM branches WHERE parentNoteId = '_hidden';
```

The SQL console results are:
```
{
	"changes": 0,
	"lastInsertRowid": 16363
}
```

This PR allows results from both queries to be returned. Note that it does look like a multi-statement SQL query that includes both `SELECT` and non-`SELECT` (eg, `UPDATE`) queries discards the result of the `SELECT` in the frontend. This is probably not a problem in real-world usages but might be worth fixing.

### DIFF
--- a/src/routes/api/sql.js
+++ b/src/routes/api/sql.js
@@ -36,6 +36,11 @@ function execute(req) {
             if (!query) {
                 continue;
             }
+            
+            while (query.startsWith('-- ') {
+                // Query starts with one or more SQL comments, discard these before we execute.
+                query = query.substr(query.indexOf('\n') + 1)
+            }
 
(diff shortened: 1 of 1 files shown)


mechanarchy:
Rereading my code, I suspect this will enter an infinite loop if the only input to be evaluated is an SQL comment.

I'll fix it when I get an opportunity by moving the comment removal code above the empty-string check and handling the newline stuff better.

mechanarchy:
Fixed the issue, @zadam at your leisure could you please review and merge. Thanks in advance.

PR #4057, opened 2023-06-27
Fix closing parenthesis

dymani:


mechanarchy:
That's embarrassing, thanks for finding and fixing this @dymani

Issue #4074, opened 2023-07-02
(Feature request) Save webclipper notes/clippings to single page


manto89:
### Describe feature

When using [Trilium Web Clipper](https://github.com/zadam/trilium-web-clipper) I want the clippings (selected excerpts of a web page) and the notes (custom text that is added with a link to the web page) to be added all in one single trilium note, no matter when i added previous clippings/notes. 
#### Current behavior
Right now the system will join clippings and notes referring to the same page only if they are in the same 'daily note'. 
#### My solution 
If there's already a note under `Inbox` (or a note with the `clipperInbox` label) that has a label `pageUrl`, the new clipping/note content will be added to that page. If the page is not under today's `daily-note` then clone the page under today's `daily-note`.
If there isn't any note with the same `pageUrl` then create a new note under today's `daily-note` (this is already the current behavior) 

### Additional Information

I already worked on a PR for this feature, will link the PR asap

PR #4075, opened 2023-07-02
Feature/add clippings existing url

manto89:
#4074 
I only touched `clipper.js`, this should keep the main functionality intact.
I had to fix both `addClipping` and `createNote` functions, since they are responsible for creating notes with type `clipping` (used in trilium-web-clipper with the select+right click menu) and the `note` (used in trilium-web-clipper main menu).
I also fixed the `processContent` function because it would sometimes add content with not closed tags, so i opted to use jsdom to parse the content before adding it to the note.
Lastly, in order for the `<br>` tag to work correctly when appending notes with content that is just a string, i opted to add such content inside a `<p>` tag. This check is done with a regex which should be sufficient although not perfect

### Additional Notes
This PR includes exposing a new clipper endpoint 
`/api/clipper/notes-by-url/:noteUrl` that is needed by zadam/trilium-web-clipper#54

### DIFF
--- a/src/routes/api/clipper.js
+++ b/src/routes/api/clipper.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const attributeService = require("../../services/attributes");
+const cloneService = require("../../services/cloning");
 const noteService = require('../../services/notes');
 const dateNoteService = require('../../services/date_notes');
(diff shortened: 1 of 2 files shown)


manto89:
Thanks!

Issue #4026, opened 2023-06-13
(Feature request) Time attribute field


Nitramakloh:
### Describe feature

I would like to use Trilium as a time tracker. Therefore, I would enter a time (in format hh:mm:ss) in each note entry using attributes. However, when I wanted to create some, I discovered that there is no field for times apart from decimal, date (without time) and currency. Would this be possible without much effort? Can I perhaps do this myself using JS code? Unfortunately, I am not familiar with JavaScript.

zadam:
Well, "17:25:00" is just text and can be stored as such. Actually all label values are just text. Or do you mean some extra support for time?

Nitramakloh:
Hello zadam, thank you for speaking up. I did not know that the values are stored internally as text. But I wish the extra support because of better comfort when entering. It is easy to mistype, so ":" then becomes ";" for example.
A sufficient alternative for me would also be, if the time attribute would be automatically filled with the current time when creating a new note entry. But even there I don't know how to solve this script-wise.

eliandoran:
This is already done.

PR #4090, opened 2023-07-10
added date and time attribute capability

harishkumarc:
I have added the capability to define datetime attributes, please check the below screenshot 

Screenshot from 2023-07-10 17-22-51

solves #4026

### DIFF
--- a/src/public/app/services/promoted_attribute_definition_parser.js
+++ b/src/public/app/services/promoted_attribute_definition_parser.js
@@ -6,7 +6,7 @@ function parse(value) {
         if (token === 'promoted') {
             defObj.isPromoted = true;
         }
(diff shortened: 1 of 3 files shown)


jwhonce:
@harishkumarc is there a knob to use a 24-hour clock vs 12-hour AM/PM?

harishkumarc:
> @harishkumarc is there a knob to use a 24-hour clock vs 12-hour AM/PM?

I don't think so,it doesn't come by default,  let me see if I can add it some other way

zadam:
These controls respect locale, so e.g. in Europe they will use 24 hour format.

The label value itself is saved in ISO date time format, e.g. \`#date="2023-07-12T11:28"\`

PR #4155, opened 2023-08-09
Add a new note type drawio

bbsweb:
Add a new note type drawio

image

I don't know if this PR is suitable, It is based on [draw.io](https://app.diagrams.net/)


zadam:
Hello, this seems to download draw.io from the internet and thus won't work offline, which is a must have requirement.

bbsweb:
yes, unfortunately, it will use the draw.io iframe

PR #4183, opened 2023-08-18
Stop note map simulation if not visible

FliegendeWurst:
Without this fix, Trilium will be sluggish for a while (usually less than a minute) after closing a large note map.

### DIFF
--- a/src/public/app/widgets/containers/ribbon_container.js
+++ b/src/public/app/widgets/containers/ribbon_container.js
@@ -198,6 +198,10 @@ export default class RibbonContainer extends NoteContextAwareWidget {
                 }
             }
         } else {
+            const activeChild = this.getActiveRibbonWidget();
+            if (activeChild) {
+                activeChild.handleEvent('deactivated', {});
+            }
             this.lastActiveComponentId = null;
         }
     }
--- a/src/public/app/widgets/note_detail.js
+++ b/src/public/app/widgets/note_detail.js
@@ -217,6 +217,9 @@ export default class NoteDetailWidget extends NoteContextAwareWidget {
 
     async beforeNoteSwitchEvent({noteContext}) {
         if (this.isNoteContext(noteContext.ntxId)) {
+            for (const x of this.children) {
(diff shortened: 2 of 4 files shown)


src/public/app/widgets/containers/ribbon_container.js
```diff
@@ -198,6 +198,10 @@ export default class RibbonContainer extends NoteContextAwareWidget {
                 }
             }
         } else {
+            const activeChild = this.getActiveRibbonWidget();
+            if (activeChild) {
+                activeChild.handleEvent('deactivated', {});
```
FliegendeWurst, line 203: I am using a new event here since I did not find any existing mechanism to shut down a widget.

src/public/app/widgets/note_detail.js
```diff
@@ -217,6 +217,9 @@ export default class NoteDetailWidget extends NoteContextAwareWidget {
 
     async beforeNoteSwitchEvent({noteContext}) {
         if (this.isNoteContext(noteContext.ntxId)) {
+            for (const x of this.children) {
```
FliegendeWurst, line 220: Is it intentional that the children array is never properly cleared after switching notes?

src/public/app/widgets/containers/ribbon_container.js
zadam: Have you tried `cleanup()` method in TypeWidget (inherited in NoteMapTypeWidget)? Seems to me it could be used for this. Although it would not work for Ribbon specifically ...

src/public/app/widgets/note_detail.js
zadam: Yes, it's there to keep e.g. CKEditor instance which is expensive to construct.

src/public/app/widgets/containers/ribbon_container.js
FliegendeWurst: It doesn't matter as much for the Ribbon widget, but could still be relevant for cases like #2608

PR #4202, opened 2023-08-26
Use a list instead of dropdown for jump-to-note

eliandoran:
This is a proof-of-concept implementation in which the jump-to-note uses a list instead of a dropdown auto-completion. This approach seems to be used by most applications having a "Jump to..." functionality such as: VS Code, GNOME Builder, Obsidian.

It's achieved by tweaking some settings for the autocomplete library, including setting `debug` to avoid hiding the list when the input loses focus. 

All the other use-cases where note auto-completion is used should not be affected by this change.

@zadam , let me know your thoughts on this.

### Further improvements

I also propose the following:

* Consider removing the "X" button and show recents button near the input box since they don't provide much functionality in this case (the X button is equivalent to Ctrl+A followed by a backspace), whereas the recents are displayed now by default.
* Add a "No results" text instead of showing an empty list. VS Code displays a "No matching results" item whereas Obsidian allows the user to create a new note, similar to how Trilium does it on the new tab screen.
* Consider removing the "X" button near the modal. The input box looks slightly unbalanced because it was moved in the header. Not sure of the usefulness of the button since the dialog can be dismissed by clicking out of the modal or pressing Escape.

### Recording

[simplescreenrecorder-2023-08-26_17.54.59.webm](https://github.com/zadam/trilium/assets/21236836/46f49ba7-c0be-4657-82da-098d15a8af67)

### The "Jump to dialog" list in other applications:

VS code:

image

GNOME builder:

image

Obsidian:

image

### DIFF
--- a/src/public/app/services/note_autocomplete.js
+++ b/src/public/app/services/note_autocomplete.js
@@ -140,8 +140,9 @@ function initNoteAutocomplete($el, options) {
     let autocompleteOptions = {};
     if (options.container) {
         autocompleteOptions.dropdownMenuContainer = options.container;
+        autocompleteOptions.debug = true;   // don't close on blur
     } else {
-        autocompleteOptions.appendTo = document.querySelector('body');
+        autocompleteOptions.appendTo = document.querySelector('body');        
     }
 
     $el.autocomplete({
--- a/src/public/app/services/note_autocomplete.js
+++ b/src/public/app/services/note_autocomplete.js
@@ -106,7 +106,7 @@ function initNoteAutocomplete($el, options) {
     $el.addClass("note-autocomplete-input");
 
     const $clearTextButton = $("<a>")
-            .addClass("input-group-text input-clearer-button bx bx-x")
+            .addClass("input-group-text input-clearer-button bx bxs-tag-x")
             .prop("title", "Clear text field");
 
     const $showRecentNotesButton = $("<a>")
--- a/src/public/app/services/note_autocomplete.js
+++ b/src/public/app/services/note_autocomplete.js
@@ -137,8 +137,15 @@ function initNoteAutocomplete($el, options) {
         return false;
     });
 
+    let autocompleteOptions = {};
+    if (options.container) {
+        autocompleteOptions.dropdownMenuContainer = options.container;
+    } else {
+        autocompleteOptions.appendTo = document.querySelector('body');
+    }
+
     $el.autocomplete({
-        appendTo: document.querySelector('body'),
+        ...autocompleteOptions,
         hint: false,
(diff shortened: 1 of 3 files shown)


rauenzi:
I tested this locally, it works great and I think it's a good change.

I also think your other proposed changes are good ideas as well. Having a `Create New Note` option when there's no results would be good for UX, and I agree the extra `x` for the modal feels unnecessary. But the `x` and recents buttons in the input modal I think are okay to keep, they might provide only minimal utility, but I'm sure some people will use them.

eliandoran:
@rauenzi , thanks for the feedback. I've added the "Create new note" option, let me know your thoughts.
It was surprisingly easy, one line change thanks to @zadam 's code.

rauenzi:
LGTM! 

image

I also removed the modal close button locally and it does make it feel more balanced like the other examples you showed in the OP.

eliandoran:
@rauenzi , for the moment I've kept the close button until @zadam also says his opinion, but I've improved its centering. I've also adjusted the paddings so that the text does not appear cut off near the margins.

For consistency, maybe we should keep the X button. However personally I'd change the "X" button inside the text box with a backspace icon like GNOME does. Should look even better.

eliandoran:
@rauenzi , what browser are you using? I see you have a horizontal scrollbar that probably shouldn't be there.

rauenzi:
I think a backspace like GNOME would work as well, as long as it's there in some way I think it's good.

I am using Firefox which automatically prepares scrolls whenever `overflow` is set to `scroll`, if you change it to to `overflow-y: auto` then it will only show a vertical scroll and it will only happen when necessary. Or `overflow: scroll` if you want to allow horizontal scroll but only when needed.

eliandoran:
@rauenzi , you're right, I've somehow missed that. I've now set it to `auto` in order to be hidden when not needed.

@zadam , this is how the backspace icon looks like:

image

I think it creates a nice balance and it's relatively more intuitive than the X button which is also used for closing modals. If you're OK with it, I'll create a separate PR to use it everywhere (for consistency).

rauenzi:
I actually like that a lot more than I expected, and I agree that it's an improvement in terms of UX since it no longer collides with the modal close icon.

eliandoran:
I couldn't find any other use cases where we have a button to clear a text box. Most other use cases are for closing modals and for bulk actions so I've pushed the change directly here.

rauenzi:
I know one doesn't exist in this search box yet, but if Zadam likes this change it might be a good idea to add one here as well.

image

zadam:
Looks very nice, thanks for the contribution!

PR #4238, opened 2023-09-05
Optimize the closing logic of the note tooltip.

dvai:
When I move the mouse out of the tooltip, in most cases it means that I am no longer interested in it. Adding this change can avoid many unexpected behaviors, such as the tooltip still remaining on the interface after clicking a link in the tooltip.

### DIFF
--- a/src/public/app/services/note_tooltip.js
+++ b/src/public/app/services/note_tooltip.js
@@ -65,11 +65,12 @@ async function mouseEnterHandler() {
     }
 
     const html = `<div class="note-tooltip-content">${content}</div>`;
+    const tooltipClass = 'tooltip-' + Math.floor(Math.random() * 999_999_999);
 
     // we need to check if we're still hovering over the element
     // since the operation to get tooltip content was async, it is possible that
     // we now create tooltip which won't close because it won't receive mouseleave event
-    if ($(this).is(":hover")) {
+    if ($(this).filter(":hover").length > 0) {
         $(this).tooltip({
             container: 'body',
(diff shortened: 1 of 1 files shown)


src/public/app/services/note_tooltip.js
```diff
@@ -93,6 +93,10 @@ async function mouseEnterHandler() {
             }
         }, 1000);
     }
+    // When the mouse is moved out, it means that the tooltip is no longer being attended to.
+    $(`.${linkId}`).on("mouseleave", () => {
+        setTimeout(() => $(this).tooltip('dispose'), 1000);
+    })
```
zadam: How does this differ from the code a few lines above disposing of the tooltip after a second if mouse doesn't hover over it? Seems to be doing the same thing ...

dvai:
@zadam  If this code segment is not added, the tooltip will not close automatically after my mouse hovers over it. You can compare these two videos:

before：

https://github.com/zadam/trilium/assets/33273040/efd4a02f-9204-41cf-ba15-f58322fdb0b2

after:


https://github.com/zadam/trilium/assets/33273040/832757ea-c8b0-40dd-8152-255ce2e67dfb

rauenzi:
I tested both on stable and beta and I couldn't recreate any stuck tooltips. Can you provide an example the causes it consistently? I think it'd be good to understand why the existing code is insufficient (if it really is).

zadam:
I managed to reproduce some of the stuck tooltips (on master) and I believe I fixed them.

I don't want to use the mouseleave event, because I want to allow the user to move the cursor over the tooltip to be able to e.g. scroll it and click links within it.

PR #4240, opened 2023-09-06
Fix right click bookmark bug

SiriusXT:
Right-click the bookmark note in the launch bar to call the right-click menu, but onAuxClick will be triggered and the note will be opened directly. This PR fixes the bug.


zadam:
Thx!

PR #4251, opened 2023-09-09
Add keyboard shortcuts for tabs

### LINKED ISSUE #3735: Rofi like Tab Switching for Open Notes
### Discussed in https://github.com/zadam/trilium/discussions/3730

<div type='discussions-op-text'>

<sup>Originally posted by **adityaverma415** March 17, 2023</sup>
Hello,

Sometimes, I find myself having multiple files open simultaneously. To navigate to the desired file, I have to use my cursor or to keep pressing `Ctrl + Tab` till I arrive at the desired file. While this is alright, I think it would be great if we could pull up a menu of open files with the press of a button combination and then just select and open a desired file. This functionality has been implemented for open windows in [`rofi`](https://github.com/davatorium/rofi).

If this is a plausible idea, would you implement it? If yes, how long would it take?

Thanks for your time.</div>

rauenzi:
Partially fixes #3735. I thought about implementing a separate menu where you can browse tabs, but I think that would better fit as part of the command palette whenever that happens.

Let me know if the implementation is alright.

### DIFF
--- a/src/public/app/widgets/tab_row.js
+++ b/src/public/app/widgets/tab_row.js
@@ -267,7 +267,27 @@ export default class TabRowWidget extends BasicWidget {
                 }
             });
         });
+
+        keyboardActionService.setupActionsForElement('tabs', $(document), this);
     }
+
+    goToTab(tabNumber) {
+        const index = tabNumber === 0 ? this.tabEls.length - 1 : tabNumber - 1;
+        const tab = this.tabEls[index];
+        if (!tab) return;
+        appContext.tabManager.activateNoteContext(tab.getAttribute('data-ntx-id'));
+    }
+
+    firstTabCommand() {this.goToTab(1);}
+    secondTabCommand() {this.goToTab(2);}
+    thirdTabCommand() {this.goToTab(3);}
+    fourthTabCommand() {this.goToTab(4);}
+    fifthTabCommand() {this.goToTab(5);}
+    sixthTabCommand() {this.goToTab(6);}
+    seventhTabCommand() {this.goToTab(7);}
+    eigthTabCommand() {this.goToTab(8);}
+    ninthTabCommand() {this.goToTab(9);}
+    lastTabCommand() {this.goToTab(0);}
 
     setupStyle() {
         this.$style = $("<style>");
--- a/src/public/app/widgets/tab_row.js
+++ b/src/public/app/widgets/tab_row.js
@@ -268,11 +268,11 @@ export default class TabRowWidget extends BasicWidget {
             });
         });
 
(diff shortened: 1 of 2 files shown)


src/public/app/widgets/tab_row.js
```diff
@@ -267,8 +267,28 @@ export default class TabRowWidget extends BasicWidget {
...
+    firstTabCommand() {this.goToTab(1);}
+    secondTabCommand() {this.goToTab(2);}
+    thirdTabCommand() {this.goToTab(3);}
+    fourthTabCommand() {this.goToTab(4);}
+    fifthTabCommand() {this.goToTab(5);}
+    sixthTabCommand() {this.goToTab(6);}
+    seventhTabCommand() {this.goToTab(7);}
+    eigthTabCommand() {this.goToTab(8);}
+    ninthTabCommand() {this.goToTab(9);}
+    lastTabCommand() {this.goToTab(0);}
```
zadam: Instead of 0, I'd use something more distinct like -1 or 999. 0 is often used for the first element and thus could be ambiguous.

src/services/keyboard_actions.js
```diff
@@ -231,6 +231,66 @@ const DEFAULT_KEYBOARD_ACTIONS = [
         description: "Open new empty window",
         scope: "window"
     },
+    {
+        actionName: "firstTab",
+        defaultShortcuts: ["CommandOrControl+1"],
+        description: "Activates the first tab in the list",
+        scope: "tabs"
```
zadam: Scope in this context means on which element the keyboard shortcuts are attached - this means that for the shortcut to work, the focus has to be inside the element.

So e.g. shortcuts with "note-tree" scope work only when the focus is in note tree. This allows to have the same shortcut have different actions attached based on the context - e.g. CTRL-C in note tree does something a bit different from CTRL-C in the text editor.

However here I don't see a need for a separate scope - these should be "global" - active on the whole application, which corresponds to the "window" scope.

zadam (changes requested):

src/public/app/widgets/tab_row.js
zerebos: I think `-1` could get confused but index not found, but I will definintely use something more unique than `0`.

src/services/keyboard_actions.js
zerebos: I definitely agree it didn't seem necessary, I just wasn't sure if doing `keyboardActionService.setupActionsForElement('window', $(document), this);` would cause some other issues with other listeners, but after testing it seems fine so I'll switch it back to `window`

PR #4281, opened 2023-09-22
Fix jump to ToC for nested headers

### LINKED ISSUE #4228: (Bug report) TOC problem
### Trilium Version

0.60.4

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 11, 22H2

### Description

Hi, I start using trilium recently,  with docker server on Debian 12 and client on Windows11. And today I found when I create a TOC with sub-headings which has a quoted format, the last few items would not work properly. For example, as the following picture shows, the last two items would not do when I click on them, just remove `1.1 test` and `1.2 testtest`, all of the toc work fine, even just remove the quoted format will do.

Not sure if it's my setting problem

image

### Error logs

_No response_

rauenzi:
This is meant to fix #4228. I saw your comment there about trying to search in arbitrary depth of the ckeditor model https://github.com/zadam/trilium/issues/4228#issuecomment-1707341883, but I thought why not try to make it use the same method as read-only notes?

I tested this out locally of course and it seems to work as expected for a bunch of different nested cases. But you can just consider this a proposal and feel free to close it if you don't think it's the best way forward.

### DIFF
--- a/src/public/app/widgets/toc.js
+++ b/src/public/app/widgets/toc.js
@@ -185,32 +185,16 @@ export default class TocWidget extends RightPanelWidget {
         // See https://github.com/zadam/trilium/issues/2828
         const isReadOnly = await this.noteContext.isReadOnly();
 
+        let $container;
         if (isReadOnly) {
-            const $container = await this.noteContext.getContentElement();
-            const headingElement = $container.find(":header:not(section.include-note :header)")[headingIndex];
-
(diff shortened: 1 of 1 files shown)


zadam:
I kind of expected some trouble because of all the advanced stuff CKEditor does on the background, but if this simple straightforward approach works, all the better ...

Issue #3430, opened 2022-12-17
(Feature request) Add basic fulltext search to share


zadam:
### Describe feature

no attribute search or advanced features, just fulltext

passionate2023:
Is there is any update on this open ticket?

sigaloid:
If you or anyone else wants to work on adding this, one possible avenue is using Lunr.js to create a search index under the share API and adding a client-side search bar.

https://opensource.com/article/21/11/client-side-javascript-search-lunrjs
https://lunrjs.com/
https://olivernn.github.io/moonwalkers/

passionate2023:
@zadam is there is any possibility that this feature get added in this year?

zadam:
Yes. Can't say how high, though.

passionate2023:
@zadam is there is any possibility that this feature get added in this year?

passionate2023:
@zadam is there is any possibility that this feature get added in this year?

rauenzi:
The basis for this actually already exists, but the default share template does not make use of it. I am working on creating a new default share template that will include this.  Of course I will try to finish it as soon as possible, but life has been quite busy and hectic for me lately. That said, I cannot guarantee that it will be completed, merged, and released by the end of the year.

zadam:
@passionate2023 

Yes. Can't say how high, though.

passionate2023:
@rauenzi Is the changed got merged and released already ?

eliandoran:
Delivered some time ago.

Issue #3429, opened 2022-12-17
(Feature request) Improve default Share theme


zadam:
### Describe feature

Current theme looks a bit too plain and could be improved.

rauenzi:
I decided to work on this a bit. I'm using the default Trilium dark theme as the inspiration for the share theme. Here's how it looks so far:

https://github.com/zadam/trilium/assets/6865942/d5f6d08b-ebd3-48ed-895a-4780865cda3b

Zadam, is it alright to make changes to the template/html? I guess it would technically break B/C but let me know what you think.

rauenzi:
More than happy to help out @meichthys! I love Trilium so I want to give back to both zadam and the community!

I've also been working on some other things in the background for the community like https://trilium.rocks/ but I don't want to go too far or step on any toes, so I don't want to really announce them until I talk to zadam/see how he feels.

rauenzi:
Since you seemed interested @meichthys, I'll just show this here... I've got a light/dark theme switch working and in my local fork users will be able to choose the default theme with a `shareTheme` attribute. I should really put all of this in a draft PR so people can see as I go and @zadam can tell me no before I try to implement something drastic 😄 

https://github.com/zadam/trilium/assets/6865942/913a364e-8922-4a74-ad86-e025a354d4e0

zadam:
Both the template and the [https://trilium.rocks/](https://trilium.rocks/) look great!

I think that providing a rich template like that out of the box would be great. But I also think that there should be a very plain one available (similar to what we have now) for easy/deep customization without having to deal with a lot of bloat. In the end I think we might want to offer both, although it's not immediately clear to me in what form should these be switchable.

Regarding the template - so far my thinking was that this shouldn't be necessary - the default template already provides everything needed (note tree, title, content) and all the extra customization can be done with the combination of linked JS and CSS (including inserting new UI elements etc.). It's possible that some very specific things would still need a custom template, but I'd rather steer most of the customization towards the more standard JS/CSS rather than niche EJS and very specific Trilium template contract. But my thinking is not very fixed in this regard and I'm open to counterarguments.

rauenzi:
Thanks! I'm definitely happy with how they came out.

[...]

Issue #4282, opened 2023-09-22
(Bug report) ~shareCss inheritance doesn't work


deajan:
### Trilium Version

0.61.7-beta

### What operating system are you using?

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 10 22H2 x64

### Description

Using a specific stylesheet via `~shareCss` relation works when I declare it in the OwnedAttributes of the note I intend to share.
But if I declare that relation in a parent note and add inheritance, the inherited attribute will show up properly, but the CSS won't be applied nor loaded (navigator console doesn't show any error / missing stuff).

Inherited CSS from 
![image](https://github.com/zadam/trilium/assets/4681318/5a55830f-9672-4766-8400-975dce4535a2)

Of course, I've shared the css note and set it's type as CSS.

deajan:
Still happens on 0.61.8-beta. Sorry to bother, but could be my "killer feature" when sharing my work ;)

rauenzi:
I tested this on 0.61.8-beta with no issues. Can you share some more details on your configuration? How does the attribute look when applied directly?

deajan:
Hmmm, I'd love this to be an "error 40"...

I've created a note at root of trilium like;
![image](https://github.com/zadam/trilium/assets/4681318/b34d5a32-a2b4-4ba2-aab1-b38a4dce2aaa)
Once shared, the root note looks like this:
![image](https://github.com/zadam/trilium/assets/4681318/393c55df-8486-418f-8a47-2da32fa5e7fc)

When I create a subnote of the root note, the `~shareCSS` attribute gets inherited:
![image](https://github.com/zadam/trilium/assets/4681318/64efa447-3c5b-4daf-a740-42bd4f0a5729)
But the share looks like plain share without any CSS:
![image](https://github.com/zadam/trilium/assets/4681318/7016938f-3962-47b7-8ec3-49e6812e5253)

deajan:
I wonder if I have other problems with my database which could explain this, because it doesn't happen on all my notes.
Some inherit css (the same css file) properly.

There are some notes I can't share.
I'll get a 404 not found once shared, or even a problem with sync (#4297).

I guess I have a "bigger" problem with my database (or note content ?), which consistency check / fixes don't find.
I must yet see if exporting, emptying the server, importing my notes in a fresh trilium database could resolve my issue.

rauenzi:
[...]

PR #4298, opened 2023-09-30
Add ~shareTemplate and #shareExternalLink

rauenzi:
This PR does 3 things, I can remove them or split them as up as needed.

### 1. `~shareTemplate`

This implements custom use share templates as I mentioned here https://github.com/zadam/trilium/issues/3429#issuecomment-1733038937. A user can set `~shareTemplate(inheritable)=xxx` to use their own template for the entire tree, and any subnote to the `~shareTemplate` will be a valid `include()` for EJS. It also automatically falls back to use the default view in the case that rendering with the user template fails so something is still shown.

### 2. `shareExternalLink`

This is a simple label attribute that allows a note in the share tree to act as an external link to an external website. Useful for referencing other sites or documentation withing your own share tree or even using the external link note as a single-source of truth and link to it internally from other notes.

### 3. `/share/api/search`

This is the one I'm least confident in and I'm sure changes will be required, but it adds a search api for the share tree for future use for things like #3430. I tried to follow the standard in `routes.js` but couldn't find any other way but requiring `:noteId` in the api to ensure we were checking the right tree.

I did not modify the current default template or js or css in this PR to support this search functionality (or improve the default as requested in #3429) because I wanted to get this part approved first. I do know that this api works as I'm currently using it (and the other two above) for https://trilium.rocks/

### Question

I asked it briefly in #3429 but it kinda got buried, but are the default templates open for changes? I've been working on improving the default share theme (as shown in that issue) but I'd also like to improve the html/templates to make theming and customizing a little easier, as well as adding in the search functionality from above. But I know that changing those could be considered breaking backwards compatibility for those that have written custom js/css for their shared trees.


src/share/routes.js
```diff
@@ -303,6 +331,37 @@ function register(router) {
...
+        const subRootPath = getSharedSubTreeRoot(note);
+        const subRoot = subRootPath.note;
+        const searchContext = new SearchContext({ancestorNoteId: subRoot.noteId});
+        const searchResults = searchService.findResultsWithQuery(query, searchContext);
+        const filteredResults = searchResults.map(sr => {
+            const fullNote = shaca.notes[sr.noteId];
+            const startIndex = sr.notePathArray.indexOf(subRoot.noteId);
+            const localPathArray = sr.notePathArray.slice(startIndex + 1);
+            const pathTitle = localPathArray.map(id => shaca.notes[id].title).join(" / ");
+            return { id: fullNote.noteId, title: fullNote.title, score: sr.score, path: pathTitle };
```
zerebos: this should be `fullNote.shareId` not `fullNote.noteId`

src/views/share/page.ejs
```diff
@@ -32,7 +32,7 @@
     <%- header %>
     <title><%= note.title %></title>
 </head>
-<body data-note-id="<%= note.noteId %>">
+<body data-note-id="<%= note.noteId %>" data-wtf="true">
```
zerebos: This should be removed (was added during testing)

src/views/share/tree_item.ejs
```diff
@@ -1,10 +1,16 @@
+<%
+const isExternalLink = note.hasLabel('shareExternalLink');
+const linkHref = isExternalLink ? note.getLabelValue('shareExternalLink') : `./${note.shareId}`;
+const target = isExternalLink ? ` target="_blank" rel="noopener noreferrer"` : '';
+console.log(note.shareId, note.title, isExternalLink, linkHref, target);
```
zerebos: This should be removed (was added during testing)

src/share/routes.js
```diff
@@ -303,6 +331,37 @@ function register(router) {
...
+    router.get('/share/api/search/:noteId', (req, res, next) => {
+        shacaLoader.ensureLoad();
+
+        let note;
+
+        if (!(note = checkNoteAccess(req.params.noteId, req, res))) {
+            return;
+        }
+
+        const {query} = req.query;
```
zerebos: Should `search` be used instead of `query` to better match the ETAPI? Only issue is that the ETAPI uses just `/notes` as an endpoint with a query string, and this one uses `/search/:noteId` since a note id is needed but would overlap with `/notes/:noteId` above.

One possible solution: Switch this endpoint go be `/notes` and require the querystring to include `search` and another parameter such as `context` that provides the current note id as context for the search.

src/share/routes.js
```diff
@@ -128,18 +132,42 @@ function register(router) {
         }
 
         const {header, content, isEmpty} = contentRenderer.getContent(note);
-
         const subRoot = getSharedSubTreeRoot(note);
+        const opts = {note, header, content, isEmpty, subRoot, assetPath, appPath};
```
zerebos, line 136: This is all currently passed to the user's custom template as well, should we only pass a subset to the user? Or are there other things that would be useful to the user template that are not currently passed?

Either way this creates a sort of pseudo-api so it would be good to be certain of what should be passed and try not to change it much going forward. But I would still call user templates an "at your own risk" type of extension.

src/share/routes.js
zerebos: Also as per the CodeQL check, this should probably be asserted to not be an array.

deajan:
@rauenzi Care to explain what shareTemplate does ?
Would it resolve my share parameter inheritance  issue #4282 ?

rauenzi:
@deajan It's explained briefly the in the first post, but it is unrelated to your issue #4282 and would not fix it

zadam:
I started looking at this, but it will likely take some time ...

rauenzi:
Take all the time you need, I'm not in a rush

src/share/routes.js
zadam: The current form `/search/:noteId` seems quite cryptic ...

How about: `/notes?query=...&ancestorNoteId=...`?

`ancestorNoteId` defaults to `_share`, but this value is allowed only when there's a `shareIndex` (which basically means that the user wishes that visitors may easily see all shared notes).

src/share/routes.js
```diff
@@ -303,6 +331,37 @@ function register(router) {
...
+        const {query} = req.query;
+
+        if (!query?.trim()) {
+            return res.status(400).json({ message: "'query' parameter is mandatory." });
+        }
+
+        const subRootPath = getSharedSubTreeRoot(note);
+        const subRoot = subRootPath.note;
+        const searchContext = new SearchContext({ancestorNoteId: subRoot.noteId});
+        const searchResults = searchService.findResultsWithQuery(query, searchContext);
```
zadam: TBH I was quite uncomfortable with this reusing of "core" Trilium logic.

Sharing functionality is intentionally kept very separated from the main codebase and especially from the data (becca) to minimize the risk of bugs in the main codebase leaking data to the public internet. This basically means that all non-trivial logic is duplicated in the share codebase and this would normally need to happen for search as well.

But since I'm not keen on writing share-only search functionality, the second look doesn't seem so bad - we're taking only the `noteId` from the search and then data retrieval still happens only through `shaca` (which contains only shared notes). Thus even with buggy search code, there shouldn't be any leak of non-shared data.

zadam (changes requested):

src/share/routes.js
zerebos: Where would the client-side JS get the ancestorNoteId/subroot to provide to the API?

src/share/routes.js
zerebos: I did think about the exposure risk of using the general search functionality but as you said since everything is filtered through `shaca` there should be no results passed back to the client-side JS that are not already shared. Plus using the `SearchContext` should already limit our results fairly well.

src/share/routes.js
zadam: `subRoot` is given to the template ...

If you want to trigger the search via standard form then something like this:

```html
<form action="notes" method="get">
  <input type="text" name="query">
  <input type="hidden" name="ancestorNoteId" value="<%= subRoot.noteId %>">
</form>
```

If you want to drive this from JS, you can either render it to JS variable or perhaps body element attribute ...

zerebos: Ah okay, just wanted to make sure this wasn't already passed somewhere and I had missed it. Didn't want to go adding anything extra without cause.

rauenzi:
Do you want me to split off the search API stuff into its own PR to be evaluated separately from custom templates/external links?

zadam:
> Do you want me to split off the search API stuff into its own PR to be evaluated separately from custom templates/external links?

No, not needed. Seems OK to me ...

rauenzi:
Alright I'll update this PR with requested changes then

zadam:
@rauenzi pls let me know once you're done with the changes...

rauenzi:
@zadam that last commit should address everything we talked about

rauenzi:
@zadam Sorry didn't realize there was also a merge conflict, I've resolved it so this should be ready to review again

deajan:
@rauenzi Still would love to know how ~shareTemplate works, looks great from what is said in first post.

rauenzi:
Hi @deajan not sure what is unclear from the first post but I will try to restate. `~shareTemplate` allows you to use an Embedded JavaScript (EJS) note within your share tree as the template/html for your shared note(s). Any EJS child notes of the target templates will be passed to the EJS `include` function.  As an example the default one used by Trilium can be seen here: https://github.com/zadam/trilium/blob/master/src/views/share/page.ejs

dvai:
@rauenzi You are amazing! I used the `~shareTemplate` feature to create a personal blog template (your website https://trilium.rocks/ gave me great inspiration, thank you). Here is my personal blog address: https://ankia.top/. Feel free to visit.

rauenzi:
Glad to hear it worked out @dvai!

mbseid:
Hey @rauenzi , Thanks for adding this capability. It's exactly what I would like to do. I'm an error when trying to use it. When i click the share link it gives this error: `{"message":"Cannot read properties of undefined (reading 'type')"}`. 

here is the attribute on the parent note. 
image

image


Is there anything I'm missing? Thanks for the help!

Mike

mbseid:
Thanks for the fast response. Here is the ejs template. 
image

Is there anything I need to do to the note to set the type? Additional context, I installed Trilium yesterday and I'm very so there may be concepts I'm missing.

Thanks again

rauenzi:
The template should have the EJS type not plaintext. You might have to go into settings and enable that language. I'm on my phone so I cannot provide screenshot guidance.

mbseid:
Setting the template language to EJS worked for me. Additionally I had to move the template file in the shared folder. Thanks for the help.

Issue #4335, opened 2023-10-18
(Feature Request) Improved mobile view


BeatLink:
### Describe feature

I've created some scripts to rework how the mobile view in trillium functions. These scripts allow mobile users to use the full capabilities of the trillium web in desktop mode while still de-cluttering the interface and increasing the size of elements for larger screens. The scripts can be applied as is to current versions of trillium but perhaps it could be incorporated into the core mobile view of Trilium. 

Give it a try and let me know what you think!

https://github.com/BeatLink/trilium-scripts/tree/main/Mobile%20View

BeatLink:
Yeah, the sidebar view is intended to work with the sidebar kept open

LokiLuciferase:
This is exactly what I needed from the mobile view, thanks!

I think passing the note IDs used in `SetupButton.js` via relations (`api.currentNote.getRelationValue`) is cleaner than putting the actual IDs in the code, but other than that, it worked out of the box like a charm!

PR #4371, opened 2023-11-01
Update webmanifest for wider compatibility

### LINKED ISSUE #920: PWA Support
Is there PWA support in the plans for Trilium?
I see a lot of potential for custom server deployed solutions and relatively simple to implement.

Possibly Even Utilize the features service worker provide

rauenzi:
Partially fixes #920, just does not enable offline usage.

### DIFF
--- a/src/public/manifest.webmanifest
+++ b/src/public/manifest.webmanifest
@@ -1,14 +1,22 @@
 {
-  "name": "Trilium",
+  "name": "Trilium Notes",
   "short_name": "Trilium",
-  "theme_color": "DarkGray",
-  "background_color": "DarkGray",
+  "description": "Trilium Notes is a hierarchical note taking application with focus on building large personal knowledge bases.",
+  "theme_color": "#333333",
+  "background_color": "#1F1F1F",
   "display": "standalone",
-  "scope": "./",
-  "start_url": "./",
-  "icons": [{
-    "src": "assets/vX/images/app-icons/ios/apple-touch-icon.png",
-    "sizes": "180x180",
-    "type": "image/png"
-  }]
+  "scope": "/",
+  "start_url": "/",
+  "icons": [
+    {
+      "src": "/assets/vX/images/app-icons/ios/apple-touch-icon.png",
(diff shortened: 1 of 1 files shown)


rauenzi:
By the way, let me know if you want me to work on offline usage for PWA, after toying around with it some is may be *mostly* possible.

zadam:
> By the way, let me know if you want me to work on offline usage for PWA, after toying around with it some is may be mostly possible.

Sure, although I'm curious what that "mostly" means :-)

rauenzi:
Some caveats would be:
1. If the user goes offline there would be no way for them to get/open existing notes that have not already been cached in some way. (i.e. viewing the note when they were online)
2. Any functionality not already cached like the different language extensions for codemirror will be lost during offline mode.
3. Users would not be able to create new notes

Number 1 can't really be worked around without explicitly attempting to cache every single note on first install, which is technically possible but can also lead to a lot of stale information.

Number 2 can be worked around by pre-caching all the _potential_ resources like that though it is a bit of a pain to do.

Number 3 can't be worked around without actual changes to Trilium.


So either way I think a network-first approach is the right one for Trilium, and combine this with some updates to the mobile interface (#3975, #4335) I think that would alleviate a lot of the calls for a "proper" mobile app.

zadam:
> So either way I think a network-first approach is the right one for Trilium

Agreed.

dahamsta:
Thanks for this. Would I be right in saying that this wasn't rolled into 0.61.12?

rauenzi:
This was merged into main which will become 0.62

dahamsta:
Great, thanks, I'll test as soon as I get it.

rauenzi:
You could modify your existing container to match this change and test for yourself if you're using Docker. Might be worth moving this part to a discussion though if you're interested.

jacksgt:
Thanks for this awesome work! I can confirm that it works with Trilium v.0.62.3 on Firefox for Android. :clap:

dahamsta:
Same, lovely job, thanks.

PR #4397, opened 2023-11-06
Add frontend note prompt API

### LINKED ISSUE #4395: (Feature request) Autocomplete dialog support via scripting API
### Describe feature

I am currently trying to write a kanban board script extension. This will support rendering a note as a Kanban board.
One of the features I would like to implement is making it easy for users to link a specific kanban board item to a note in their library. For this, it would be great to leverage the `noteAutocompleteService`. Something as simple as `api.getNoteAutocomplete()` that resolves with a FNote object for the front-end API would be great.

What are your thoughts on implementing this? Should I attempt a pull-request to add this functionality?

### Additional Information

_No response_

benkaiser:
We would like to allow users with custom scripts to leverage the internal note selector interface. This allows for easily creating links to internal notes from custom rendered views.

Fixes #4395 .

### DIFF
--- a/src/public/app/widgets/dialogs/add_link.js
+++ b/src/public/app/widgets/dialogs/add_link.js
@@ -44,7 +44,7 @@ const TPL = `
                             <br/>
                             <label>
                                 Link title
-                                
+
                                 <input class="link-title form-control" style="width: 100%;">
                             </label>
                         </div>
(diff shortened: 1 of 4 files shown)


src/public/app/widgets/dialogs/add_link.js
```diff
@@ -77,30 +77,52 @@ export default class AddLinkDialog extends BasicWidget {
 
                 const linkTitle = this.getLinkType() === 'reference-link' ? null : this.$linkTitle.val();
 
-                this.textTypeWidget.addLink(this.$autoComplete.getSelectedNotePath(), linkTitle);
+                if (this.textTypeWidget) {
```
benkaiser, line 80: This check appears to be needed because in the context of a script extension theremay not be a textTypeWidget.

src/public/app/widgets/dialogs/add_link.js
```diff
@@ -77,30 +77,52 @@ export default class AddLinkDialog extends BasicWidget {
...
+                    this.textTypeWidget.addLink(this.$autoComplete.getSelectedNotePath(), linkTitle);
+                }
+                if (this.resolve) {
+                    this.resolve({
```
benkaiser, line 84: Thoughts on this return structure @zadam ?

src/public/app/widgets/dialogs/add_link.js
```diff
@@ -121,7 +143,7 @@ export default class AddLinkDialog extends BasicWidget {
         };
 
         noteAutocompleteService.initNoteAutocomplete(this.$autoComplete, {
-            allowExternalLinks: true,
+            allowExternalLinks: false,
```
benkaiser, line 146: @zadam when using the text editor control+L (i.e. using this dialog directly) I was never actually able to get an external link selected. Are we best off disabling it here, or should we fix external link selection inside the note selector?

zadam:
@benkaiser I don't like that the Add Link dialog now does two things - it inserts the link into the text note AND resolves the link. What if you just want the latter and not the former? Add Link dialog is also more specialized than mere selecting note (it allows defining a different link title)

What I thought you want to do is to create a pure note selection dialog ...

Alternatively, the Add Link dialog could be generalized in the sense that it would just resolve the selected note / title, and then the triggering code would handle inserting the content into the text note.

Issue #3431, opened 2022-12-17
Migrate frontend dependencies to npm


zadam:
### Describe feature

Similar to excalidraw.

CKEditor, CodeMirror and FancyTree might be exceptions since they have modifications

rauenzi:
> CKEditor, CodeMirror and FancyTree might be exceptions since they have modifications

You could fork these repos, or just host your edited version on github and still use npm to install via `npm install zadam/ckeditor` and similar. That would allow things to act separately while still using modified versions.

zadam:
Yep, that would make sense as well.

rauenzi:
I can help with the others, but those I'll have to leave to you since they'd need to be under your account or a GitHub org.  :smile:

eliandoran:
We've solved all hard-coded dependencies.

PR #4417, opened 2023-11-08
Migrate KaTeX@0.16.0 from `libraries` to KaTeX@0.16.9 in `node_modules`

ljgonzalez1:
Bumps up KaTeX version from 0.16.0 to 0.16.9
Just replaced the original files in [/libraries/katex](https://github.com/zadam/trilium/tree/master/libraries/katex) for the ones in the [0.16.9 release](https://github.com/KaTeX/KaTeX/releases) ([KaTeX Changelog here](https://github.com/KaTeX/KaTeX/blob/main/CHANGELOG.md))

This is mostly beacause the version in trilium was quite old by now and for the bug fixed in 0.16.3 which was bothering me:

> [v0.16.3](https://github.com/KaTeX/KaTeX/releases/tag/v0.16.3)
[0.16.3](https://github.com/KaTeX/KaTeX/compare/v0.16.2...v0.16.3) (2022-10-22)
Bug Fixes
\hline after \cr (https://github.com/KaTeX/KaTeX/issues/3735) ([ebf6bf5](https://github.com/KaTeX/KaTeX/commit/ebf6bf5b50a98ac6c5aca1896c0a6ba985c1c91c)), closes https://github.com/KaTeX/KaTeX/issues/3734


rauenzi:
This could also be a good opportunity to try to move it to an npm dependency #3431

ljgonzalez1:
@rauenzi 
Sorry, not that experienced with node. Or js. Just learning actually.

I think I can try though.

Would add the package to the `packages.json` file (and the lockfile)  and change the import method work?

I'll test if of course, but if ithere is more to do apart from that, it would be nice to know.
Also, can you link me to another pr where they changed a library import to a package import so I can use it as an example?

ljgonzalez1:
Also took the opportunity to update the `.gitignore` file as it was a bit leftout.
Not sure if it is a good idea to combine a couple of things in a single _pr_ (my first attempt at contribution to FOSS), but I'll leave it with the idea it helps.

Worst case you tell me and I make a rollback / undo

---

Here are the changes to the .gitignore:
- Grouped al files and folder by category
- Left comments about what is ignored or un-ignored
- Ignored/Un-ignored common files for:
  - OS:
    - Mac OS
    - Linux (+ KDE .directory)
  - IDEs/Editors
    - VS Code, Visual Studio Code, VS Codium
    - Most of the Jetbrains IDE (ignored almost the whole folder /.idea as is usually not needed)
    - Kate

rauenzi:
> can you link me to another pr where they changed a library import to a package import so I can use it as an example?

I don't know of one off the top of my head. As for how to move it to npm, it's probably slightly more involved than just adding to `package.json` since it needs to be properly included with the server files on build.

> Not sure if it is a good idea to combine a couple of things in a single pr

I won't speak for zadam here obviously, so I'll just say in my experience, usually it's kept to one feature/fix per PR otherwise unrelated pieces can hold back things that are otherwise ready to roll.

That said, to me this change to `.gitignore` feels unneeded and very unrelated. There might be a couple lines that could be useful but most seem to not really apply to Trilium imo.

ljgonzalez1:
>  won't speak for zadam here obviously, so I'll just say in my experience, usually it's kept to one feature/fix per PR otherwise unrelated pieces can hold back things that are otherwise ready to roll.

Removed the `.gitignore` stuff I did.

Almost there with the migration.

ljgonzalez1:
> This could also be a good opportunity to try to move it to an npm dependency #3431

Woohoo.

Done!! (I hope so)

---

No is no longer an independent module which has to be updated manually.
Replaced all the paths and references so it imports the `node_module`'s version instead of the `libraries`'s version.

Tested with everything I thought of to make sure nothing was (more) borken.

It works in the editor
image

And in shared notes
image

zadam:
That's great, thank you!

PR #4418, opened 2023-11-08
Add Windows installer (fixes #2081)

### LINKED ISSUE #2081: [Feature request] Installer for windows
It's uncomfortable to update on windows with current setup(zip archive).

eliandoran:
- Using [Electron Forge](https://www.electronforge.io/) instead of using directly tools like [Squirrel.Windows](https://github.com/Squirrel/Squirrel.Windows) since it will allow us in the future to build more variants such as .dmg. And maybe even simplify the existing build system.
- Removed `electron-build` dependency since it doesn't seem to be used, see commit for more info.

Still to be done:

- Set up automation using GitHub Actions.
- Document the build process.


eliandoran:
Too many modifications in `package-lock.json`; maybe I'm running a different version of Node than @zadam . Using Node 18.15 since that seems to be the matching version that the current Electron version (25.9.3) being used by Trilium.

eliandoran:
@zadam , @rauenzi ,

Here's a video showing how the Windows installer looks like.

Let me know your thoughts.

https://github.com/zadam/trilium/assets/21236836/f2f22ce4-3340-476e-bae4-d428e8f5fec1

rauenzi:
I was just testing this out locally myself as well, it seems to work very smoothly and install as expected! I may not be a fan of the light-theme banner, but I'm definitely a fan of this overall 😄 Am I right that this banner can be an animated `gif` (and potentially transparent)? If so, I might work on a fancy loading animation.

It's a very seamless transition from install to running and it doesn't have the traditional Windows installation prompt which I personally think is nice. But there are a couple questions I can see coming up from end-users because of this:
- Where is it installed? (`%localappdata%\trilium`) & can I change where it is installed?
- Can I prevent it from adding to start menu?
- Can I prevent it from adding a desktop icon?
- How can I prevent it from updating automatically? (some users won't want to update immediately especially with sync servers)

I think this is a huge step in the right direction and also opens the door to other types of distributions for mac and linux. It can even help simplify the current build process.

eliandoran:
> I may not be a fan of the light-theme banner, but I'm definitely a fan of this overall 😄 Am I right that this banner can be an animated `gif` (and potentially transparent)? If so, I might work on a fancy loading animation.

You're right, by default it's an animated GIF. I just replaced it with something I made really quick in order not to have the default. Feel free to open a PR on top of mine to change it. I would really like it if you kept the "original format" of your animation as well (for me it was a GIMP file, for you it might be something Adobe-related).

> It's a very seamless transition from install to running and it doesn't have the traditional Windows installation prompt which I personally think is nice. But there are a couple questions I can see coming up from end-users because of this:

Yes I like that as well.

>     * Where is it installed? (`%localappdata%\trilium`) & can I change where it is installed?

I would say this would not be a very common question for most users, however we can document the path. There is always the possibility to have an MSI installer for those power users that would prefer to have a custom installation dir (including installing in Program Files instead of locally).

> 
>     * Can I prevent it from adding to start menu?
>     * Can I prevent it from adding a desktop icon?

Same here, maybe the MSI installer will allow this kind of customization. I'll see what I can do, nevertheless I'd prefer to have a separate PR for it.

> 
>     * How can I prevent it from updating automatically? (some users won't want to update immediately especially with sync servers)
> 

The way I see this is that the "squirrel" installer is going to be the most frictionless: no prompts, no configuration, just plug-and-play. I think this will account for at least 90% of users. And this includes automatic updates (which I haven't set up yet). The reasoning for that is that some people prefer things to _just work_, I have colleagues at work that use Trillium Notes on Windows but they are 5-7 versions behind, because... it just works. Background updates will solve this problem. For the remainder of users, MSI installer.

> I think this is a huge step in the right direction and also opens the door to other types of distributions for mac and linux. It can even help simplify the current build process.

Absolutely, if/after we get this merged, I'll try to do the same for macOS (.dmg), Linux (.deb/.rpm) and maybe even have a look to see if we can improve flatpak/snap build system.

rauenzi:
> Feel free to open a PR on top of mine to change it. 

Will do once I get some time.

> There is always the possibility to have an MSI installer for those power users

Yeah that's probably a good idea for those users. I think the vast majority of end users will want all those things by default, and just the power users will want customization.

> Background updates will solve this problem. 

I agree mostly, but I know there are users that use like the paid trilium.cc service where sudden unexpected background updates can cause sync to break and they may not be ready (or able) to upgrade their server instance as well. I think sync is fairly common for a lot of users.

eliandoran:
Closing in favor of https://github.com/TriliumNext/Notes/issues/22.

Issue #4416, opened 2023-11-08
(Feature request) Provide a attribute to disable the Highlights List Widget in a text note.


fang64:
### Describe feature

Hello, I am not sure anyone else would want this, but I have a case where I do not want that widget to show in a particular note since all I am doing is formatting the text layout and it doesn't extend beyond a full screen. 

I imagine it'd be a simple attribute highlistslistwidgetDisable I looked in the code base for anything that would explicitly disable it on a note type of text, but it is apparently enabled by this. So I could in theory change the note type and it wouldn't apply. Although it would be nice to just disable it entirely from displaying on that note.

lndjenkinson:
I am also someone who wants this. There is an X to close it, but it's like whack-a-mole to find the right spot to click. Also, clicking the x only lasts until the next time you access the note. Then it's wack-a-mole again. I'd lie to see someway to add an attribute to disable that, same as you can disable the scratch pad or the TOC widget.

fang64:
@lndjenkinson hopefully we get this added :smiley:

lndjenkinson:
Zadam is very good at getting the kinks out. I'm sure he'll work on it.

olsonpm:
I came here looking for a way to disable it from popping up.  For my use-case a global setting to disable it suffices

FliegendeWurst:
> I came here looking for a way to disable it from popping up. For my use-case a global setting to disable it suffices

It can be configured in Options → Text Notes → Highlights List.

FliegendeWurst:
> I looked in the code base for anything that would explicitly disable it on a note type of text, but it is apparently enabled by this.

You can add a `#hideHighlightWidget` label to hide the highlight widget. I think it would be good to document this somewhere. Perhaps in the description of the global setting?

lndjenkinson:
> I came here looking for a way to disable it from popping up. For my use-case a global setting to disable it suffices

Thanks. I was trying disableHighlight and nothing happened.

zadam:
Hi, this label actually already exists - `#hideHighlightWidget`. I've added it to autocomplete and mention it now in the options dialog.

lndjenkinson:
Thank you!

PR #4421, opened 2023-11-10
Add attribute disableHighlightsList to disable Highlights List Widget

fang64:
I'm not sure if this is sufficient but this should add the functionality I was looking for, might need to add a bit of documentation to the wiki about the attribute. But this should be fine.

Addressing my Feature Request.
#4416


zadam:
Actually I reverted the commit. It turns out this functionality already existed in the form of `#hideHighlightWidget` widget.

fang64:
@zadam thanks I didn't know that, this wasn't documented anywhere I could find.

PR #4427, opened 2023-11-10
Create DockerHealthcheck.sh

tilltmk:
Added: DockerHealthcheck.sh


tilltmk:
**Requires curl to be added in Dockerfile!**

PR #4432, opened 2023-11-12
Fixed: DockerHealthcheck.sh

tilltmk:
Fixed "unhealthy" container status in docker

### DIFF
--- a/DockerHealthcheck.sh
+++ b/DockerHealthcheck.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+URL=${TRILIUM_URL:-"http://localhost:8080"}
+
+if curl -s "$URL" > /dev/null; then
+    echo "Trilium is up and running."
+    exit 0
+else
+    echo "Trilium is not responding."
+    exit 1
+fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN set -x \
         nasm \
         libpng-dev \
         python3 \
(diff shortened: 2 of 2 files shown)


rauenzi:
But Trilium already has a healthcheck that works? Here's how it looks for me

image

image

And here is where Trilium sets up the Healthcheck https://github.com/zadam/trilium/blob/master/Dockerfile#L42 and the actual implementation here https://github.com/zadam/trilium/blob/master/docker_healthcheck.js

tilltmk:
yes, i know it has a js Healthcheck; but it didn't work, in portainer the container status was shown as 'unhealthy'. i searched online for the problem and some people had issues with it as well. maybe it is because of environmental configurations, but the DockerHealthcheck.sh is necessary for the container in some peoples env to be shown as healthy.

tilltmk:
referencing: https://github.com/zadam/trilium/issues/3665

rauenzi:
I'm just saying it feels unnecessary to have two separate healthchecks especially when the one is already working for the majority of cases. Obviously it's not an incompatibility with docker or portainer in general as I've just shown both, so there's some other issue that could/should be fixed, or removed in favor of a different health check.

zadam:
As @rauenzi said, it's much better to figure out why the existing health check doesn't work and fix that.

The health check you're adding is kind not integrated into the app at all - it refers to non existing `TRILIUM_URL` env. variable etc.

Issue #4412, opened 2023-11-08
(Bug report) Upgrade caused minor corruption in some Chinese character note titles


Nriver:
### Trilium Version

0.61.13

### What operating system are you using?

Windows

### What is your setup?

Local + server sync

### Operating System Version

win10

### Description

Some user tells me that he got some note title broken after upgrade. I checked my own notes and found that a handful of note titles were slightly corrupted, about 30 out of 15,000.

During my investigation, I came across the peculiar character `�` within these titles.

![ksnip_20231108-170321](https://github.com/zadam/trilium/assets/6752679/6c0b9f05-1c80-4918-82cb-52a094ec4a21)

Notably, all the affected titles contained Chinese characters, although the original characters varied.

zadam:
That's strange ... in the 0.61 upgrade, there was quite a lot of data migration, but specifically note titles were not touched at all.

The only thing I can think of is the better-sqlite3 upgrade (containing sqlite upgrade) and electron upgrade, perhaps some Unicode support change / bug.

Nriver:
After numerous rounds of testing—literally hundreds—I have pinpointed the root cause and successfully resolved the issue. In summary, the culprit was the automatic encoding conversion, which was causing the problem.

Nriver:
I utilize [mitmproxy](https://github.com/mitmproxy/mitmproxy) to monitor HTTP requests for data initialization with an existing server.

The data flow is structured as follows: server -> mitmproxy -> client. Notably, incorporating mitmproxy has significantly reduced the occurrence of corrupted notes during initialization from 30-40 to 3. Additionally, the random corruption observed across all notes has become consistent for these three notes.

Upon conducting debugging, I discovered that the `entity_changes` component is receiving corrupted data.

![ksnip_20231213-173505](https://github.com/zadam/trilium/assets/6752679/9aad2d8e-f49e-4b05-af78-d125033824f4)

However, the data remains intact within the mitm proxy, suggesting that the server sends correct data, but the client somehow fails to receive it accurately.

![ksnip_20231213-174438](https://github.com/zadam/trilium/assets/6752679/a20fe39d-f49c-44b8-85c9-3e720526c4f8)

Further debugging led me to identify a suspicious code section:

![ksnip_20231214-094259](https://github.com/zadam/trilium/assets/6752679/55c21815-3f39-4a28-9dd1-6f1a176f9800)

[...]

PR #4522, opened 2023-12-14
fix decoding issue for request data chunks

Nriver:


Nriver:
This fixes #4412

zadam:
Awesome, great catch! Although I don't understand why this would cause problems specifically in 0.61 and later.

Nriver:
Perhaps this issue has existed in earlier versions, but I hadn't noticed it until someone pointed it out to me. After being informed about the issue, I started checking my data and confirmed its presence. Additionally, the occurrence of this issue is probabilistic—it becomes more likely as the data reaches a certain quantity.

PR #4624, opened 2024-01-30
Add ordinalPadded day title generation template variable

ProfDoof:
Heyo!

I really like this project and want to start contributing in small ways as I have time. I was hoping to add support for an ordinalPadded generation template variable as currently regular ordinal won't sort correctly which is not ideal.


ProfDoof:
Also, this is a random idea I had while I was looking at the code, but would you possibly be interested in integrating a template language in at some point like mustache or jinja to allow people to integrate more complex and powerful templates? 

If so, I might be interested in adding support for it as I have time but I wanted to check on if such an idea would be welcome first.

PR #4676, opened 2024-03-05
fix: invisible unchecked to-do items

st3iny:
Fix https://github.com/zadam/trilium/issues/4672

| Before | After |
| --- | --- |
| grafik | grafik |


zadam:
Thanks!

Issue #3302, opened 2022-11-09
(Feature request) I want a version for mac m1 chip


supermanmsc:
### Describe feature

(Feature request) I want a version for mac m1 chip

### Additional Information

(Feature request) I want a version for mac m1 chip

zadam:
Unlikely to happen soon. If I understand correctly, Apple tightened the screws and M1 apps require notarization (besides also needing to be built on M1). From my experience, the Intel version works decently well through Rosetta.

supermanmsc:
> Unlikely to happen soon, if ever.
> 
> If I understand correctly, Apple tightened the screws and M1 apps require notarization (besides also needing to be built on M1). From my experience, the Intel build works decently well through Rosetta.

Thanks for you attention ,   i hava a core dump case.  either caused by m1 or not . 
https://github.com/zadam/trilium/issues/3307

Myzel394:
@supermanmsc Trilium works mine for me on my M1 mac (2020 version) - Did you try running it with and without rosetta?

delight:
> Unlikely to happen soon, if ever.
> 
> If I understand correctly, Apple tightened the screws and M1 apps require notarization (besides also needing to be built on M1). From my experience, the Intel build works decently well through Rosetta.

Could you please elaborate on what you mean by tightened screws and required notarization. There are galore open source projects and commercial programs available for Apple Silicon by now. As I understand this is an electron app. This should be a big problem to get this working on an M1/M2. I uninstalled every non native app by now ... some things are missing but (at least for me) its way less then 5% of apps that are not available in a native Version. Kodi being one of them and they plan to support M1 in the next version. Technically this project should be portable way easier.

delight:
I think the apple certificate is only need if you want to distribute the app via the app store. It's not the case for simple downloads from a webpage or installations via brew. There are big and small  open source projects without certificates. I could list you some or you could have a look at the list of brew apps.

Obviously there is this button in system settings where you have to agree to run the app even though it's not certified. I even disabled that hate keeper completely because I know that I'm doing and it's annoying otherwise to confirm every second app.

delight:
> Of course, if somebody is willing to do this (i.e. do the release builds and notarization), then that would be awesome!

I would like to give it a try to build the binaries.

@zadam do you have some build instructions or can give me some here ?

zadam:
You can use `bin/build-mac.sh` as a reference

delight:
I think I need a little bit more then that 

[...]

Issue #3125, opened 2022-09-13
(Feature request) Add Apple Silicon release too to the releases section in next build


coolerspeex:
### Describe feature

Add Apple Silicon release too to the releases section in next build

zadam:
This is probably not going to happen soon. AFAIK building for M1 essentially needs to be done on Mac which would complicate release process dramatically. There are some more difficulties like mandatory notarization.

 In my experience, the emulation works quite well, so this does not seem pressing.

PR #4772, opened 2024-05-18
Support compiling native MacOS arm64 (Apple Silicon) builds

gigamonster256:
Following along with the original request in #3125 and the followup plus community troubleshooting in #3302, I was able to successfully build a working native arm64 build of trilium for MacOS on my M1 mac as well using my Windows 11 wsl environment.

Using the native package rather than relying on the Rosetta translation layer leads to a much better startup time (a couple seconds at most vs 10-20 seconds before) and a much snappier experience. (as originally reported in #3302)

A user in #3302 voiced concerns about an arm64 native build being bigger than the x86-64 build, but their observations seem to come from using a different packaging system than trilium uses for its other builds. I replicated the current build flow for arm64 and the resulting binaries are comparable to the released build sizes. (See below)

Screenshot 2024-05-18 at 3 38 18 AM I built both the latest and second latest versions of trilium to test. The top row of applications was compiled by me (both intel and arm builds) and the bottom are from the trilium releases. 

Concerning future maintainability, better-sqlite3 does not have prebuilt binaries for version 8.4.0 with electron 25 so I compiled that combination myself for macos arm64 and added it to the [precompiled binaries used for trilium releases](../blob/master/bin/better-sqlite3). If in future releases of trilium the version of electron and/or better-sqlite3 is changed, precompiled binaries will most likely be available through better-sqlite3 [releases](https://github.com/WiseLibs/better-sqlite3/releases), removing the need for an Apple Silicon computer to keep arm64 builds up to date.


gigamonster256:
> Get work! Do you have a sense of how difficult it would be to also build a Linux arm 64 build? I attempted it a few times but haven't been successful.

Based on my very limited understanding of electron apps (literally all learned trying to get this working) the only prebuilt architecture dependent component that changes between trilium builds is the better-sqlite3 database bindings. It seems that the main workhorse of the packaging, electron-packager, is able to "cross-compile" everything else (as much as you can compile an electron app.) Unfortunately the bindings of better-sqlite3 8.4.0 with electron 25 (NODE_MODULE_VERSION = 116 and what trilium currently uses) are not available as binaries from the developers so I had to build it on a native platform to get maxos arm64 working. It looks like all that is needed to get linux_arm64 working is to make/find a copy of the correctly versioned better_sqlite3.node file for linux_arm64 and add/modify the build scripts like I did in this pull request to add it.

Interestingly enough, it seems that electron-packager does not support building on linux_arm64 but does support it as a target so you'd need to build on a supported system then bring the zip over to the arm linux device.

zadam:
That's great, thanks for the contribution. I was under impression that building an M1 electron package required building on M1.

PR #4820, opened 2024-07-23
Add syntax highlight using highlight.js for codeblocks

J-141:
Added highlight.js as a library.

Use highlight.js on top of CKEditor to syntax-highlight codeblocks in real-time.


eliandoran:
@J-141 , your work here seems pretty promising. Maybe we could have it for https://github.com/TriliumNext/Notes.

Thanks for keeping it up to date.

PR #5832, opened 2025-06-21
Feat/clean share url

eliandoran:
Continuation of https://github.com/TriliumNext/Notes/pull/1781.


JYC333:
@eliandoran This is ready for reivew.

@pano9000 FYI, if you have time would like to check it also.

Brramble:
Really looking forward to this feature, I'm assuming it's also going to fix the current state of "Redirect bare domain to Share page" not working as intended?

eliandoran:
We are no longer going to pursue this avenue of having the server switch to share-first, since we now have a different way to achieve this via static HTML exports. See https://docs.triliumnotes.org/user-guide/advanced-usage/sharing/static-html for more information (not yet released).

eliandoran:
> I'm assuming it's also going to fix the current state of "Redirect bare domain to Share page" not working as intended?

@Brramble , please open a separate issue if this is still problematic.

PR #5834, opened 2025-06-21
feat(ocr): add OCR

perfectra1n:


perfectra1n:
I'll label you a merge conflict next 😂

.github/workflows/playwright.yml
```diff
@@ -35,7 +35,8 @@ jobs:
         run: pnpm install --frozen-lockfile
       - run: pnpm exec playwright install --with-deps
       - uses: nrwl/nx-set-shas@v4
-
+      - name: Install playwright
```
eliandoran: This already done at the previous step. Please remove it.

apps/client/src/widgets/type_widgets/options/images/images.ts
```diff
@@ -44,6 +83,123 @@ const TPL = /*html*/`
...
+        <label class="tn-checkbox">
+            <input class="ocr-auto-process" type="checkbox" name="ocr-auto-process">
+            ${t("images.ocr_auto_process")}
+        </label>
+
+        <div class="form-group">
+            <label>${t("images.ocr_language")}</label>
+            <p class="form-text">${t("images.ocr_multi_language_description")}</p>
+            <div class="ocr-language-checkboxes">
+                <label class="tn-checkbox">
```
eliandoran: Either use the content language settings which are already available or refactor it to be generated dynamically from the list of locales in Commons.

apps/server/src/becca/entities/bblob.ts
```diff
@@ -10,11 +10,12 @@ class BBlob extends AbstractBeccaEntity<BBlob> {
         return "blobId";
     }
     static get hashedProperties() {
-        return ["blobId", "content"];
+        return ["blobId", "content", "ocr_text"];
     }
 
     content!: string | Buffer;
     contentLength!: number;
+    ocr_text?: string | null;
```
eliandoran: Please rename to `textRepresentation ` as per our discussion. Make sure to use camel case.

apps/server/src/migrations/migrations.ts
```diff
@@ -6,6 +6,25 @@
 
 // Migrations should be kept in descending order, so the latest migration is first.
 const MIGRATIONS: (SqlMigration | JsMigration)[] = [
+    // Add OCR text column and last processed timestamp to blobs table
+    {
+        version: 234,
+        sql: /*sql*/`\
+            -- Add OCR text column to blobs table
+            ALTER TABLE blobs ADD COLUMN ocr_text TEXT DEFAULT NULL;
```
eliandoran: Same here, column names should not contain the term OCR as we will reuse the column for other ways to extract the text (for example for PDFs having already the text information instead of simply containing pictures).

apps/server/src/services/ocr/processors/office_processor.ts
```diff
@@ -0,0 +1,128 @@
...
+    async cleanup(): Promise<void> {
+        await this.imageProcessor.cleanup();
+    }
+
+    /**
+     * Get default OCR language from options
+     */
+    private getDefaultOCRLanguage(): string {
+        try {
+            const options = require('../../options.js').default;
```
eliandoran: Please avoid the use of require. Use sync imports if needed.

apps/server/src/services/ocr/processors/image_processor.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Image processor for extracting text from image files using Tesseract
+ */
+export class ImageProcessor extends FileProcessor {
+    private worker: Tesseract.Worker | null = null;
+    private isInitialized = false;
+
+    canProcess(mimeType: string): boolean {
+        const supportedTypes = [
```
eliandoran: Extract outside function.

apps/server/src/services/ocr/processors/office_processor.ts
```diff
@@ -0,0 +1,128 @@
...
+export class OfficeProcessor extends FileProcessor {
+    private imageProcessor: ImageProcessor;
+
+    constructor() {
+        super();
+        this.imageProcessor = new ImageProcessor();
+    }
+
+    canProcess(mimeType: string): boolean {
+        const supportedTypes = [
```
eliandoran: Extract outside function.

eliandoran (changes requested):
In addition, we would need tests for all the processors since they rely on third part libs which are going to get updated by Renovate and we would otherwise have no coverage to ensure they don't break at some point.

.github/workflows/playwright.yml
perfectra1n: This is an artifact of before you had updated this pipeline - I'll remove it :)

developercrocodiles:
This is a really good idea! Hope you get this all done and merged soon. 🙂

perfectra1n:
I can't tell you how happy this feature would make me 😂

PR #6429, opened 2025-07-22
Add grid to canvas

Papierkorb2292:
This pull requests enables the grid toggle from excalidraw, which can be accessed by right clicking on the canvas or by using its shortcut. Whether the grid is enabled for a canvas is also saved along with the other data so it doesn't have to be re-enabled every time the canvas is loaded.


eliandoran (approved):
Looks good.

PR #6577, opened 2025-08-07
Adds duplicateSubtree to backend API.

Geekswordsman:
As discussed previously on Matrix; this PR adds in the ability to call api.duplicateSubtree on the backend.

It also removes the unnecessary null check on origBranch in the duplicateSubtree process - this check prevents the backend API from actually functioning and is not needed.


apps/server/src/services/backend_script_api.ts
```diff
@@ -412,6 +412,13 @@ export interface Api {
      */
     backupNow(backupName: string): Promise<string>;
 
+    /**
+     * Exposes the backend duplicateSubtree to the API
```
eliandoran: The documentation should reflect what the method does, exposing a backend function is more an implementation detail.

apps/server/src/services/notes.ts
```diff
@@ -931,9 +931,10 @@ function duplicateSubtree(origNoteId: string, newParentNoteId: string) {
 
     const noteIdMapping = getNoteIdMapping(origNote);
 
-    if (!origBranch) {
-        throw new Error("Unable to find original branch to duplicate.");
-    }
+    // Not required - if origBranch doesn't exist, the subtree can still be created.
+    // if (!origBranch) {
+    //     throw new Error("Unable to find original branch to duplicate.");
+    // }
```
eliandoran: Simply remove the code. No reason to have it commented-out.

apps/server/src/services/backend_script_api.ts
Geekswordsman: Updated the conversation as requested.

apps/server/src/services/notes.ts
Geekswordsman: Removed.

eliandoran (approved):

PR #6584, opened 2025-08-08
Added zen mode to mobile layout

Papierkorb2292:
With this change, zen mode is also available through the menu in the mobile layout. While zen mode doesn't provide a huge benefit on a phone, this feature can be useful when working on a tablet by allowing you to have both the mobile controls as well as the option to hide those controls without having to switch back to the desktop layout first.

Note: a future improvement for the usage of the mobile layout on a tablet could be to also allow the user to enable it in the electron app, but this is enough for me for now

### DIFF
--- a/apps/client/src/widgets/close_zen_button.ts
+++ b/apps/client/src/widgets/close_zen_button.ts
@@ -1,5 +1,6 @@
 import BasicWidget from "./basic_widget.js";
 import { t } from "../services/i18n.js";
+import utils from "../services/utils.js";
 
 const TPL = /*html*/`\
 <div class="close-zen-container">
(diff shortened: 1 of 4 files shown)


apps/client/src/widgets/close_zen_button.ts
```diff
@@ -28,6 +29,10 @@ const TPL = /*html*/`\
         -webkit-app-region: no-drag;
     }
 
+    body.zen .mobile-close-zen-container {
+        top: -2px;
+    }
```
eliandoran: No need for a dedicated class name for mobile, you could do `body.mobile.zen`

```suggestion
    body.zen.mobile .close-zen-container {
        top: -2px;
    }
```

apps/client/src/widgets/close_zen_button.ts
```diff
@@ -40,6 +45,7 @@ export default class CloseZenButton extends BasicWidget {
 
     doRender(): void {
         this.$widget = $(TPL);
+        this.$widget.toggleClass("mobile-close-zen-container", utils.isMobile())
```
eliandoran:
```suggestion
```

eliandoran (changes requested):

Papierkorb2292:
Yep that works, thanks for catching that, I didn't see this class before

eliandoran (approved):

PR #6596, opened 2025-08-09
Adds a get/set to bNote to allow getting an Attribute by it's Id, or …

Geekswordsman:
…setting an Attribute's value by it's Id

### DIFF
--- a/apps/server/src/becca/entities/bnote.ts
+++ b/apps/server/src/becca/entities/bnote.ts
@@ -1758,6 +1758,41 @@ class BNote extends AbstractBeccaEntity<BNote> {
         return childBranches;
     }
 
+    /**
+     * Return an attribute by it's attributeId.  Requires the attribute cache to be available.
+     * @param attributeId - the id of the attribute owned by this note
+     * @returns - the BAttribute with the given id or undefined if not found.
+     */
+    getAttributeById(attributeId : string): BAttribute | undefined {
+        this.__ensureAttributeCacheIsAvailable();
+
+        if (!this.__attributeCache) {
+            throw new Error("Attribute cache not available.");
+        }
+
+        return this.__attributeCache.find((attr) => attr.attributeId === attributeId);
+    }
+
+    /**
+     * Sets an attribute's value by it's attributeId.
+     * @param attributeId - the id of the attribute owned by this note
+     * @param value - the new value to replace
+     */
+    setAttributeValueById(attributeId : string, value? : string) {
+        const attributes = this.getOwnedAttributes();
+        const attr = attributes.find((attr) => attr.attributeId === attributeId);
+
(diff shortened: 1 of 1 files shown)


apps/server/src/becca/entities/bnote.ts
```diff
@@ -1758,6 +1758,41 @@ class BNote extends AbstractBeccaEntity<BNote> {
...
+     * Return an attribute by it's attributeId.  Requires the attribute cache to be available.
+     * @param attributeId - the id of the attribute owned by this note
+     * @returns - the BAttribute with the given id or undefined if not found.
+     */
+    getAttributeById(attributeId : string): BAttribute | undefined {
+        this.__ensureAttributeCacheIsAvailable();
+
+        if (!this.__attributeCache) {
+            throw new Error("Attribute cache not available.");
+        }
```
eliandoran, line 1771: Since `ensure` is called, this can never happen. To bypass a null error you might have, append `!` to `__attributeCache` when accessing it.

apps/server/src/becca/entities/bnote.ts
```diff
@@ -1758,6 +1758,41 @@ class BNote extends AbstractBeccaEntity<BNote> {
         return childBranches;
     }
 
+    /**
+     * Return an attribute by it's attributeId.  Requires the attribute cache to be available.
```
eliandoran, line 1762:
```suggestion
     * Return an attribute by its attribute ID.  Requires the attribute cache to be available.
```

apps/server/src/becca/entities/bnote.ts
Geekswordsman: This is following the exact pattern in bnote.getAttributes:

```ts
    getAttributes(type?: string, name?: string): BAttribute[] {
        this.__validateTypeName(type, name);
        this.__ensureAttributeCacheIsAvailable();

        if (!this.__attributeCache) {
            throw new Error("Attribute cache not available.");
        }

        if (type && name) {
            return this.__attributeCache.filter((attr) => attr.name === name && attr.type === type);
        } else if (type) {
            return this.__attributeCache.filter((attr) => attr.type === type);
        } else if (name) {
            return this.__attributeCache.filter((attr) => attr.name === name);
        } else {
            return this.__attributeCache;
        }
    }
```

Which is the base called by other getAttribute methods.  As I am not calling this method directly, holding this validation check to the same standard as the original method seems prudent.

apps/server/src/becca/entities/bnote.ts
Geekswordsman: Referencing standard of documentation elsewhere, it is standard to use the property's name.  Attributes do not have an ID field, they have an attributeId field.

eliandoran (approved):

PR #6602, opened 2025-08-10
feat(etapi): also save note revision via etapi if needed too

perfectra1n:


perfectra1n:
I didn't even know spurious was a word

perfectra1n:
I can't believe that was actually causing the issue

PR #6609, opened 2025-08-11
feat(logs): cleanup physical log files after 90 days by default

perfectra1n:
So that this doesn't happen anymore lol:
image


apps/server/src/services/log.ts
```diff
@@ -27,16 +32,96 @@ function getTodaysMidnight() {
...
 
+async function cleanupOldLogFiles() {
+    try {
+        // Get retention days from environment or options
+        const envRetention = process.env.TRILIUM_LOG_RETENTION_DAYS;
+        let retentionDays = DEFAULT_RETENTION_DAYS;
+        
+        if (envRetention) {
+            const parsed = parseInt(envRetention, 10);
+            if (!isNaN(parsed) && parsed > 0 && parsed <= 3650) {
```
rom1dep: Perhaps we could have `DEFAULT_RETENTION_DAYS = -1` for the user to signify that they want to retain all log files?

rom1dep: FTR, cont: https://github.com/TriliumNext/Trilium/pull/6644

PR #6642, opened 2025-08-13
Fix: activate the nearest path when opening a cloned note

SiriusXT:
```
root
└─ A
   ├─ B
   │  └─ C (clone)
   │      └─ D
   └─ E
       └─ C
           └─ D
```

When opening the **C (clone)** note and clicking its child note **D** (in the `note-list-container`),

* **Actual activated path** (the path selected in the left note tree):

```
root > A > E > C > D
```

* **Expected activated path** (the correct path showing the clone’s location):

```
root > A > B > C (clone) > D
```

### DIFF
--- a/apps/client/src/entities/fnote.ts
+++ b/apps/client/src/entities/fnote.ts
@@ -416,7 +416,7 @@ class FNote {
         return notePaths;
     }
 
-    getSortedNotePathRecords(hoistedNoteId = "root"): NotePathRecord[] {
+    getSortedNotePathRecords(hoistedNoteId = "root", activeNotePath: string | null = null): NotePathRecord[] {
         const isHoistedRoot = hoistedNoteId === "root";
 
         const notePaths: NotePathRecord[] = this.getAllNotePaths().map((path) => ({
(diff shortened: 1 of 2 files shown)


eliandoran:
@SiriusXT , why close it?

SiriusXT:
Because the current fix still has gaps.

For example, given the tree:

```
root
└─ A
   ├─ B
   │  └─ C (clone)
   │      └─ D
└─ C
   └─ D
```

After executing `path.push("root")` ([[link to code](https://github.com/TriliumNext/Trilium/blob/ec74f5f1de4ede27a4097c2b67e1986788a6cfa9/apps/client/src/services/tree.ts#L33-L35)](https://github.com/TriliumNext/Trilium/blob/ec74f5f1de4ede27a4097c2b67e1986788a6cfa9/apps/client/src/services/tree.ts#L33-L35)), since `root/C/D` exists, it will still activate `root/C/D`. Therefore, this fix needs to be improved.

I think it might be better to remove:

* [[tree.ts#L33-L35](https://github.com/TriliumNext/Trilium/blob/ec74f5f1de4ede27a4097c2b67e1986788a6cfa9/apps/client/src/services/tree.ts#L33-L35)](https://github.com/TriliumNext/Trilium/blob/ec74f5f1de4ede27a4097c2b67e1986788a6cfa9/apps/client/src/services/tree.ts#L33-L35)
* [[link.ts#L88-L92](https://github.com/TriliumNext/Trilium/blob/ec74f5f1de4ede27a4097c2b67e1986788a6cfa9/apps/client/src/services/link.ts#L88-L92)](https://github.com/TriliumNext/Trilium/blob/ec74f5f1de4ede27a4097c2b67e1986788a6cfa9/apps/client/src/services/link.ts#L88-L92)
…and instead always complete the full path at the point where navigation is needed.
However, I’m not sure if removing these lines would cause other issues.

---

Currently, I cannot debug this because I want to switch to the latest branch, but it seems that either the repository or my computer has a problem: every time I try to discard this change on the `main` branch, it automatically gets added back — even though I haven’t made any changes to this file.
Even if I delete the folder and re-clone, the problem persists.
preview

eliandoran:
@SiriusXT , I have the same problem with the `.md` file. I'll try to figure out a way to fix it.

I've pushed a fix. Try:

```sh
git fetch
git checkout origin/main
git branch -d main
git checkout main
```

PR #6645, opened 2025-08-14
Add Traditional Chinese translation for README file & fix Docker Hub URL

francistw:


francistw:
Also updated the Simplified Chinese translation while I was at it.

eliandoran (approved):

PR #6672, opened 2025-08-16
feat(quick_search): format multi-line results better

perfectra1n:
Now the results look like this if they have linebreaks :)
image

I made sure to have it still respect the character limit for the context of a search result, so that even if it's multi-line, that it doesn't just immediately have huuuuuuge context previews 😂


perfectra1n:
Also fixed the results to show/work for tags/attributes:
image

Issue #6546, opened 2025-08-03
Store file notes in the filesystem rather than the database


alexdachin:
### Describe feature

Currently Trilium stores the contents of a File note type embedded inside the SQLite database.

While this is extremely nice for simplicity and portability, it can become problematic once you start uploading many reference files (for example some large quality images, large PDFs, videos etc). While in theory SQLite supports huge databases, storing many files in a single database makes the database file harder to backup, increases database lock times and increases the database corruption risk slightly.

I propose storing the files in the filesystem and storing a reference to them and extra metadata in the database. The files can be stored in the trilium-data directory to begin with for simplicity, but it also opens the gate for storing them in other places in the future (like S3 compatible storage systems). This keeps the SQLite database small and performant, enables faster incremental backups (like rsync), reduces the risk of database corruption and still allows Trilium to manage things like note hierarchy with cloning.

Would love to hear your thoughts on weather this approach could be added as a feature or if you feel it might complicate the setup too much.

eliandoran:
Trilium was implemented with a database in order to store the meta information that doesn’t always map well to a file system, such as being able to place a note in multiple locations (this can be partially supported at FS level via symlinks or hardlinks but FS support can be janky or non-portable).

Another important aspect is that notes in Trilium can also have children, whereas folders in a file system don’t.

Rewriting the entire approach to storing the notes would be too problematic at this stage of the application, however we can consider syncing a subset (similar to how shared notes work), which is described here: https://github.com/TriliumNext/Trilium/issues/6351

The problem is that it’s a very complicated problem and the workload to do it is too high at the moment.

alexdachin:
Thanks for looking into this and for all the effort you put into this project! I think what I mentioned was misunderstood, I was not referring to storing every note as a file and I totally understand why the architectural decision is the way it is for Trilium, it provides a lot of strengths.

I was rather referring only to the content of the actual uploaded [File](https://github.com/TriliumNext/Trilium/blob/main/docs/User%20Guide/User%20Guide/Note%20Types/File.md) note types (for example pdfs, images and videos). They are now stored as a blob in the database, but they can simply be stored flatly or partitioned in the file system with filename being either note id or a UUID (and a reference to this file in the database). **Edit:** It is similar to [this issue](https://github.com/TriliumNext/Trilium/issues/6553) which was opened slightly after I commented this, but with a smaller scope (I promise I don't know the author, it was only a coincidence we posted close to each other 😄)

[...]

PR #6677, opened 2025-08-17
feat(db): allow external sqlite blobs

alexdachin:
Currently Trilium stores the contents of a File note type inside the SQLite database. While this is extremely nice for simplicity and portability, it can become problematic once you start uploading many reference files (for example some large quality images, large PDFs, videos etc). While in theory SQLite supports huge databases, storing many files in a single database:
- makes the database file harder to backup
- increases database lock times
- increases the database corruption risk slightly

I propose **storing big blobs in the filesystem** and storing a reference to them in the database. The files are stored in the data directory to begin with for simplicity, but it also opens the gate for storing them in other places in the future (like s3 compatible storage systems). This keeps the SQLite database small and performant, enables faster incremental backups (like rsync), reduces the risk of database corruption and still allows Trilium to manage things like note hierarchy with cloning.

In order to not disturb the workflow of other people, this feature is disabled by default (so even large blobs are still stored internally in the database). To enable it you need to set `TRILIUM_EXTERNAL_BLOB_STORAGE` environment variable.

The threshold of storing notes externally vs internally is 100kb by default but can be changed with `TRILIUM_EXTERNAL_BLOB_THRESHOLD` environment variable. This way small blobs are read from the database more efficiently compared to storing all blobs in the filesystem.

Closes https://github.com/TriliumNext/Trilium/issues/6546

### DIFF
--- a/apps/server/src/services/sync_update.ts
+++ b/apps/server/src/services/sync_update.ts
@@ -4,7 +4,8 @@ import entityChangesService from "./entity_changes.js";
 import eventService from "./events.js";
 import entityConstructor from "../becca/entity_constructor.js";
 import ws from "./ws.js";
-import type { EntityChange, EntityChangeRecord, EntityRow } from "@triliumnext/commons";
+import blobStorageService from "./blob-storage.js";
+import type { EntityChange, EntityChangeRecord, EntityRow, BlobRow } from "@triliumnext/commons";
 
 interface UpdateContext {
     alreadyErased: number;
@@ -125,6 +126,14 @@ function preProcessContent(remoteEC: EntityChange, remoteEntityRow: EntityRow) {
                 // (possibly not a problem anymore with the newer better-sqlite3)
                 remoteEntityRow.content = "";
             }
+        }
+
+        // store external blobs on this instance too
+        const blobRow = remoteEntityRow as BlobRow;
+        if (blobRow.contentLocation && blobRow.contentLocation !== "internal" && remoteEntityRow.content) {
+            const newContentLocation = blobStorageService.saveExternal(blobRow.blobId, remoteEntityRow.content);
+            blobRow.contentLocation = newContentLocation;
+            blobRow.content = null;
         }
     }
 }


apps/server/src/becca/entities/abstract_becca_entity.ts
```diff
@@ -195,6 +198,21 @@ abstract class AbstractBeccaEntity<T extends AbstractBeccaEntity<T>> {
...
         }
 
+        try {
+            const row = sql.getRow<{ contentLocation: string }>("SELECT contentLocation FROM blobs WHERE blobId = ?", [oldBlobId]);
+            if (row?.contentLocation.startsWith('file://')) {
+                const filePath = row.contentLocation.replace('file://', '');
+                blobStorageService.deleteExternal(filePath);
+            }
+        } catch (error) {
+            // contentLocation column might not be present when applying older migrations
```
alexdachin: I tried inlining some of these AbstractBeccaEntity in the 0233__migrate_geo_map_to_collection.ts migration, but some things are event based and the subscribers are using AbstractBeccaEntity as well.

The alternative would be to add a warning that people have to upgrade to the the previous Trilium version first and wait for migrations to complete, but I wanted to avoid that

perfectra1n:
Can you explain more as to what it’s doing, how it works, and what it’d be useful for? Looks cool at first glance! :)

alexdachin:
> Can you explain more as to what it’s doing, how it works, and what it’d be useful for? Looks cool at first glance! :)

Thanks, I updated the PR description.

There is some more context in the issue I linked in the description as well, but I tried to summarize everything here as well.

adoriandoran:
I have noticed an increased risk of BLOB key collisions when the server is running on Windows. The BLOB key is case-sensitive, whereas the underlying file system is case-insensitive.

adoriandoran:
Once the data directory is relocated, the external BLOB paths break, since they are relative to Trilium’s directory. I suggest storing these paths in the database as relative to the “external-blobs” directory instead.

alexdachin:
I see, thank you all for the feedback! I'm away for a few days, but I'll have a look as soon as I get back

alexdachin:
Thank you again for checking this out!

I found out what the issue was with the large json file and fixed it.

Also, good points regarding case insensitive file systems, I switched to a random uuid instead and stored the relative path.

Do you mind checking it one more time? 🙏

alexdachin:
Hi. I see over the last 2 months it acquired some merge conflicts. Since it didn’t get much traction after the latest changes, I wanted to check if anyone would be available to review it if I resolve the conflicts

eliandoran:
@alexdachin , yes.

eliandoran (changes requested):
There are two bugs:

The first is a critical one (data loss):

1. Import the Trace.json file. It will get stored as an external blob.
2. Replace the entire text with a few characters to delete the external blob.
3. Paste the Trace.json back in (or any big text to trigger saving).
4. Restart the server.
5. The server will restart with something like:
   ```
   Consistency issue fixed: Note '5zzWKjTnXz82' content was set to '' since it was null even though it is not deleted
   Consistency issue fixed: Note 'i2EbsDSQkdeb' content was set to '{}' since it was null even though it is not deleted
   ```
6. Look at the Trace.json note, the content will be replaced with `{}`.

The second one involves uploading a video. Regardless of the video size, it will not be saved as an external blob.

In addition, please address the topics from the bot.

capi:
This makes backups a bit more complicated. Currently you can backup by doing a backup of the SQLite database, which you can do transaction save via e.g. `VACCUM INTO`, or other SQLite mechanisms.

With this you actually need to either stop the process before the backup or take a consistent filesystem snapshot.

That said, I really think this is a sensible thing to do. It should just be pointed out in the backup documentation.

apps/server/src/services/blob-storage.ts
```diff
@@ -0,0 +1,102 @@
...
+        const relativePath = path.join(partition, `${uuid}.blob`);
+        const absolutePath = path.join(this.externalBlobPath, relativePath);
+
+        try {
+            const dir = path.dirname(absolutePath);
+            if (!fs.existsSync(dir)) {
+                fs.mkdirSync(dir, { recursive: true, mode: 0o700 });
+            }
+
+            fs.writeFileSync(absolutePath, content, { mode: 0o600 });
```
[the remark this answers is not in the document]
alexdachin: Good point that async would be ideal. Currently the whole app is using better-sqlite3, which is synchronous. Before this change, blobs were stored in sqlite and read/written with blocking calls, so this is maintaining that same pattern just with filesystem storage instead.

Making this async would mean refactoring the call chain (and potentially switching to an async sqlite library), but I figured it shouldn't be in scope for this feature, what do you think?

apps/server/src/becca/becca-interface.ts
```diff
@@ -171,7 +171,7 @@ export default class Becca {
         opts.includeContentLength = !!opts.includeContentLength;
```
alexdachin, line 171: Next step would probably be removing this parameter since getting the content length is a very cheap operation now, but I didn't want to increase the size of this PR and make it harder to review.

This change can be in a follow-up PR

alexdachin:
Thanks a lot for the review @eliandoran and @capi!

That critical bug was a really good catch, sorry for missing it initially, I wasn't familiar with the consistency checks Trilium is doing. I couldn't replicate your issue regarding video uploads though, do you mind checking one more time? I tried different video formats and they all follow the threshold for me.

Really good point regarding backup, I updated the documentation to reflect that.

apps/server/src/services/blob-storage.ts
```diff
@@ -0,0 +1,141 @@
...
+            log.error(`Failed to retrieve blob from external storage ${filePath}: ${error}`);
+            throw error;
+        }
+    }
+
+    /**
+     * Get the content of a blob row
+     */
+    getContent(row: Pick<Blob, "content" | "contentLocation">): string | Buffer | null {
+        return row.contentLocation === "internal" ? row.content : this.getExternal(row.contentLocation.replace("file://", ""));
```
dedmen, line 93: I think instead of just cutting off the file://, there should already be a check here that the location starts with file://, and throw an exception if it doesn't?
As preparation for later when other protocols might be supported, for which there would be another condition added right here, to check the protocol.

Something like

```
if (contentLocation.startsWith("file://") return this.getExternal(...);
else if (contentLocation.startsWith("s3://") return this.getS3File(...);
else throw ...;
```
Later on we could have an array of file providers here, and would loop over them asking each if they can provide the file or not.

apps/server/src/services/erase.ts
```diff
@@ -92,26 +93,34 @@ function eraseRevisions(revisionIdsToErase: string[]) {
...
+        AND revisions.revisionId IS NULL`, []);
 
-    if (unusedBlobIds.length === 0) {
+    if (unusedBlobRows.length === 0) {
         return;
     }
 
-    sql.executeMany(/*sql*/`DELETE FROM blobs WHERE blobId IN (???)`, unusedBlobIds);
+    // Clean up external blob files before deleting from database
+    unusedBlobRows.forEach(row => blobStorageService.deleteExternal(row));
```
dedmen, line 115: Just a note.
If just one of the many files fails to delete (I know it's unlikely currently, might become less unlikely if other storage protocols are added later), all files will be left inside the database.
Should not be an issue as they are not referenced anywhere anyway and just lay around unused.

But might be nice to, if _some_ fail to delete, at least remove the ones from database that successfully deleted.

Also if you have one file that always fails, that means no blobs DB entries are ever deleted, even if all others would delete fine.

One case where a file might fail to delete, is if it is currently open in a external application (which shouldn't happen tbh).
Which brings the other idea, currently we export files into a temp file to open them, we don't need to do that if we already have the standalone file in the external-blobs folder, but that is only something to potentially consider later, not now. (Note it would have security implications because the file would need the correct extension to be openable)

alexdachin:
Thanks for the review @dedmen and to everyone that took the time to review this earlier, I really appreciate the feedback and the time you took to look into this.

At this point I think I'm going to step away from this PR. I've been working on it since August, and while I've tried to address feedback right after receiving it, the long gaps between review cycles have made it hard to keep the branch up to date, and it's become increasingly hard for me to stay motivated to continue iterating on it.

If someone else would like to pick this up or continue from the current state, please feel free to do so. Thanks again for the discussion and insights 🙏

PR #6681, opened 2025-08-17
fix(hotkeys): interpret shortcut in the user's locale

### LINKED ISSUE #6547: Keyboard shortcut catchss QWERTY keys instead of owner's
### Description

From 0.97.2, I noticed the Z of my AZERTY keyboard would trigger window closing upon `Ctrl+Z` and W would trigger Zen mode upon `Ctrl+Z`.
So I guess thee shortcuts are stuck in QWERTY mode...

### TriliumNext Version

0.97.2

macOS

### What is your setup?

Local + server sync

### Operating System Version

MacOS latest + Debian 12 latest

### Error logs

_No response_

rom1dep:
fixes #6547

### DIFF
--- a/apps/client/src/services/shortcuts.ts
+++ b/apps/client/src/services/shortcuts.ts
@@ -14,6 +14,32 @@ interface ShortcutBinding {
 // Store all active shortcut bindings for management
 const activeBindings: Map<string, ShortcutBinding[]> = new Map();
 
+// Handle special key mappings and aliases
+const keyMap: { [key: string]: string[] } = {
+    'return': ['Enter'],
+    'enter': ['Enter'],  // alias for return
+    'del': ['Delete'],
+    'delete': ['Delete'], // alias for del
+    'esc': ['Escape'],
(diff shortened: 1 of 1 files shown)


rom1dep:
I also took the opportunity to move the definition and build-up of `keyMap` out of `function keyMatches`, since I suspect it might help with performance (needlessly re-built on every key stroke × defined shortcut)

rom1dep:
(for reference, this broke following https://github.com/TriliumNext/Trilium/pull/6507 )

perfectra1n:
Some additional context provided via Matrix:
> Yeah it works for me, typically on the case where I have a shortcut mapped to a key that is different on my layout (being A) vs the English one (where it's a Q),
As stated on MDN, using e.code is incorrect there because it's not locale-dependant.
Also, it's not changing the "return" from upper to lower, just the basis for comparison, which I aligned with the one 3 lines below (I'm not sure it's required, but that casing wrangling wasn't introduced by me in the first place)

PR #6701, opened 2025-08-19
feat: update "jump to" dialog to use better search results, and separate renderer for "quick search" and "jump to"

perfectra1n:
image

image

image


perfectra1n:
Probably just going to close this one in the AM and start from scratch because it's just a mess...

perfectra1n:
Continued in #6752

PR #6736, opened 2025-08-21
feat(docs): add some more documentation around features that previously documented

perfectra1n:


eliandoran:
I've looked through the documentation, although it's not bad it adds more complexity and verbosity to the documentation than I believe is necessary. There are also subtle things that are hard to validate, given that there are 22k additions.

Will revisit the PR if necessary.

PR #6737, opened 2025-08-21
feat(oidc): perhaps this will provide better logging and output for t…

perfectra1n:
…he user


perfectra1n:
Example output from this commit:
https://github.com/TriliumNext/Trilium/issues/6444#issuecomment-3215410223

eliandoran:
Closing for now, perhaps we can revisit it if we still have OIDC issues.

Issue #6698, opened 2025-08-19
Reduce or disable search animation


werererer:
### Describe feature

Using search all day, the opening animation feels slow. Coming from neovim, every ms matters. This becomes an issue when spamming search out of habit, since notes are used frequently.

It would be great to make the animations faster by default or add an opt-in/out to skip animations.

### Additional Information

Specifically concerning feature 'Jump To...'

CobriMediaJulien:
I like that suggestion! Ty.

rom1dep:
The quick edit/popup animation can probably use the same treatment

eliandoran:
Done in https://github.com/TriliumNext/Trilium/pull/6747

PR #6744, opened 2025-08-22
Settings/Appearance: add support to disable transitions and animations 

adoriandoran:
Adds an option to completely disable UI motion (not just reduce it).


adoriandoran:
Closes #6698.

Issue #6671, opened 2025-08-16
The incorrect handling of double slashes led to the loss of the entire note!!


cerjuzi:
### Description

When I was trying to reproduce the #6670 error, I encountered a much larger bug. My entire document was lost and there was no history to recover!!
I found that after inputting double slashes, I couldn't use the backspace key to delete the double slashes, and then the entire document suddenly disappeared. I will record a video to reproduce the operation because this is the second time I have encountered this bug

### TriliumNext Version

0.92

### What operating system are you using?

Other Linux

### What is your setup?

Server access only

### Operating System Version

centos 9

cerjuzi:
I've confirmed that the 100% reproduction method is the same as the trigger method of #6670. This is caused by input through the input method. Even when typing English letters with the input method, as shown in the video, I kept pressing the backspace key at the end of the video, but I couldn't delete the characters. And when I moved the input cursor to other text and pressed the backspace key at this time, the entire note was cleared.

https://github.com/user-attachments/assets/1072b242-a16b-4182-9a7f-bd2bcff7fe17

cerjuzi:
This should be a bug added by adding the slash shortcut command of CKEdit in some update. I hope there could be an option to disable the slash shortcut command to temporarily solve this problem and prevent the notes from being cleared

ecodiv:
@cerjuzi I was looking for the, for me, new behavior that entering a double slash can be used to enter the current date/time, and came across this bug report.

I tried but can't reproduce your problem. I can enter a double slash in a code block followed by text and subsequently use the backspace to delete the text and double slash. I am not sure if it matters, but I am on Linux (Ubuntu 22.04).

cerjuzi:
@ecodiv It is indeed very useful, but as shown in the video, the problem was reproduced in a brand new note. I think it is necessary to add an enable switch for the shortcut command

herrkami:
> The slash command is enabled by default and managed in the code, specifically in the CKEditor configuration file (ckeditor.ts). 

@dosu, 
1) the file is called [config.ts](https://github.com/TriliumNext/Trilium/tree/main/apps/client/src/widgets/type_widgets/ckeditor) and 
2) nothing is hard about fixing this imo.  

If I'm not mistaken, we could simply introduce a boolean parameter in the CKEditor options page and handle it like the Emoji completion (`removePlugins`, `getDisabledPlugins()`, etc.). It would require some testing but it appears fairly low effort to me. 

[...]

Issue #6670, opened 2025-08-16
The content entered after the double slashes will disappear


cerjuzi:
### Description

I'm not sure if there is a switch, but when I use double slashes as comments in the code block, the comments are not actually saved. Switching to another place to input or refreshing the page will cause the comments to disappear

### TriliumNext Version

0.92

### What operating system are you using?

Other Linux

### What is your setup?

Server access only

### Operating System Version

centos 9

cerjuzi:
Reproduction steps
1. Entering double slashes will bring up the option "Insert Date/Time", but I ignored it and entered a space

<img width="421" height="125" alt="Image" src="https://github.com/user-attachments/assets/c9adc144-d477-47f7-a32c-4b0c7fce8973" />

2. It can be seen that the color of the subsequent input text has not been rendered to the color of the comments. At this point, if you input or refresh the page elsewhere, this text will disappear

<img width="287" height="103" alt="Image" src="https://github.com/user-attachments/assets/983ef3d2-e4cb-412d-a7ea-b48c0de1935c" />

PR #6764, opened 2025-08-24
feat(options): allow for user selection of ckeditor plugins

perfectra1n:


apps/server/src/routes/api/ckeditor_plugins.ts
eliandoran, line 1: Interesting approach, but I believe there's no need to have the server involved for a client-side-only affair (CKEditor).

This seems like it could be a service inside the client.

apps/client/src/widgets/type_widgets/options/ckeditor_plugins.tsx
```diff
@@ -0,0 +1,397 @@
...
+    tables: t("ckeditor_plugins.category_tables"),
+    advanced: t("ckeditor_plugins.category_advanced"),
+    trilium: t("ckeditor_plugins.category_trilium"),
+    external: t("ckeditor_plugins.category_external")
+};
+
+export default function CKEditorPluginSettings() {
+    const [pluginRegistry, setPluginRegistry] = useState<PluginRegistry | null>(null);
+    const [userConfig, setUserConfig] = useState<PluginConfiguration[]>([]);
+    const [stats, setStats] = useState<PluginStats | null>(null);
```
eliandoran, line 44: You can make the plugins list dynamically update by using the Trilium-specific hooks `useTriliumOption` or `useTriliumOptionJson` or `useTriliumOptions`.

eliandoran:
Since both #6670, #6671 and #7057 are closed, I believe we can close this PR.
Although it's interesting work, it needs further polishing and I'm not sure we need to be able to disable _all_ plugins, for example like tables.

PR #6769, opened 2025-08-24
feat(quick_search): also allow for the equals operator in note title's quick search

perfectra1n:


apps/server/test_search_integration.js
eliandoran, line 1: Do we still need this file? If we do, then we probably need to move it somewhere else.

perfectra1n: I think it's a good idea to keep, just in case we break that functionality again?

PR #6810, opened 2025-08-27
feat(quick_search): just fuzzy match note titles for larger notes, while still matching on exact strings

### LINKED ISSUE #6808: Search function slow since 0.98.0
### Description

The new search UX is really nice!

I have a rather large Trilium database. The search function used to be a little slow before, but still within a reasonable range.

Since version 0.98.0, however, it has become very slow - about five times slower compared to the previous version.

It would be great to see some optimization of the search performance.

Image

### TriliumNext Version

0.98.1

Ubuntu

### What is your setup?

Server access only

### Operating System Version

Ubuntu 24.04.2 LTS

### Error logs

```
High word count for note CgHdgKy91P2I: 187927 words - phrase matching may take time but will attempt full search
High word count for note CgHdgKy91P2I: 187927 words - phrase matching may take time but will attempt full search
Slow 200 GET /api/quick-search/repkav with 8959 bytes took 8877ms
```

perfectra1n:
Should close #6808

### DIFF
--- a/apps/server/src/services/search/expressions/note_content_fulltext.ts
+++ b/apps/server/src/services/search/expressions/note_content_fulltext.ts
@@ -120,17 +120,54 @@ class NoteContentFulltextExp extends Expression {
         }
         content = processedContent;
         
-        // Check if this is a large note that needs optimized search strategy
-        const wordCount = content.split(/\s+/).length;
-        const isLargeNote = wordCount > FUZZY_SEARCH_CONFIG.LARGE_NOTE_THRESHOLD;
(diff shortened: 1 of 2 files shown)


SiriusXT:
I was wondering if we could make quick search default to searching by title while typing, similar to how the new tab page searches by name — and then only search the full content when pressing Enter.

perfectra1n:
That would irritate me so much as a user 😂

eliandoran:
Closing in favor of #6839.

Issue #6798, opened 2025-08-27
#cssClass isn't applied to note content in mobile layout


Papierkorb2292:
### Description

I have the following`#appCss`:
```css
.myClass {
    background-color: red !important;
}
```
and I have a text note with `#cssClass=myClass`. However, the custom styling is only applied to the content of the note when I use the desktop layout, not when I use the mobile layout. This doesn't affect the tree view btw, there the styling is applied either way.
If it helps, I already found out that the `#cssClass` is applied by the `NoteWrapperWidget` class, which is only used by the desktop layout.

### TriliumNext Version

0.98.1

### What operating system are you using?

Other Linux

### What is your setup?

Server access only

### Operating System Version

Linux Mint 22.1 Cinnamon

eliandoran:
Done in https://github.com/TriliumNext/Trilium/pull/6847.

PR #6847, opened 2025-09-01
Apply note wrapper on mobile

Papierkorb2292:
I apologize if I made any mistakes due to insufficient knowledge of the code base, this is just my attempt at fixing #6798, which I did by adding the wrapper widget to the mobile layout and then making sure it still receives the note context to apply the css class.


Papierkorb2292:
Actually yeah, this doesn't work with multiple tabs on mobile, hold on...

PR #6851, opened 2025-09-01
fix(shortcuts): try to fix ime composition checks

### LINKED ISSUE #6846: 使用中文输入法在输入框输入时，如果直接按回车，会导致输入异常。
使用中文输入法在输入框输入时，如果直接按回车，会导致输入异常。
When using the Chinese input method to enter the input box, if you press enter directly, it will cause an abnormal input.

这个现象是从0.97.2开始的，会严重影响到使用体验。我认为这是一个bug。
This phenomenon started in 0.97.2 and will seriously affect the user experience. I think it's a bug.

下面是详细描述：
Here's a detailed description:

### 新建笔记 New note

Image

### 搜索 Search

Image

note：

微软拼音输入法(Microsoft Pinyin Input Method)

### TriliumNext Version

0.97.2、0.98.0、0.98.1

Windows

### What is your setup?

Server access only

### Operating System Version

Windows 11：26100.4652  Edge：139.0.3405.125 (正式版本) (64 位)

perfectra1n:
Should resolve #6846


rom1dep:
Excuse my ignorance, but, aren't we fighting the platform here?

If the default supports IME's composition quirks, and we broke that somehow, would it be possible to just undo the things we did to break them (instead of re-implementing new quirks on top)?

perfectra1n:
This is continuing to fix the things that #6507 broke. I don't believe we want to go back and undo everything that we've done to fix hotkeys since that change.

rom1dep:
Sure, without going as far as proposing to undo "everything", it might be good to see if there aren't "some things" from #6507 that could be "undone".
From a [quick look into the unvendored repo](https://github.com/search?q=repo%3Ajeresig%2Fjquery.hotkeys+%28composing+OR+229%29&type=code), I can't find anything comparable to checking for `event.isComposing` or `e.keyCode === 229`. Wasn't it required back then? Can we get to the bottom of why it would be now?
I'm a bit wary of adding more checks and logic to the event loop, and make the overall typing and navigation in trilium more laggy.

perfectra1n:
`jQuery.hotkeys` uses `event.which`/`event.keyCode` to determine which key was pressed:

**During IME composition (when pressing Enter to confirm Chinese input):**
- `event.keyCode = 229` (IME processing code)
- `event.key` = Could be "Enter", "Process", or "Unidentified" (browser-dependent)
- jQuery.hotkeys checks: `specialKeys[229] = undefined`
- Falls back to: `String.fromCharCode(229) = "å"`
- Checks if "å" matches "return" (NO MATCH)
- Result: Handler not triggered


**Normal Enter press (not during IME):**
- `event.keyCode = 13`
- `jQuery.hotkeys` checks: `specialKeys[13] = "return"`
- Checks if "return" matches "return" (MATCH)
- Result: Handler triggered


The new implementation uses `event.key` to determine which key was pressed:
```js
const mappedKeys = keyMap['return']; // ['Enter']
if (mappedKeys.includes(e.key)) { // Checks if e.key === 'Enter'
    // Trigger handler
}
```

The problem occurs in certain scenarios:
- Some browsers/IMEs send `e.key = "Enter"` even when `keyCode = 229`
- Or there might be two events: One with keyCode 229, then another with keyCode 13
- The new code checks `e.key === "Enter"` which could be true during IME

So the main diff:
- `jQuery.hotkeys`: Uses keyCode as the primary determiner -> 229 never matches 13
- New implementation (this change): Uses `e.key` as the primary determiner -> "Enter" matches "Enter" regardless of IME state

But yeah, it looks like `jQuery.hotkeys` accidentally provided IME protection through its `keyCode`-based approach. When `keyCode` 229 (IME) is converted to a character ("å"), it naturally doesn't match any registered shortcuts. The new implementation's `e.key`-based approach is more modern but needed explicit IME handling to work correctly with Input Method Editors.

rom1dep:
@perfectra1n kudos for the detailed explanation, it's much clearer, now. Thanks :-)

PR #6867, opened 2025-09-02
feat(editor): try to have ckeditor not crash when handling stranger tags

perfectra1n:
this definitely isn't being merged, but keeping it open for Elian


eliandoran:
Changing from code to text note is not supported and the user is already warned that corruption might occur, so I would prefer not to have the maintenance burden of handling this edge case.

Thanks anyway. 😄

Issue #6715, opened 2025-08-20
Custom History | 自定义历史记录


youli42:
### Describe feature

扩展一下历史记录的功能：允许手动创建历史记录
Expand the function of the history record: allow manual creation of history records

具体来说：在历史记录页面提供一个创建此刻的文档的历史记录的按钮，并允许为其设置自定义备注，类似git
Specifically: Provide a button on the history page to create a history record of the document at the current moment, and allow setting a custom note for it, similar to git.

### Additional Information

更进一步：如果可以，希望可以提供一个简单的差异，对比工具用来对比历史记录
Going a step further: if possible, it is hoped that a simple difference comparison tool can be provided to compare historical records.

perfectra1n:
Hmm, a “diff” between versions would be nice

youli42:
对于我来说，可以手动创建历史记录和为每个历史记录添加“自定义备注”，似乎是更有用的功能，因为文本差异对比可以通过大量第三方工具进行，而 trilium 本身提供了一个方便的界面获取复制历史版本。

For me, the ability to manually create history records and add "custom notes" to each history record seems to be a more useful feature. This is because text difference comparison can be done through a large number of third-party tools, while Trilium itself provides a convenient interface to access and copy historical versions.

eliandoran:
Interestingly, @zadam seemed opposed on the idea of named revisions in https://github.com/TriliumNext/Trilium/issues/561#issuecomment-519797864

The diff part is already handled in #6887.

youli42:
或许这意味着，大多数情况下，无需手动控制，只有在极其特殊的情况下，提供一个额外的选项，允许用户自定义。

Perhaps this means that in most cases, manual control is not needed, and only in extremely special circumstances, an additional option is provided to allow users to customize.

Issue #334, opened 2019-01-13
Implement note revision diff for text and code notes


zadam:
depends on TriliumNext/Notes#333

meichthys:
I just came here to request this but realized it was already an issue. 
My use case would primarily need a diff between the current note and the selected revision.
Bonus: If two revisions could also be diffed.

deajan:
I'd love to see that feature too.
I'm using trilium to store some config files, and seeing differences in diff format would make things really easy for me.

meichthys:
[Trilium has entered maintenance mode](https://github.com/zadam/trilium/issues/4620#issuecomment-2094495698). Future enhancements will be addressed in TrilumNext: https://github.com/TriliumNext/trilium/issues/4974

eliandoran:
Done in #6887.

PR #6887, opened 2025-09-04
feat: show source diff between note and revision

SiriusXT:
Related to [https://github.com/TriliumNext/Trilium/issues/6715](https://github.com/TriliumNext/Trilium/issues/6715) and [https://github.com/TriliumNext/Trilium/issues/334](https://github.com/TriliumNext/Trilium/issues/334)

Compare the source code differences between notes and revision items
Preview:
preview

---

The comparison is performed at the source code level rather than the rendered content level due to the lack of mature libraries for directly comparing rendered output. Using diff to compare rendered content would cause the HTML tags within notes to be treated as plain text and incorrectly displayed as part of the differences.

For example, when comparing content differences, what should have been marked as a revision ends up showing the HTML tags themselves as differences:

Revision content:
图片

Content difference:
图片


eliandoran:
Diffing content like this would work pretty well for code, but for HTML I still think we should find some solution to do proper visual diffing. 

Maybe something like https://github.com/fiduswriter/diffDOM to compare the DOM and only then apply diffing for each "block"?

SiriusXT:
> Diffing content like this would work pretty well for code, but for HTML I still think we should find some solution to do proper visual diffing.
> 
> Maybe something like https://github.com/fiduswriter/diffDOM to compare the DOM and only then apply diffing for each "block"?

@eliandoran I spent all afternoon trying out DiffDOM, but didn’t make any real progress—it’s just too tricky for me. I don’t know how to get it done.

eliandoran:
@SiriusXT , no worries, let's keep it as it is as a first version, maybe have a look at what CodeQL complains about.

apps/client/src/widgets/dialogs/revisions.tsx
```diff
@@ -143,6 +156,17 @@ function RevisionPreview({ revisionItem, setShown, onRevisionDeleted }: {
                 {(revisionItem && <div className="revision-title-buttons">
                     {(!revisionItem.isProtected || protected_session_holder.isProtectedSessionAvailable()) &&
                         <>
+                            {["text", "code", "mermaid"].includes(revisionItem.type) && (
+                                <Button
```
eliandoran: Instead of a Button, try a `FormToggle` to make it look like one of the switches from the basic properties.

apps/client/src/widgets/dialogs/revisions.tsx
```diff
@@ -267,6 +294,34 @@ function RevisionContentText({ content }: { content: string | Buffer<ArrayBuffer
...
+                return utils.escapeHtml(part.value);
+            }
+        }).join("");
+    } else {
+        return <>{t("revisions.diff_not_available")}</>
+    }
+    
+    return (
+        <div className="ck-content" style={{ whiteSpace: "pre-wrap" }}
+            dangerouslySetInnerHTML={{ __html: diffHtml }}></div>
```
eliandoran: Use the `RawHtml` component.

apps/client/src/widgets/dialogs/revisions.tsx
```diff
@@ -267,6 +294,34 @@ function RevisionContentText({ content }: { content: string | Buffer<ArrayBuffer
...
+        if (itemType === "text") {
+            noteContent = utils.formatHtml(noteContent);
+            itemContent = utils.formatHtml(itemContent);
+        }
+        const diff = diffWords(noteContent, itemContent);
+        diffHtml = diff.map(part => {
+            if (part.added) {
+                return `<span style="background:rgba(100, 200, 100, 0.5)">${utils.escapeHtml(part.value)}</span>`;
+            } else if (part.removed) {
+                return `<span style="background:rgba(255, 100, 100, 0.5);text-decoration:line-through;">${utils.escapeHtml(part.value)}</span>`;
```
eliandoran: Since styling the highlight is probably going to be quite requested, try using classes and putting those styles in `style.css` instead.

apps/client/src/services/utils.ts
```diff
@@ -297,6 +297,54 @@ function isHtmlEmpty(html: string) {
     );
 }
 
+function formatHtml(html: string) {
```
eliandoran, line 300: There is already a formatting function in `ReadOnlyCodeTypeWidget`. Is it similar, can we use that one instead?
Either way, it's OK to have it in `utils`, but I'd like to deduplicate the one in `read_only_code.ts`.

eliandoran (changes requested):

eliandoran (approved):
Pretty nice!

Issue #6799, opened 2025-08-27
Better Markdown Export


tuuuni0scouts:
### Describe feature

It woud be Relaly nice when in the Markdown Export the Name of the Note woud only stand in the Filename and not also as a #Header in the first line of the export.

GrantZhu1001:
I committed a PR to optimize this issue, hoping it can be of help to you #6918

eliandoran:
@GrantZhu1001 , @tuuuni0scouts , can you please describe why this is an issue for you?

The presence of the # (h1 title) in the Markdown document is important since it reflects the title of the note.

This is generally how most applications treat the title, as far as I know.

The title is also used to determine the title of the note when importing it as a single file. This is especially important if the title of the note cannot be well represented in the file system (for example if contains question marks or other symbols).

tuuuni0scouts:
No problem, i understand i yust taught it would be nice, then it would work whit obsidian without having 2 titles.

eliandoran:
Not planned, in accordance in https://github.com/TriliumNext/Trilium/pull/6918.
The Markdown export can always be processed manually or automatically via a script, to strip the heading 1.

m3thm4th:
@eliandoran 
> This is generally how most applications treat the title, as far as I know.

It is not. What application does that? The title of the .md file is the title, headings are within the file.

The title of the .md file should be the title of the note. The headings should start with H1 as it is standardized with markdown. Headings are headings, not the title of a document.

Importing a markdown document in Trilium messes up everything.

There is no point reinventing the wheel.

eliandoran:
@m3thm4th ,

> It is not. What application does that?

MkDocs is an application like that: https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data

You have to think of Trilium not as a WYSIWYG editor, but as a personal wiki system. That means it gets to have its own conventions that you might or might not agree with.

> Headings are headings, not the title of a document.

If that would be the case, then how would the title be able to contain:

* < > : " / \ | ? * on Windows
* CON, PRN, AUX, NUL
* COM1 … COM9
* LPT1 … LPT9
* End with a dot.

Trilium does not restrict on the title of the note, compared to the file name which depends on the naming conventions of the platforms it's running on.

> There is no point reinventing the wheel.

Trilium is not reinventing the wheel where, it's simply using a different wheel than the one you prefer.

> Importing a markdown document in Trilium messes up everything.

[...]

PR #6918, opened 2025-09-07
opt: better markdown export (#6799)

GrantZhu1001:
when exporting notes in markdown format, remove the first line comment in markdown


eliandoran:
Sorry, but I think we need first to discuss why removing the title in a Markdown file is needed.

We can reopen the PR if needed, until them let's continue the discussion over in the related issue.

The biggest problem I see is that this is a change in behaviour that many people will be surprised to see. We can't simply remove it, if it's really necessary then I suppose we need have a way to toggle it (although we still need a valid usecase explanation for it).

Issue #5824, opened 2025-06-21
issue with .deb install > 0.93


Pada0:
### Description

Hi there,

0.92.x and 0.93 are ok
But with 0.94.1 and 0.95 I have this error message

A JavaScript error occurred in the main process
Uncaught Exception:
Error: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /usr/lib/trilium/resources/app.asar.unpacked/node_modules/better-sqlite3/build/Release/better_sqlite3.node)
    at process.func [as dlopen] (node:electron/js2c/node_init:2:2569)
    at Module._extensions..node (node:internal/modules/cjs/loader:1930:18)
    at Object.func [as .node] (node:electron/js2c/node_init:2:2796)
    at Module.load (node:internal/modules/cjs/loader:1472:32)
    at Module._load (node:internal/modules/cjs/loader:1289:12)
    at c._load (node:electron/js2c/node_init:2:17950)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:242:24)
    at Module.require (node:internal/modules/cjs/loader:1494:12)
    at require (node:internal/modules/helpers:135:16)
[27335:0621/152627.255461:ERROR:components/viz/service/main/viz_main_impl.cc:183] Exiting GPU process due to errors during initialization

### TriliumNext Version

0.94.1

### What operating system are you using?

Other Linux

### What is your setup?

Local (no sync)

### Operating System Version

debian 12

### Error logs

A JavaScript error occurred in the main process
Uncaught Exception:
Error: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /usr/lib/trilium/resources/app.asar.unpacked/node_modules/better-sqlite3/build/Release/better_sqlite3.node)
    at process.func [as dlopen] (node:electron/js2c/node_init:2:2569)
    at Module._extensions..node (node:internal/modules/cjs/loader:1930:18)
    at Object.func [as .node] (node:electron/js2c/node_init:2:2796)
    at Module.load (node:internal/modules/cjs/loader:1472:32)
    at Module._load (node:internal/modules/cjs/loader:1289:12)
    at c._load (node:electron/js2c/node_init:2:17950)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:242:24)
[...]

PR #7002, opened 2025-09-16
Fix(build): Fix the issue that on ARM64 Linux failed to launch due to missing better_sqlite3.node

linull24:
## Issue
I've been experiencing the same issue over and over again as reported in #5824 on my ArchlinuxARM system. The error message is:
‘/app/lib/com.triliumnext.notes/resources/app.asar.unpacked/node_modules/better-sqlite3/build/Release/better_sqlite3.node: cannot open shared object file: No such file or directory'
## Cause
After digging into the build process, I found the real issue is not related to GLIBC version as the bot suggested. The bot's response has been misleading users for months.

The actual problem is in the build pipeline for ARM64 Linux packages:
1.  ARM64 Linux builds are cross-compiled on x64 Ubuntu runners (ubuntu-22.04)
2. The better-sqlite3 native module gets built for x64 architecture instead of ARM64
3. When the package runs on actual ARM64 systems, the x64 .node file can't be loaded
## Solution
I used AI tools to help analyze and fix this issue. Full disclosure: I'm taking full responsibility for this fix, but I want to be transparent that I relied heavily on AI assistance for the implementation. The fix has been tested and works correctly.

The changes:
1. Use native ubuntu-24.04-arm runner for ARM64 Linux builds
2. Add TARGET_ARCH environment variable to the rebuild process
3. Pass architecture parameter to @electron/rebuild
## Files Changed
1. github/workflows/release.yml - Use native ARM64 runner for ARM64 Linux
2. .github/actions/build-electron/action.yml - Add TARGET_ARCH environment variable
3. scripts/electron-rebuild.mts - Add architecture parameter to rebuild config

## Some BS Talk
I'm honestly frustrated that #5824 was **closed** when the bot gave a completely wrong diagnosis. The GLIBC explanation sent users down the wrong path - using flatpak workarounds or building from source when the real issue was just incorrect build architecture.

This could have been fixed months ago if the actual build process was examined instead of accepting the bot's surface-level analysis.


eliandoran:
> After digging into the build process, I found the real issue is not related to GLIBC version as the bot suggested. The bot's response has been misleading users for months.

This is not really accurate. There really existed an issue with the GLIBC, which was solved by downgrading the image to 22.04 (see https://github.com/TriliumNext/Trilium/commit/a334e82eb0c182283be55f8ca06f7b6bee9eb825). A lot of people complained about this specific issue, so probably the bot thought the issue was there as well.

> I'm honestly frustrated that https://github.com/TriliumNext/Trilium/issues/5824 was closed when the bot gave a completely wrong diagnosis.

The issue was closed by the original poster of the issue. 

> This could have been fixed months ago if the actual build process was examined instead of accepting the bot's surface-level analysis.

This has nothing to do with the bot. The only bottleneck here was the lack of equipment for me to actually test it, since I don't have any ARM equipment on hand. If you have ARM equipment to donate, let me know. ;)

As a side note, ARM support is really best-effort only and not a "main" platform for us since the amount of users is not really that high:

```mermaid
xychart-beta horizontal
    title "Total downloads in GitHub per release (48198)"
    y-axis "Num downloads"
    x-axis [ desktop-windows-x64-exe, server-linux-x64-tar.xz, desktop-windows-x64-zip, desktop-linux-x64-deb, desktop-mac-arm64-dmg, desktop-linux-x64-tar.xz, desktop-linux-x64-zip, desktop-windows-arm64-exe, desktop-linux-x64-rpm, desktop-mac-x64-dmg, desktop-mac-arm64-zip, desktop-linux-arm64-deb, desktop-linux-x64-flatpak, desktop-mac-x64-zip, desktop-windows-arm64-zip, desktop-linux-arm64-zip, server-linux-arm64-tar.xz, desktop-linux-arm64-flatpak, desktop-linux-arm64-rpm, desktop-linux-arm64-tar.xz ]
    bar [ 10919, 8371, 8240, 4417, 2625, 2145, 2106, 1936, 1794, 1239, 880, 818, 622, 518, 428, 341, 325, 178, 173, 123 ]
```

PR #7004, opened 2025-09-16
chore(dx/nix): fix flake fully

FliegendeWurst:
I am not exactly happy with this solution, but it works. Especially the pnpm patch is rather annoying.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,7 +23,15 @@
         pkgs = import nixpkgs { inherit system; };
         electron = pkgs."electron_${lib.versions.major packageJsonDesktop.devDependencies.electron}";
         nodejs = pkgs.nodejs_22;
-        pnpm = pkgs.pnpm_10;
+        # pnpm creates an overly long PATH env variable for child processes.
+        # This patch deduplicates entries in PATH, which results in an equivalent but shorter entry.
+        # https://github.com/pnpm/pnpm/issues/6106
+        # https://github.com/pnpm/pnpm/issues/8552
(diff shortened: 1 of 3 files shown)


patches/pnpm-PATH-reduction.patch
eliandoran, line 1: I wonder, should we report this upstream as being an issue?

flake.nix
```diff
@@ -181,7 +195,13 @@
 
         desktop = makeApp {
           app = "desktop";
-          preBuildCommands = "export npm_config_nodedir=${electron.headers}; pnpm postinstall";
+          # pnpm throws an error at the end of `pnpm postinstall`, but it doesn't seem to matter:
+          # ENOENT: no such file or directory, lstat
+          # '/build/source/apps/desktop/node_modules/better-sqlite3/build/node_gyp_bins'
```
eliandoran, line 200: I did notice something interesting about it, this one is triggered by the second rebuild, done for `edit-docs` and not for the `desktop`. This is probably also why it doesn't break anything.

I'm not sure why `electron-rebuild` is touching the `desktop` again, even though it's rebuilding `edit-docs` with its own dependency.

patches/pnpm-PATH-reduction.patch
FliegendeWurst: The relevant repo doesn't have issues enabled https://github.com/pnpm/npm-lifecycle

But I could post it as a PR.

eliandoran: @FliegendeWurst , that would be nice.

PR #7053, opened 2025-09-22
Update typo in ru translation

greenfork:


eliandoran:
@greenfork , thanks for the contribution.
For translations, it's best to do the changes directly in https://hosted.weblate.org/projects/trilium/-/ru/.

PR #7056, opened 2025-09-22
Fix tree state preservation for table view

bangbangbanggit:
**Fixes:**
- Table view tree no longer expands during add row/delete row/edit row operations
- Expansion/collapse state persists across app restarts and navigation
- Context menu 'insert row above/below' options remain available
- No interference with existing table functionality

Features added:
- Preserve tree expansion state during data refreshes using dataTreeStartExpanded
- Add localStorage persistence for expansion state across app sessions
- Track expand/collapse events externally to avoid Tabulator setData() limitations
- Each note maintains independent expansion state keyed by noteId
- Prevent visual collapse/re-expand flicker during table operations

Implementation details:
- Use Tabulator's native dataTreeStartExpanded instead of post-hoc restoration
- External state tracking with React refs for session persistence
- localStorage integration for cross-session state preservation
- Proper event chaining to maintain existing context menu functionality
- Debounced refresh handling to prevent overlapping state updates

### DIFF
--- a/apps/client/src/widgets/collections/table/index.tsx
+++ b/apps/client/src/widgets/collections/table/index.tsx
@@ -26,6 +26,34 @@ interface TableConfig {
 export default function TableView({ note, noteIds, notePath, viewConfig, saveConfig }: ViewModeProps<TableConfig>) {
     const tabulatorRef = useRef<VanillaTabulator>(null);
     const parentComponent = useContext(ParentComponent);
+    const expandedRowsRef = useRef<Set<string>>(new Set());
+    const isDataRefreshingRef = useRef<boolean>(false);
+    
+    // Load persisted expansion state on mount
+    useEffect(() => {
+        const storageKey = `trilium-tree-expanded-${note.noteId}`;
+        try {
+            const stored = localStorage.getItem(storageKey);
+            if (stored) {
+                const expandedIds = JSON.parse(stored);
+                expandedRowsRef.current = new Set(expandedIds);
+                console.log('Loaded expansion state from storage:', expandedIds);
+            }
+        } catch (e) {
+            console.warn('Failed to load tree expansion state:', e);
+        }
+    }, [note.noteId]);
+    
+    // Save expansion state changes to localStorage
+    const persistExpandedState = useCallback(() => {
+        const storageKey = `trilium-tree-expanded-${note.noteId}`;
+        try {
+            const expandedIds = Array.from(expandedRowsRef.current);
+            localStorage.setItem(storageKey, JSON.stringify(expandedIds));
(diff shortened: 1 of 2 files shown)


apps/client/src/widgets/collections/table/index.tsx
```diff
@@ -26,6 +26,34 @@ interface TableConfig {
...
+            const stored = localStorage.getItem(storageKey);
+            if (stored) {
+                const expandedIds = JSON.parse(stored);
+                expandedRowsRef.current = new Set(expandedIds);
+                console.log('Loaded expansion state from storage:', expandedIds);
+            }
+        } catch (e) {
+            console.warn('Failed to load tree expansion state:', e);
+        }
+    }, [note.noteId]);
```
eliandoran, line 45: The idea is good but Trilium does not use the local storage feature at all. All data in Trilium is managed within the SQL database. For storing tree expansion state, ideally you should use the already provided view storage `see ViewModeProps<TableConfig>`.

Since the user can be expanding/collapsing a lot of times, it's best to delay the saving in a spaced update.

apps/client/src/widgets/collections/table/index.tsx
```diff
@@ -39,10 +67,14 @@ export default function TableView({ note, noteIds, notePath, viewConfig, saveCon
...
-            dataTreeElementColumn: "title",
+            dataTreeElementColumn: "title", 
             dataTreeChildIndent: 20,
             dataTreeExpandElement: `<button class="tree-expand"><span class="bx bx-chevron-right"></span></button>`,
-            dataTreeCollapseElement: `<button class="tree-collapse"><span class="bx bx-chevron-down"></span></button>`
+            dataTreeCollapseElement: `<button class="tree-collapse"><span class="bx bx-chevron-down"></span></button>`,
+            persistenceMode: "local",
+            persistence: {
+                tree: true
+            }
```
eliandoran, line 77: As mentioned previously, we don't store data in local storage, we store it in an attachment inside the note. We must not break this functionality.

apps/client/src/widgets/collections/table/index.tsx
```diff
@@ -64,15 +96,94 @@ export default function TableView({ note, noteIds, notePath, viewConfig, saveCon
                         footerElement={<TableFooter note={note} />}
                         events={{
                             ...contextMenuEvents,
-                            ...rowEditingEvents
+                            ...rowEditingEvents,
+                            tableBuilt: () => {
+                                console.log('Table built - setting up tree event tracking');
```
eliandoran, line 101: There are many debug logs inside the PR, please get rid of them.

Issue #7156, opened 2025-10-01
Feature Request: Support custom HTML snippets in shared pages


kleutzinger:
## Problem

Currently, shared pages support custom CSS (`~shareCss`) and JavaScript (`~shareJs`) but there's no way to inject custom HTML snippets. This limits customization options for shared documentation and published notes.

## Use Cases

Users want to add:
- Analytics/tracking scripts (e.g., Google Analytics, Plausible)
- Custom meta tags for SEO or social media
- Comment systems (e.g., Disqus, Utterances)
- Custom headers or footers
- Badges, widgets, or embedded content
- Custom styling or layout modifications via HTML

## Proposed Solution

Add a new `~shareHTML` relation similar to `~shareJs` and `~shareCss`:

```
~shareHTML=@snippetNote
```

The related note would contain raw HTML to be injected into the shared page.

### Location Control

Add an optional `#shareHTMLLocation` label on the snippet note to control injection point:

- `head` - inject in `<head>` section
- `body:start` - inject at start of `<body>`
- `body:end` - inject at end of `<body>`
- `content:start` - inject at start of content area
- `content:end` - inject at end of content area (default)

### Example

**Share page note:**
```
~shareHTML=@analyticsSnippet
~shareHTML=@customFooter
```

**Analytics snippet note:**
```html
<!-- Content -->
<script async src="https://example.com/analytics.js"></script>

<!-- Attributes -->
#shareHTMLLocation=head
```

**Custom footer note:**
```html
<!-- Content -->
<div class="custom-footer">
  <p>© 2025 My Company</p>
</div>

<!-- Attributes -->
#shareHTMLLocation=content:end
```

## Benefits

- Consistent with existing `~shareJs` and `~shareCss` patterns
- Flexible location control for different snippet types
- Supports multiple snippets per page
- No breaking changes to existing functionality
- Enables rich customization of shared pages

## Alternative Considered

Using `~shareJs` with inline HTML strings - but this is awkward and doesn't support non-JS content like meta tags or pure HTML elements.

kleutzinger:
This may be more complex than it has to be. The main place I personally want to inject data is at the end of the head element. So if we only supported that and did not support body and content injection that would be fine by me. And that would let us get rid of the `shareHTMLLocation` label.

kleutzinger:
[...]

PR #7157, opened 2025-10-01
feat: Add shareHtml relation for custom HTML snippets in shared pages

kleutzinger:
closes https://github.com/TriliumNext/Trilium/issues/7156

## Summary

Adds support for injecting custom HTML snippets into shared pages using the `~shareHTML` relation, similar to existing `~shareJs` and `~shareCss` relations.

## Usage

Add `~shareHTML=@snippetNote` relation to your shared page. Optionally add `#shareHTMLLocation=<location>[:<position>]` label to the snippet note to control placement:

**Locations:** `head`, `body`, `content` (default)
**Positions:** `:start`, `:end` (default)

**Examples:**
- `#shareHTMLLocation=head` → end of `<head>`
- `#shareHTMLLocation=body:start` → start of `<body>`
- No label → defaults to `content:end`

Multiple `~shareHTML` relations are supported.

## Changes

- Added `~shareHTML` relation and `#shareHTMLLocation` label to `builtin_attributes.ts`
- Modified `page.ejs` template to collect and inject HTML snippets at specified locations

## Use Cases

Analytics scripts, custom meta tags, comment systems, custom widgets, headers/footers.


apps/server/src/services/builtin_attributes.ts
```diff
@@ -105,6 +106,7 @@ export default [
     { type: "relation", name: "renderNote", isDangerous: true },
     { type: "relation", name: "shareCss" },
     { type: "relation", name: "shareJs" },
+    { type: "relation", name: "shareHTML" },
```
eliandoran: The casing of the relation name needs to be `shareHtml` to match `shareCss`.

kleutzinger: done

eliandoran:
@kleutzinger , when editing documentation, please use `pnpm edit-docs:edit-docs` instead of manually modifying the Markdown files. The reason is that we're actually using the HTML files for the in-app documentation and not the Markdown files.

For this particular case, I will push the change on `main` directly.

eliandoran (approved):

Issue #7085, opened 2025-09-24
[regression] Archived notes are listed in note search/quick search


rom1dep:
### Description

Since the introduction of the new fuzzy search algorithm, archived notes (carrying the attribute `#archived`) started appearing in the result of notes completion/quick search,
This goes against the previous behaviour described [here](https://triliumnext.github.io/Docs/Wiki/attributes.html#:~:text=archived).

CC @perfectra1n 

### TriliumNext Version

0.98

### What operating system are you using?

Windows

### What is your setup?

Local (no sync)

### Operating System Version

Linux

maphew:
Not reproduced from code on current main (1a2f27f8d3).

Observed:
- The search parser puts an `isArchived = false` filter first in every query unless `includeArchivedNotes` is set (`packages/trilium-core/src/services/search/services/parse.ts`, root `AndExp`). The flat-text, content and OCR expressions only see the filtered note set. The fuzzy fallback phase runs the same expression again, so it gets the same filter.
- Quick search (`routes/api/search.ts` `quickSearch`) and jump-to/autocomplete (`searchNotesForAutocomplete`) both set `includeArchivedNotes: false`. A saved or full search follows the `#includeArchivedNotes` label.
- The same filter and defaults are present at v0.97.2 and v0.98.0. The fuzzy-search change did not touch them.
- `search.spec.ts` ("test that fulltext does not match archived notes") exercises the fuzzy fallback phase and expects the archived subtree to stay excluded.

Two paths can still show an archived note. Both are older than 0.98:
- Jump-to with an empty query lists recent notes (`routes/api/autocomplete.ts` `getRecentNotes`). This list has no archived filter.
- A non-inheritable `#archived` label on a parent hides the parent only. Its children are not archived and stay searchable.

Uncertainty: there may be a path that code reading did not find. Advise adding a reproduction: current version, the exact query, whether the note owns or inherits `#archived` and whether the label is inheritable, and whether the list was the empty-query recent-notes list or results for a typed query.

_claude-fable-5-1-high on behalf of matt wilkie_

eliandoran:
Closing as stale.

PR #6839, opened 2025-08-30
feat(search): implement FTS5 w/ sqlite for faster and better searching

perfectra1n:
| Metric | Value |
|--------|-------|
| **Files Changed** | 109 |
| **Lines Added** | +15,575 |
| **Lines Removed** | -1,557 |
| **Net Change** | +14,018 lines |
| **Commits** | 34 |
| **Migration File** | `0234__add_fts5_search.ts` |
| **New Test Files** | 16 spec files |
| **New Indexes Created** | 18 performance indexes |

---

Replaces the JavaScript-based fulltext search with native **SQLite FTS5** (Full-Text Search 5) using trigram tokenization. This enables:

- **50-100x faster** substring matching compared to LIKE queries
- **Language-agnostic search** (works for CJK, Cyrillic, Arabic, etc.)
- **Fuzzy matching** with automatic fallback when exact matches are insufficient
- **Content snippets** with highlighted matches in search results

### Why It Was Implemented

The existing JavaScript-based search had significant performance issues:
- Slow on large databases (10,000+ notes)
- Required loading all note content into memory
- No support for efficient substring matching
- Poor handling of non-Latin scripts

### Risk Assessment

| Area | Risk Level | Notes |
|------|------------|-------|
| Database Migration | **Medium** | Creates virtual tables and triggers; includes rollback strategy |
| Protected Notes | **Low** | Explicitly excluded from FTS index; searched separately via decryption |
| Performance | **Low** | Extensive indexing; 18 new indexes with ANALYZE |
| Backwards Compatibility | **Low** | API unchanged; internal implementation only |

### Priority Review Files

1. `apps/server/src/migrations/0234__add_fts5_search.ts` - Migration logic
2. `apps/server/src/services/search/fts/search_service.ts` - Core FTS operations
3. `apps/server/src/services/search/expressions/note_content_fulltext.ts` - Integration point
4. `apps/server/src/services/search/sqlite_functions.ts` - Custom SQL functions

---

### Database Migration Requirements

- **FTS5 Virtual Tables**: Creates `notes_fts` and `attributes_fts`
- **12 Database Triggers**: Auto-sync FTS indexes with source tables
- **Legacy Table Cleanup**: Removes `note_search_content` and `note_tokens` tables

### API Changes

**None** - All changes are internal implementation details. The search API remains unchanged.

### Backwards Compatibility

- Existing search queries continue to work
- Search operators (`=`, `*=*`, `~=`, etc.) retain their behavior

---

### Protected Notes Handling

[...]

PR #6536, opened 2025-08-03
feat(search): improve search weights and operators

perfectra1n:

PR #7222, opened 2025-10-08
Draft: fix(search-ranking): add attributes and labels to search ranking

werererer:
To improve information Retrieval I made some changes to the ranking function:
- Search now considers note label key and values different and tunable in ranking (It didn't before was a bug prob).
- Search now considers the Note-ID
- Made Ranking algorithm more straight forward (likely the first version was similar to the now second) 

_Note_: The search result order will differ due to new rankings. However I compared the ranking with the ranking from the original trilium, and tried to match and improve upon it, by testing it out in practice and _hoping_ that my experience generalizes xD.

The code works in the current stage. But to fine tune it, I will run with it for a bit, tweak it and make it compatible with other wishes of users.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14718,8 +14718,6 @@ snapshots:
       '@ckeditor/ckeditor5-utils': 47.0.0
       '@ckeditor/ckeditor5-watchdog': 47.0.0
       es-toolkit: 1.39.5
-    transitivePeerDependencies:
-      - supports-color
 
   '@ckeditor/ckeditor5-dev-build-tools@43.1.0(@swc/helpers@0.5.17)(tslib@2.8.1)(typescript@5.9.3)':
     dependencies:
@@ -14933,8 +14931,6 @@ snapshots:
       '@ckeditor/ckeditor5-table': 47.0.0
       '@ckeditor/ckeditor5-utils': 47.0.0
       ckeditor5: 47.0.0(patch_hash=8331a09d41443b39ea1c784daaccfeb0da4f9065ed556e7de92e9c77edd9eb41)
-    transitivePeerDependencies:
-      - supports-color
 
   '@ckeditor/ckeditor5-emoji@47.0.0':
     dependencies:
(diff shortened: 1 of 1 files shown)


itsjakobx:
## Issues Solved
- [X] Remove Archived Notes from ranking
https://github.com/TriliumNext/Trilium/issues/7085 
- [X] **PARTIAL SOLVED**? (Untested for that particular usecase) Still testing. Caches were introduced to hopefully increase speed.
https://github.com/TriliumNext/Trilium/issues/6808

itsjakobx:
Associations: 
- Improves:  https://github.com/TriliumNext/Trilium/pull/6536 

Dissociations:
- It should not interfer much with. Maybe smaller conflicts to be expected :/ resulting from _syntax_ not semantics
https://github.com/TriliumNext/Trilium/pull/6839 
OR 
https://github.com/TriliumNext/Trilium/pull/6810

itsjakobx:
I am still quite confused tbh, since it is kinda an open question in research how to scale up search reliably: However looking at trilium as a potential folksonomy, the question of how to scale up search is a bit non trivial, but might be solved using equivalence classes instead.

Syntax matching won't scale past a certain note count. Web 2.0 was build using folksonomies. And folksonomies failed using that approach.

One solution might be to actually implement equivalence classes. I am currently drafting ideas here:
https://github.com/TriliumNext/Trilium/issues/1045

**Research pointing out that Equivalence classes might be correct abstraction**:

Varese & Castano, 2009, Building Collective Tag Intelligence through Folksonomy Coordination, https://link.springer.com/chapter/10.1007/978-3-642-20344-2_4 [They say that legit equivalence classes might be the solution]

Effective Retrieval of Resources in Folksonomies Using a New Tag Similarity Measure, https://arxiv.org/abs/1207.6033 [They propose clustering, which is just (imo) statistical equivalence classes]

Tag Similarity in Folksonomies, https://inforsid.fr/actes/2013/2013_5_1%20Mousselly%20Sergieh.pdf [Again clustering, and again i believe just statistical equivalence classes]

**Conclusion**
I will have this open for a bit more, and will work on equivalence classes instead, till a clearer pattern emerges. Maybe there is a generalized solution to scaling up search that no PKM has touched yet.

eliandoran:
Hi, @werererer .

What's the status of this PR? Are you comfortable with the changes here or should we drop it?

werererer:
@eliandoran  lets drop it, its hard to come up with a satisfying solution

PR #7268, opened 2025-10-10
fix(search): resolve issue when using `=` operator in search

perfectra1n:
- make sure that `=text` returns matches that _only_ include "text"
- create additional e2e and unit tests to validate change

Phrase example:
image

Word example (with working highlighting!)
image

### DIFF
--- a/apps/server/src/services/search/expressions/note_content_fulltext.ts
+++ b/apps/server/src/services/search/expressions/note_content_fulltext.ts
@@ -96,14 +96,14 @@ class NoteContentFulltextExp extends Expression {
 
         // For exact match with flatText, also search notes WITHOUT content (they may have matching attributes)
         if (this.flatText && (this.operator === "=" || this.operator === "!=")) {
-            for (const noteId of inputNoteSet.noteIdSet) {
+            for (const note of inputNoteSet.notes) {
                 // Skip if already found or doesn't exist
-                if (resultNoteSet.hasNoteId(noteId) || !(noteId in becca.notes)) {
+                if (resultNoteSet.hasNoteId(note.noteId) || !(note.noteId in becca.notes)) {
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


contributor:
@perfectra1n search with `=` operator only shows one note for me in this PR. But actually there are multiple notes in db.

perfectra1n:
@contributor can you provide more details? Was the search term `='test'` or was it `='test phrase'`? How many notes would you expect for it to have found, were they all text notes?

contributor:
@perfectra1n 

Commit 4fa41128407c5ce92b1a45e8fb58f8d14d148df1

The search term `=pagio` and `='pagio'` returns the same (incorrect) result - only one note.
I have 45 notes with word pagio.

The search term `pagio` return 114 notes - this includes false positives due to fuzzy search (the first 45 of them are correct notes).

apps/server/src/services/search/expressions/note_content_fulltext.ts
```diff
@@ -75,20 +75,101 @@ class NoteContentFulltextExp extends Expression {
...
-                  AND isDeleted = 0 
+                WHERE type IN ('text', 'code', 'mermaid', 'canvas', 'mindMap')
+                  AND isDeleted = 0
                   AND LENGTH(content) < ${MAX_SEARCH_CONTENT_SIZE}`)) {
             this.findInText(row, inputNoteSet, resultNoteSet);
         }
 
+        // For exact match with flatText, also search notes WITHOUT content (they may have matching attributes)
+        if (this.flatText && (this.operator === "=" || this.operator === "!=")) {
+            for (const noteId of inputNoteSet.noteIdSet) {
```
eliandoran: Can you investigate this build failure?

```
Property 'noteIdSet' is private and only accessible within class 'NoteSet'.
```

eliandoran (changes requested):

PR #7406, opened 2025-10-19
docs: remove garbage from Amazon URL

tredondo:
This sort of sloppiness undermines trust in the application: what other cruft is there that we don't see?


eliandoran:
> This sort of sloppiness undermines trust in the application: what other cruft is there that we don't see?

That's kind of an overstatement, don't you think? Clean URLs are not a requirement, especially in a _demo_ note.

tredondo:
I agree that clean URLs are not a requirement in a demo note, I just think they show a lack of attention: when pasting the URL, when committing the code containing it, and from whoever reviewed the commit.

Anyway, sorry for my comment, and thanks for accepting the PR.

Issue #4956, opened 2024-04-16
Milestone: Multi-user support


meichthys:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24832%20Funded-%2300A156.svg" />](https://oss.issuehunt.io/r/TriliumNext/Trilium/issues/4956)

Many feature requests and issues exist for multiple user support. Some of the most relevant issues in the original Trilium repo are:
(https://github.com/zadam/trilium/issues/3546, https://github.com/zadam/trilium/discussions/1375, https://github.com/zadam/trilium/issues/2679, https://github.com/zadam/trilium/issues/2143, https://github.com/zadam/trilium/issues/2109, https://github.com/zadam/trilium/issues/1527, https://github.com/zadam/trilium/issues/1436, https://github.com/zadam/trilium/issues/964, https://github.com/zadam/trilium/issues/600, https://github.com/zadam/trilium/issues/450, https://github.com/zadam/trilium/issues/169)

Instead of migrating each of the issues, we will link to them here, and if desired, Zadam can close the issues on his repo and link to this issue.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $832.00)

- [<img src='https://avatars.githubusercontent.com/u/245382?v=4' alt='eitch' width=24 height=24> eitch](https://oss.issuehunt.io/u/eitch) ($52.00)
- [<img src='https://avatars.githubusercontent.com/u/4681318?v=4' alt='deajan' width=24 height=24> deajan](https://oss.issuehunt.io/u/deajan) ($500.00)
- [<img src='https://avatars.githubusercontent.com/u/1061866?v=4' alt='metrax' width=24 height=24> metrax](https://oss.issuehunt.io/u/metrax) ($50.00)
- [<img src='https://avatars.githubusercontent.com/u/42046082?v=4' alt='werererer' width=24 height=24> werererer](https://oss.issuehunt.io/u/werererer) ($20.00)
- [<img src='https://avatars.githubusercontent.com/u/5337145?v=4' alt='sirhaug' width=24 height=24> sirhaug](https://oss.issuehunt.io/u/sirhaug) ($200.00)
- [<img src='https://avatars.githubusercontent.com/u/3864731?v=4' alt='josegabrielhc' width=24 height=24> josegabrielhc](https://oss.issuehunt.io/u/josegabrielhc) ($10.00)

### Submitted pull Requests
[...]

PR #7441, opened 2025-10-21
feat: add multi-user support (issue #4956)

Somoru:
## Summary

This PR implements comprehensive multi-user support for Trilium Notes, enabling multiple users to collaborate on the same Trilium instance with role-based access control while maintaining full backward compatibility with existing single-user installations.

## Changes

- Add database migration v234 for multi-user schema
- Implement users, roles, user_roles, and note_shares tables
- Add user management service with CRUD operations
- Implement role-based permission system (Admin/Editor/Reader)
- Add RESTful user management API endpoints
- Update login flow to support username + password authentication
- Maintain backward compatibility with legacy password-only login
- Create default admin user from existing credentials during migration
- Add session management for multi-user authentication
- Include TypeScript type definitions for Node.js globals

**Tests:** 948 passed | 17 skipped (965 total)  
**Build:** Successful (server and client)  
**TypeScript:** Zero errors

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#4956: multi-user support](https://oss.issuehunt.io/repos/92111509/issues/4956)
---
</details>


apps/server/src/migrations/migrations.ts
```diff
@@ -6,6 +6,11 @@
 
 // Migrations should be kept in descending order, so the latest migration is first.
 const MIGRATIONS: (SqlMigration | JsMigration)[] = [
+    // Multi-user support
+    {
+        version: 234,
+        module: async () => import("./0234__multi_user_support.js")
+    },
```
eliandoran, line 13: The migration is not applied at all since the database version in `app_info` was not changed.

apps/server/src/migrations/0234__multi_user_support.ts
```diff
@@ -0,0 +1,204 @@
...
+            sharedWithUserId TEXT NOT NULL,
+            permission TEXT NOT NULL DEFAULT 'read',
+            utcDateCreated TEXT NOT NULL,
+            utcDateModified TEXT NOT NULL,
+            isDeleted INTEGER NOT NULL DEFAULT 0,
+            FOREIGN KEY (noteId) REFERENCES notes(noteId) ON DELETE CASCADE,
+            FOREIGN KEY (ownerId) REFERENCES users(userId) ON DELETE CASCADE,
+            FOREIGN KEY (sharedWithUserId) REFERENCES users(userId) ON DELETE CASCADE
+        )
+    `);
```
eliandoran: If the tables are not used yet (e.g. in order to be able to share notes), there is no reason to create them.

apps/server/src/migrations/0234__multi_user_support.ts
```diff
@@ -0,0 +1,204 @@
...
+import optionService from "../services/options.js";
+import { randomSecureToken, toBase64 } from "../services/utils.js";
+import myScryptService from "../services/encryption/my_scrypt.js";
+
+export default async () => {
+    console.log("Starting multi-user support migration (v234)...");
+
+    // 1. Create users table
+    sql.execute(`
+        CREATE TABLE IF NOT EXISTS "users" (
```
eliandoran: There is already a `user_info` table that came from multi-factor authentication, do you think we can use that instead? This would avoid requiring two different tables for user management.

apps/server/src/routes/login.ts
```diff
@@ -114,6 +116,7 @@ function login(req: Request, res: Response) {
 
     const submittedPassword = req.body.password;
     const submittedTotpToken = req.body.totpToken;
+    const submittedUsername = req.body.username; // New field for multi-user mode
```
eliandoran, line 127: There seems to be server-side support for user management, but I don't see any way to create a new user or to login, as user.

Without the front-end functionality to back it up, it would be non-functional.

apps/server/src/routes/routes.ts
```diff
@@ -224,6 +225,15 @@ function register(app: express.Application) {
...
     apiRoute(PST, "/api/password/reset", passwordApiRoute.resetPassword);
 
+    // User management routes
+    apiRoute(GET, "/api/users/current", usersRoute.getCurrentUser);
+    apiRoute(GET, "/api/users/check-username", usersRoute.checkUsername);
+    route(GET, "/api/users", [auth.checkApiAuth, csrfMiddleware, auth.checkAdmin], usersRoute.getUsers, apiResultHandler);
+    route(GET, "/api/users/:userId", [auth.checkApiAuth, csrfMiddleware], usersRoute.getUser, apiResultHandler);
+    route(PST, "/api/users", [auth.checkApiAuth, csrfMiddleware, auth.checkAdmin], usersRoute.createUser, apiResultHandler);
+    route(PUT, "/api/users/:userId", [auth.checkApiAuth, csrfMiddleware], usersRoute.updateUser, apiResultHandler);
+    route(DEL, "/api/users/:userId", [auth.checkApiAuth, csrfMiddleware, auth.checkAdmin], usersRoute.deleteUser, apiResultHandler);
```
eliandoran, line 237: These endpoints need to be used somewhere in the client. Preferably as a separate setting where the CRUD operations can take place.

apps/server/src/services/user_management.ts
```diff
@@ -0,0 +1,421 @@
...
+        passwordHash: user.passwordHash,
+        passwordSalt: user.passwordSalt,
+        derivedKeySalt: user.derivedKeySalt,
+        encryptedDataKey: user.encryptedDataKey,
+        isActive: Boolean(user.isActive),
+        isAdmin: Boolean(user.isAdmin),
+        utcDateCreated: user.utcDateCreated,
+        utcDateModified: user.utcDateModified
+    };
+}
```
eliandoran: The user management doesn't seem to match Trilium's model for the becca, where each user is a model.

How are the users synchronized across multiple instances?

apps/server/src/migrations/migrations.ts
Somoru: Fixed!  I've updated APP_DB_VERSION from 233 to 234 in apps/server/src/services/app_info.ts. The migration now triggers automatically on server start.

Addressed in commit ccaabcf93.

apps/server/src/migrations/0234__multi_user_support.ts
Somoru: Fixed! You're absolutely right. I've removed the unused user_roles and note_shares tables from the migration. The implementation now only creates what's actually being used - we extend the existing user_data table with role/status fields instead.

Per-note sharing is documented as a future enhancement in MULTI_USER.md.

Addressed in commit 883ca1ffc.

apps/server/src/migrations/0234__multi_user_support.ts
Somoru: Good question! After reviewing the codebase, I discovered there's actually a user_data table (not user_info) from migration v229 for OAuth support. That's what I'm using now.

Why user_data?
- Already has the password hashing infrastructure (userIDVerificationHash, salt, derivedKey)
- Uses tmpID as INTEGER primary key (matches OAuth design)
- Has username/email fields we need
- Established table with existing scrypt parameters

I've extended this table with just the fields we need: role, isActive, and timestamps. This integrates cleanly with Trilium's existing authentication infrastructure.

Documented in detail in MULTI_USER.md under "Architecture" section.

apps/server/src/services/user_management.ts
Somoru: Great question - this is an architectural decision I made deliberately:

Why users are NOT Becca entities:
• Security: User authentication data should NEVER be synced across instances. Each Trilium instance needs its own isolated user database.
• Becca is for content: The Becca entity model is designed for synchronized content (notes, branches, attributes). User credentials are infrastructure, not content.
• Sync risks: Syncing passwords/hashes across instances would create massive security vulnerabilities.

Future sync support:
When we do implement multi-user sync, it will require:
- Per-user sync credentials on each instance
- User-to-user mappings (Alice on Instance A → Alice_Remote on Instance B)
- Content sync separate from authentication
- Each instance maintains its own user_data table

I've documented this extensively in MULTI_USER.md under "Architecture" > "Why not Becca entities?" and listed sync as a future enhancement with specific requirements.

Does this approach make sense to you, or would you like to discuss alternative architectures?

apps/server/src/routes/routes.ts
Somoru: Currently these are backend-only REST APIs. I've documented them comprehensively in MULTI_USER.md with:
- Full API reference for all 7 endpoints
- curl examples for each operation
- Authentication requirements
- Response formats and error handling

Question: Would you prefer:
1. Ship backend-only (documented API) for initial release, add UI later
2. I create a basic settings dialog for user management before merging
3. Wait until someone creates a full UI implementation

My recommendation would be (1) - get the solid backend merged, then iteratively add UI. But I'm happy to implement whatever approach you think is best for Trilium.

apps/server/src/routes/login.ts
Somoru: Login UI: Already implemented! The login page (login.ejs) shows a username field when multi-user mode is detected (when user_data table has users). Users can log in with username + password.

User Creation: Currently API-only (no UI yet). This is documented in MULTI_USER.md with complete curl examples:

curl -X POST http://localhost:8080/api/users \
  -H "Content-Type: application/json" \
  -d '{"username": "newuser", "password": "securepass", "role": "user"}'

I've documented this as a known limitation with "Future Enhancement: UI for User Management" in the documentation. Would you prefer I add a basic UI dialog for user management before merging, or is API-only acceptable for initial release?

apps/server/src/services/user_management.ts
eliandoran: @Somoru , I understand and there's nothing wrong with that.
However, from your statement I also understand that syncing does not work when multi-user is enabled? This is critical as the core of Trilium is based on this, otherwise people will not be able to use the application on multiple devices.

eliandoran:
@Somoru ,

I've left a comment here: https://github.com/TriliumNext/Trilium/pull/7441#discussion_r2448468752

> However, from your statement I also understand that syncing does not work when multi-user is enabled? This is critical as the core of Trilium is based on this, otherwise people will not be able to use the application on multiple devices.

apps/server/src/services/user_management.ts
Somoru: @eliandoran You're absolutely right - this is a critical point I should have highlighted more clearly upfront.

Current State & Limitation:
The current implementation does NOT support sync when multi-user is enabled. This is indeed a blocker for users who rely on Trilium across multiple devices.

My recommendation for this PR:
Since this is a significant architectural challenge, I propose we:

1. Merge this PR as "multi-user backend foundation" with clear documentation that sync is not yet supported
2. OR - I add a runtime check that disables multi-user mode if sync is configured (fail-safe approach)
3. OR - I implement basic sync support before merging (would need guidance on your preferred approach)

Which approach would you prefer?

For future sync implementation, I envision:
• Each synced instance maintains its own user_data table (local auth)
• Sync protocol includes userId metadata with each entity
• Content is synced, but filtered per-user on each instance
• User mappings stored locally (Alice@Instance1 → Alice@Instance2)

This way authentication stays local (secure) but content syncs properly. Would this architecture work with Trilium's sync protocol?

I want to make sure we ship something that works correctly rather than breaking sync for existing users. Let me know how you'd like me to proceed.

rom1dep:
Interesting PR. May I ask what's the incentive/motivation for it and for that approach specifically?
_warning: my technical understanding of the subject is admittedly superficial_

On a purely practical level, Trilium is a personal note taking application: users edit notes for themselves only (there is no "multiplayer" feature involving collaboration on shared notes). In that regard, multi-users support is already delivered today by running concurrently multiple servers (for instance, on my physical server, I effectively host 4 different Trilium users, each having their own Trilium instance and notes, in isolation). The caveat here is that each user needs to be directed to their own/dedicated instance (under a specific domain, path, …) instead of sharing one canonical host.  
If this is all what this PR stands to address, perhaps a simpler approach to consider would be to develop a new (+optional) "Trilium proxy server", taking requests centrally, and routing them to the corresponding user-dedicated Trilium instance. I suppose that the implementation complexity would be drastically simplified.

Now, if the motivation behind this PR is to serve as a stepping stone towards real collaborative workflows with multiple users, I think the whole roadmap would have first to be laid out and carefully reviewed: lots of people already have great but competing ideas (along the lines of making Trilium a fully-encrypted/zero-knowledge tool, using CRDTs over it for real-time collaboration, or meshing peer-to-peer over a federation of public/private servers, etc).

What I would personally consider an incremental (but welcome) improvement in the area of collaboration would be to add the ability to specify additional sync targets, and share a specific subtree with them (essentially the equivalent of having a new document.db "in the middle" of two existing instances). That would entail improving the existing sync code to be selective (which can help in certain scenarios, like, with only replicating the "work" subtree on a work computer, or lowering the bandwidth for mobile users).

I take that this last bit is irrelevant to this PR, but it serves to illustrate that multi-user support can be achieved in different manners and it would be nice to have consensus on the best way forward.

Somoru:
@rom1dep Great points - thanks for the thoughtful feedback! Let me address your questions:

Motivation & Use Case:
The original issue #4956 requested multi-user support for shared Trilium instances).
 The use case is:
- Multiple users on ONE server instance (e.g., family NAS, small office server)
- Each user has isolated notes with their own authentication
- Admins can manage users via API
- Lower resource usage than running N separate Trilium instances

You're absolutely right that for pure isolation, multiple instances work fine. This PR targets the "shared instance" use case where that's overkill.

Current Approach vs. Alternatives:

1. Proxy/Router approach: Valid alternative! The tradeoff is:
   - Proxy: N databases, N processes, higher memory/CPU, but perfect isolation
   - This PR: 1 database, 1 process, shared resources, user-level permissions
   
2. Your point about collaborative workflows: Agreed - this is NOT about real-time collaboration. It's about multi-tenancy on a single instance. True collaboration (CRDTs, etc.) would be a separate, much larger effort.

3. Selective sync/subtree sharing: Love this idea! That's actually closer to what collaborative Trilium might look like long-term.

Critical Issue (Sync):
@eliandoran correctly identified that this PR currently BREAKS sync for multi-user scenarios. I'm waiting for guidance on whether to:
- Add a fail-safe (disable multi-user if sync enabled)
- Implement basic multi-user sync
- Ship as "single-instance only" feature

Architecture Discussion:
You raise a valid question: is modifying core Trilium for this use case the right approach? Alternatives:
- Separate proxy layer (as you suggested)
- Optional plugin architecture
- Helm chart / container orchestration for multiple instances

I'd love @eliandoran's input on whether this belongs in core or should be reconsidered as an external tool.

Bottom line: I'm not attached to this specific approach - I'm trying to solve the "family server" use case. If there's a better architecture, I'm happy to pivot.

deajan:
@Somoru one of the goals of Trilium is to be able to work offline (eg sync later).
I would strongly avocate that multi user support still needs to work with sync (the whole idea behind the bounty is small business / family scenarios).

Perhaps the user that syncs will only sync it's own credentials & notes he's allowed to in order to avoid credential changes / privilege escalation.

Somoru:
@deajan Absolutely - you're 100% right! The whole point of the bounty IS multi-user with sync support for family/small business scenarios.

Current Status:
Right now, this PR has the backend foundation but breaks sync. This needs to be fixed before it's ready.

My Plan:
I'll implement basic multi-user sync support where:
- Each user's notes sync to their own devices
- User credentials stay local (not synced)
- Each synced instance knows which user owns which notes
- Users can work offline and sync later (as you mentioned)

Implementation approach:
1. Add userId metadata to synced entities (notes, branches, etc.)
2. Each instance filters synced content by user
3. Credentials/auth stays local per instance
4. User can sync their notes across their laptop/phone/tablet

This way:
✅ Alice can use Trilium offline on her laptop, sync later
✅ Bob can use Trilium offline on his phone, sync later  
✅ Both share the same server, but their notes stay isolated
✅ No credential syncing (security risk)

Question for @eliandoran:
Should I:
Implement this sync approach in this PR?

I want to make sure this delivers on the bounty requirements properly. Let me know your preference and I'll get to work!

deajan:
@Somoru There's something in your design that doesn't check out.

Alice creates note X.
Alice and Bob both belong to group "Family" which allows create/read/update but no delete.
Bob should be able to sync note X to his local instance, modify it, and resync later.

Note isolation is already achievable quite cheaply, so that's not the scope here.
The point is to be able to view/edit notes from other users in the same instance, eg I have thousands of notes, a coworker should be able to view my notes on project A, and edit them on project B (projects being notes with child notes in this example).


I also think that credentials should be synced with server instance, for an admin to be able to modify them when forgotten by a user, and more generally add/edit users and groups (the admin interface should only exist on server obviously).
There's no security concern about syncing the credentials here, since a user generally trusts the server where he'd put all his notes. The only point is that user A should never sync user B credentials or notes on which he has no grants.
The sync process would require to verify that user B has permissions to update user A's notes, in order to reflect permission changes that could happen server side by an admin between two user syncs.

Somoru:
I've implemented an alternative approach to multi-user support that addresses the sync blocker mentioned in the comments. Instead of isolating users into separate databases, this implementation uses a **collaborative model with permission-based filtering**.

### Key Difference: Sync Support

The critical issue raised was that sync doesn't work in multi-user mode. This implementation solves that by filtering sync data based on permissions:

```typescript
// In sync.ts - permission filtering during sync
const userId = cls.get('userId');
if (userId) {
    entityChanges = permissions.filterEntityChangesForUser(userId, entityChanges);
}
```

Users only sync notes they have access to, enabling:
- Multiple users working on shared notes
- Private notes that don't sync to other users
- Granular read/write/manage permissions
- Group-based access control

### What's Implemented

**Database Schema:**
- `users` - User accounts with scrypt password hashing
- `groups` - User groups for permission management  
- `note_ownership` - Tracks note owners
- `note_permissions` - Granular access control (read/write/manage)
- `group_members` - Group membership

**API Endpoints (14 total):**
- Permission management: share, revoke, transfer ownership
- Group management: create, delete, add/remove members
- User management: create, list, update roles

**Core Services:**
- Permission checking and filtering
- Group lifecycle management
- User authentication with timing-safe validation

### Architecture

- **Backward compatible** - existing single-user setups work unchanged
- **Permission-aware sync** - the key differentiator from PR #7441
- **Context propagation** - userId flows through request chain via CLS
- **Ownership tracking** - all notes have owners, existing notes assigned to admin

### Documentation

- `MULTI_USER_README.md` - User guide
- `COLLABORATIVE_ARCHITECTURE.md` - Technical deep-dive
- `ADDRESSING_PR_7441.md` - Detailed comparison with PR #7441

Branch: `feat/multi-user-support`

deajan:
@Somoru Your new PR looks way better and cleaner integrated into existing sync process.
As side question, did you include an admin UI on the server so users/groups can be created/updated or is that a separate job ?

[EDIT] Read a bit of your code. There's something missing in the user deletion function. When a user gets deleted, the nodes he owns become orphans, perhaps not readable by anyone, but cluttering the database. The deletion code should be improved to allow nodes to be assigned to another user, or be deleted together with owner user.

Do you have a test build somewhere I can try ?

Also, why do you update alot of ckeditor stuff in the PR ?
[/EDIT]
[EDIT 2] Be careful when copying AI responses, you copied everything double in https://github.com/TriliumNext/Trilium/pull/7441#issuecomment-3433175333 [/EDIT 2]

contributor:
> What I would personally consider an incremental (but welcome) improvement in the area of collaboration would be to add the ability to specify additional sync targets, and share a specific subtree with them (essentially the equivalent of having a new document.db "in the middle" of two existing instances). That would entail improving the existing sync code to be selective (which can help in certain scenarios, like, with only replicating the "work" subtree on a work computer, or lowering the bandwidth for mobile users)

@rom1dep Partial sync is a great suggestion. Is there an issue that tracks that?

> Backward compatible - existing single-user setups work unchanged

@Somoru Could you provide more details on how is it going to work? Currently desktop electron uses the same server app that server web instance. Will there be two server apps? With multi-user support to run it on server? And without multi-user support to ship it in desktop electron app?

deajan:
> @Somoru Could you provide more details on how is it going to work? Currently desktop electron uses the same server app that server web instance. Will there be two server apps? With multi-user support to run it on server? And without multi-user support to ship it in desktop electron app?

AFAIK, the server has admin API and keeps track of users/groups. The desktop app only syncs the subset of what is allowed by the server permissions.

@rom1dep Partial sync is a great suggestion. Is there an issue that tracks that?

For instance, partial sync is implemented per user basis in this PR. I think it wouldn't be much more effort to create another token that would allow to specify sync targets, but that's out of scope here.
Perhaps another bounty ?

contributor:
> AFAIK, the server has admin API and keeps track of users/groups. The desktop app only syncs the subset of what is allowed by the server permissions.

@deajan But the desktop app includes the server app in its bundle, and runs the same server in background. Adding multi-user support to server is equal to adding that to desktop app. This is what concerns me. Ideally, all multi-user code should be external to server app, so there is a possibility to include it only in server instance (but not by default, because most of users still only need single-user server). So the goal is to exclude multi-user overhead from:
- single-user server
- all desktop apps

rom1dep:
> @rom1dep Partial sync is a great suggestion. Is there an issue that tracks that?

@contributor not that I'm aware of, but I doubt my suggestion truly is original!

> For instance, partial sync is implemented per user basis in this PR. I think it wouldn't be much more effort to create another token that would allow to specify sync targets, but that's out of scope here.

@deajan sounds sensible. I might argue that in the current state of the ecosystem, this approach would probably have a greater impact than the 1-server:N-users being promoted here (there are already tens of thousands of existing and historically secluded Trilium instances in the wild that, all a sudden, would start "seeing each other" and turning able to share/edit notes together). But this ship is sailing, it appears :-)

> Perhaps another bounty ?

Ohh, I hadn't noticed that this is pegged to a bounty. And holly molly, that's a generous one!

Somoru:
@deajan Thanks for the positive feedback! Addressing your questions:

### Admin UI
Currently there's no admin UI - only REST API endpoints. Admin users would need to use curl/Postman or build a simple frontend. I can add a basic admin panel if that's required for the bounty. Let me know if you'd prefer that.

### User Deletion - Orphan Notes (Fixed)
Good catch! I've just pushed a fix. When a user is deleted:
- All their owned notes are automatically transferred to the first active admin
- Optionally specify a target user: `DELETE /api/users/:id?transferTo=<userId>`
- User is removed from all groups and note permissions
- Prevents orphan notes cluttering the database

### CKEditor Changes
These weren't intentional - just dependency version bumps (`@typescript-eslint/parser`, `lint-staged`) and generated build artifacts (`.js.map` files) that came from merging the latest main branch. I can revert them if they shouldn't be in the PR.

### Test Build
No hosted demo currently. To test locally:

```bash
git clone -b feat/multi-user-support https://github.com/Somoru/Trilium.git
cd Trilium
pnpm install
pnpm run client:build
pnpm run server:build
pnpm run desktop:start

**Comment by Somoru** [comment]
@contributor Good point about the desktop app overhead - this is a valid architectural concern.

**Current Implementation:**
You're right that the desktop app bundles the server code, and currently the multi-user code is integrated into the core server. This does add overhead to:
- Desktop apps (bundled code)
- Single-user server deployments
- Code complexity even when unused

**The Overhead:**
- ~5 new database tables (users, groups, note_ownership, note_permissions, group_members)
- Permission checking services
- Group management logic
- 14 REST API endpoints

**Trade-offs with Current Approach:**
- Simpler codebase (one unified server)
- Backward compatible (detects single-user scenarios automatically)
- Bloat for desktop/single-user deployments
- Code runs even when multi-user is unused

**Possible Solutions:**

1. **Plugin Architecture** (cleanest but most work):
   - Multi-user as optional plugin
   - Desktop/single-user ships without it
   - Server deployments can install plugin

2. **Build-time Flag** (moderate effort):
   - Compile two server variants: `server-single` and `server-multi`
   - Desktop uses `server-single`
   - Deploy multi-user servers with `server-multi`

3. **Runtime Detection with Tree-Shaking** (less ideal):
   - Current approach but with better dead-code elimination
   - Still ships unused code

Would the maintainers prefer I refactor to a plugin architecture? That would address your concern about keeping multi-user overhead external to the core server. I'm happy to pivot if that's the preferred direction.

**Comment by deajan** [comment]
@Somoru 
> Admin UI
Currently there's no admin UI - only REST API endpoints. Admin users would need to use curl/Postman or build a simple frontend. I can add a basic admin panel if that's required for the bounty. Let me know if you'd prefer that.

There is definitly a need for a admin user panel, since that feature needs to be usable for everyone having more than one user (can't imagine a family sharing trilium having tu use curl requests).
I'd go for the admin UI, but indeed @eliandoran will have the last word on this.

@Somoru @contributor Multi-user overhead is fairly small (5 almost empty SQL tables, and permission code).
Sync permission checks could be handled optionally, eg if there is more than one user, only then have sync perform permission checks.
This would reduce code splitting between desktop and server app which is IMO very important to keep the codebase clean and bug-free(TM) for future modifications that won't need to handle multiple different codebases, while keeping desktop app fast as if there weren't multiple users.

**Comment by contributor** [comment]
> @Somoru @contributor Multi-user overhead is fairly small (5 almost empty SQL tables, and permission code).

@deajan 
I doubt permission code overhead will be small. Especially read permissions, when you need to to check that for massive amount of notes (search, trees etc)

**Comment by deajan** [comment]
> > @Somoru @contributor Multi-user overhead is fairly small (5 almost empty SQL tables, and permission code).
> 
> @deajan I doubt permission code overhead will be small. Especially read permissions, when you need to to check that for massive amount of notes (search, trees etc)

That's where I propose to bypass permission checks when only one user exists, which obviously happens on desktop client.

**Comment by Somoru** [comment]
@deajan @contributor @eliandoran 
Before I make the UI changes to implement Admin dashboard and other functionalities, I just want to know, "Does electron app also need multi-user support" if so, I will have to implement a login screen for that aswell and "not bypass" authentication for it. If not, I will just bypass the multi-user feature for electron app.

**Comment by contributor** [comment]
> Before I make the UI changes to implement Admin dashboard and other functionalities, I just want to know, "Does electron app also need multi-user support" if so, I will have to implement a login screen for that aswell and "not bypass" authentication for it. If not, I will just bypass the multi-user feature for electron app.

It is not so simple. Electron does not need username if it is not synced to any server (the fact that you don't know at advance). It definitely need username if it syncs to multi-user server.

**Comment by deajan** [comment]
> @deajan @contributor @eliandoran Before I make the UI changes to implement Admin dashboard and other functionalities, I just want to know, "Does electron app also need multi-user support" if so, I will have to implement a login screen for that aswell and "not bypass" authentication for it. If not, I will just bypass the multi-user feature for electron app.

Basically, the electron app needs to identify with the server in order to know which user's notes need to be synced, but it does not need full multi user support since only one user will use it at the time.

My best guess is to:
- Server's first user (password only ATM) is admin
- If "classic" login (ie password) is used on electron client, no behavior changes, just sync everything
- Electron client should have possibility to login as user (existing on server), and sync said user's data

In terms of UX, there should be something like "login as server admin" or "login as user" when starting electron client.
Once user is logged in, there could be a checkbox like "remember me" to avoid user logins on every run, for those who would want this workflow

**Comment by eliandoran** [comment]
@Somoru , before implementing I would suggest you take a step back and propose a full end-to-end plan, especially given the login and the sync constraints.

**Comment by deajan** [comment]
> @Somoru , before implementing I would suggest you take a step back and propose a full end-to-end plan, especially given the login and the sync constraints.

Can't agree more, there have been enough implementation discussions for a good recap to be necessary.

**Comment by deajan** [comment]
@Somoru I do understand that we asked you alot of stuff in this thread, but this feature isn't exactly the easiest out there, and will have some major implications. Perhaps you would use more of the bounty if Elian is okay with that.
I think Warp can write a up a summary for you, but keep in mind that it creates logical fallacy sometimes, which is why we devs still exist :)

Looking forward hearing from you.

**Comment by Somoru** [comment]
# PR #7441 Current State & Implementation Plan

@deajan @eliandoran thank you for your responses, suggestions and the motivation. please check my plan below. If it suits with you, I will start building and hopefully, will deliver soon

The backend is substantially complete with working authentication, permissions, sync filtering, and group management. All users authenticate with username+password (mandatory, no fallbacks). Username is the primary user identifier throughout the system. The application currently lacks the frontend UI layer required to make these features accessible to users.

The simplified authentication model eliminates complexity: every user must have a username, password is always required, and this serves as the foundation for team generation and collaboration features.

---

## What's Already Implemented

### Backend Services (95% Complete)

#### 1. Authentication (`services/auth.ts`)
- Multi-user authentication middleware
- Session management with userId propagation via Continuation-Local-Storage (CLS)
- Support for both single-user password mode and multi-user username+password mode
- Backward compatibility: Single-user instances default to admin user (userId=1)

#### 2. User Management (`services/user_management_collaborative.ts`)
- User creation with secure scrypt hashing (N=16384, r=8, p=1)
- Credential validation with timing-attack protection
- Password changing functionality
- User listing and retrieval
- User roles: admin and user

#### 3. Permissions (`services/permissions.ts`)
- Fine-grained permission levels: read, write, admin
- Permission resolution: owner implicit admin > direct permissions > group permissions
- Functions for checking access, getting permission levels, filtering changes
- Support for both user and group-based permissions

#### 4. Group Management (`services/group_management.ts`)
- Group creation and deletion
- User group membership management
- Group-based permission inheritance
- "All Users" group for organization-wide sharing

#### 5. Sync with Permission Filtering (`routes/api/sync.ts`)
- Pull sync: `getChanged()` filters entity changes by user permissions
- Push sync: `update()` validates write permissions per note
- Permission filtering happens post-sync retrieval, not in database
- Each user receives only notes they own or have access to
- No credentials synced (stays local per instance)

#### 6. Database Schema (Migration v234)
```sql
-- New tables
users                    -- User accounts with authentication
groups                   -- User groups  
group_members           -- Group membership
note_ownership          -- Note owner tracking
note_permissions        -- Granular access control (read/write/admin)
```

### API Endpoints (20+ routes)

**User Management:**
- `POST /api/users` - Create user (admin only)
- `GET /api/users` - List users
- `PUT /api/users/:id` - Update user (admin only)
- `DELETE /api/users/:id` - Delete user (admin only)

**Group Management:**
- `POST /api/groups` - Create group
- `GET /api/groups` - List groups
- `PUT /api/groups/:id` - Update group
- `DELETE /api/groups/:id` - Delete group
- `POST /api/groups/:id/members` - Add user to group
- `DELETE /api/groups/:id/members/:userId` - Remove user from group
- `GET /api/groups/:id/members` - List group members

**Permission Management:**
- `POST /api/notes/:noteId/share` - Share note with user/group
- `GET /api/notes/:noteId/permissions` - List note permissions
- `DELETE /api/notes/:noteId/permissions/:id` - Revoke permission
- `GET /api/notes/accessible` - List notes user can access
- `GET /api/notes/:noteId/my-permission` - Check user's permission on note
- `POST /api/notes/:noteId/transfer-ownership` - Transfer note ownership

### Authentication Flow
1. All logins require username + password (no password-only fallback)
2. Username is the user identifier throughout the system
3. Username maps to userId in database for permission/ownership checks
4. Session stores userId + username for all subsequent requests
5. All protected routes use `req.session.userId` for permission checks
6. First-time setup: creates default admin user (admin/admin) for initial access
7. Admin creates additional users with their own username+password combinations

---

## What's Missing (Frontend Layer)

### 1. Web Admin Dashboard
**Status:** Not implemented (0%)

The backend APIs exist but there is no UI for them. Users cannot:
- Create new users
- Manage groups
- Assign permissions
- View permission settings
- Transfer note ownership

**Implementation Required:**
- React component (`components/admin/UserManagement.tsx`) with:
  - User list table with create/delete buttons
  - Modal for user creation (username, password, email, role)
  - Modal for editing user roles/status
  - Group management interface
  - Permission sharing interface
  - Bulk permission management

**Integration Points:**
- Add admin panel link to main navigation (only for admin users)
- Call existing API endpoints: `/api/users`, `/api/groups`, `/api/notes/:noteId/share`
- Handle 401/403 responses for permission denials

### 2. Login Form Updates
**Status:** Partially implemented

Login page must enforce username+password for all scenarios (no password-only fallback):
- Username field is always required and visible
- Password field is always required
- Client-side validation ensures both fields present
- Server-side auth validates against users table (no legacy password mode)
- Redirect unauthenticated users to login page
- First-time setup: direct to initial admin account creation

**Current State:**
- Backend has username field support in multi-user mode
- Backward compatibility code for password-only login still exists (remove this)
- EJS template needs to always show username field with proper styling
- Session management updated to store both userId and username

**Changes Needed:**
1. Remove password-only authentication fallback from `login.ts`
2. Make username field always required in login form
3. Update error messages to specify username/password mismatch
4. Add first-time setup flow: create initial admin user if no users exist

### 3. Note Sharing UI
**Status:** Not implemented (0%)

Current workflow requires API calls manually. Need UI components for:
- Share button on note header
- Modal to select user/group for sharing
- Permission level selector (read/write/admin)
- List of current note permissions
- Revoke permission button per entry
- Transfer ownership option

**Integration Points:**
- Call `/api/notes/:noteId/share` endpoint
- Call `/api/notes/:noteId/permissions` to list current sharing
- Call `/api/users` and `/api/groups` to populate share recipient lists

### 4. User Indicator / Session UI
**Status:** Missing entirely (0%)

Need to show:
- Currently logged-in user in application header
- User profile dropdown with logout
- Role indicator (admin vs user)
- Option to change password
- Session timeout warnings (if configured)

**Integration Points:**
- Call `/api/app-info` or new `/api/auth/me` endpoint to get current user
- Show in top-right corner of application

### 5. Desktop App Updates
**Status:** Not implemented (0%)

Electron desktop application needs to support username-based login:
- Login dialog asks for username + password (not just password)
- Credential storage per user by username (not global password)
- Store encrypted credentials locally per user profile in config file
- Add user switcher: logout current user, switch to different username
- Remember last logged-in user per device (optional convenience feature)

**Current State:**
- Desktop uses single-user architecture with global password only
- No concept of switching between usernames
- Password stored globally, not per username/user

**Required Changes:**
1. Update login dialog to require username field
2. Store credentials per username in secure config (encrypted)
3. Add logout + user switch menu option in main window
4. Update auto-login logic to use stored username+password per user
5. Allow switching to different user from within app (logout → different username)

---

## Architecture Decisions

### Simplified Authentication: Username as User Identifier

The implementation uses a straightforward username+password authentication model:

**Model:**
- Every user has a unique username and password
- Username is the primary identifier (human-readable, memorable)
- All logins require both username and password (no exceptions, no fallbacks)
- userId in database maps to username for permissions and ownership
- First-time setup creates a default admin user (admin/admin)
- Admin creates additional users by specifying their username and initial password

**Benefits:**
- No complexity switching between password-only and username+password modes
- Consistent authentication across all platforms (web, desktop, mobile)
- Username serves as natural team/group identifier for collaboration
- Team generation builds on top of this foundation: "users with @domain.com usernames can form teams"
- Simpler mental model: "log in as your username"

**Impact on Sync:**
- Each user identified by username throughout sync process
- Permission filtering uses userId (which maps 1:1 to username)
- No ambiguity about which user is syncing
- Desktop app can remember which user was last logged in

### Why Permission-Aware Filtering, Not Database Isolation?

The implementation uses permission-aware sync filtering rather than separate databases per user:

**Pro:**
- True multi-device sync: Alice's desktop and mobile both sync with server
- Shared notes actually work: Bob gets updates Alice makes, and vice versa
- Simple mental model: everyone works on same database, filtered by permissions
- Matches Google Docs/Notion architecture

**Con:**
- Filtering happens in application memory, not database query
- For users with access to many notes, filtering cost increases
- Requires explicit permission checking on every sync pull


## System Architecture Diagram

```
User A (Desktop)              User B (Web)              User C (Mobile)
     |                              |                           |
     +--------- Sync Protocol -------+----------- Sync Protocol --------+
                                     |
                              Server Database
                              - All notes in one DB
                              - Sync filtering by user
                              - Permission validation
                              
User A sees: Notes A owns + Notes shared with A
User B sees: Notes B owns + Notes shared with B  
User C sees: Notes C owns + Notes shared with C (same as Desktop)
```

All three users can edit shared notes; changes sync back and forth automatically.

---

## Next Steps to Production

1. **Enforce username+password login** - remove password-only fallback from auth
2. **Build first-time setup flow** - create default admin user on fresh install
3. **Update login form** - always show username field with proper validation
4. **Build web admin dashboard** - unblocks user creation (username/password/role)
5. **Add sharing UI** - allows users to collaborate (share notes with usernames)
6. **Add session UI** - shows who's logged in (username) and logout button
7. **Update desktop app** - username+password login with per-user credential storage
8. **Team generation** - build on username foundation (e.g., @domain teams)
9. **Testing & validation** - verify all workflows across platforms
10. **Documentation** - user guides showing username-based workflows

deajan:
@Somoru Sorry for the late reply, week was complicated.
So far, the architecture seems fine to me.

Please bear in mind that the existing TOTP auth step must still work with the username/password scheme.
Also, in the admin UI, there must be that model upon user deletion which allows to delete all notes or make them own by another user. Generally speaking, notes ownership changes UI should exist anyways.

@eliandoran Could you have another view to make sure I didn't miss anything ?

deajan:
@Somoru Any news on this ?

kneepel:
> @Somoru Any news on this ?

This entire PR (including his own replies!) is very likely AI generated, especially considering the initial PR was non-functional.

Evidently the submitter may not have the first hand knowledge necessary to audit his own work.

eliandoran:
Unfortunately, I'm going to disregard this work for now.

The reasoning behind this is:
1. I'm not happy with the heavy use of LLMs, even for interacting in GitHub comments. It is my expectation to be discussing with a human.
2. The initial work was severely lacking, without having any functional elements (e.g. a login screen, sync completely disregarded).

My apologies for not stating this earlier in some kind of policy on the use of AI and issue bounty.

Implementing multi-user support requires deep knowledge of the application, so I would expect it to **come from someone who has already worked on the project before**.

Let me know if you have any further questions or concerns.

PR #7720, opened 2025-11-13
fix markdown headings are rendered with extra character

contributor:
Fixes markdown headings displays 2004 literally:

image


eliandoran (approved):
Thanks for noticing it.

PR #7757, opened 2025-11-15
fix search returns empty results

contributor:


eliandoran:
Thanks, but there should already be a fix in place that I pushed to main: https://github.com/TriliumNext/Trilium/commit/1844a7d66683b27738a219dbac9698e7b1fd7182

The reason why we do a search instead of obtaining all notes is for the List and Grid collections which support pagination. Since we have pagination, we only load the notes on that page instead of loading hundreds of notes. This is generally why we prefer computation on the server (which already has rapid access to all notes) instead of the client.

Issue #7641, opened 2025-11-06
Cloned collection, after sharing, on shared page have generated wrong links for left menu, when you have more than 2 clones.


x1arch:
### Description

Cloned collection, after sharing, on shared page have generated wrong links for left menu, when you have more than 2 clones.

When you select some note from embedded collection, you will redirect to latest cloned (linked) note.

I create three collections, one of them cloned to two anothers:

<img width="709" height="637" alt="Image" src="https://github.com/user-attachments/assets/85477413-75bb-4ec1-b748-7ce955a44963" />

Now share first collection and select some point from cloned collection:
<img width="979" height="409" alt="Image" src="https://github.com/user-attachments/assets/648df6f7-f097-4946-8c6d-32c77b1564cd" />

When I select some link from cloned collection (I was on `Test` collection) in the left menu I see `Test 2` collection, instead of `Test`.
 
<img width="661" height="408" alt="Image" src="https://github.com/user-attachments/assets/5caca510-635e-45be-9e67-8c06bc000c21" />

### TriliumNext Version

0.99.3

### What operating system are you using?

Ubuntu

### What is your setup?

Server access only

### Operating System Version

24.04

x1arch:
@dosu I pulled latest `triliumnext/trilium:main` docker image, which was updated 23 minutes ago, but the problem still there.

```
Homepage:	https://github.com/TriliumNext/Trilium
App version:	0.99.3
DB version:	233
Sync version:	36
Build date:	Nov 6, 2025, 4:05:38 PM
Build revision:	[e28da416ba215790cd506b58a7792a72e1b95b03](https://github.com/TriliumNext/Trilium/commit/e28da416ba215790cd506b58a7792a72e1b95b03)
Data directory:	/home/node/trilium-data
```

Should I build my own image based on #7552 or  `triliumnext/trilium:main` image already should include this fixes?

x1arch:
@dosu I checked, looks like code from the patch exists in the main branch. I deleted old notes and created new three and result are the same. Attached video.

```
trilium-1  |  _____     _ _ _
trilium-1  | |_   _| __(_) (_)_   _ _ __ ___   | \ | | ___ | |_ ___  ___
trilium-1  |   | || '__| | | | | | | '_ ` _ \  |  \| |/ _ \| __/ _ \/ __|
trilium-1  |   | || |  | | | | |_| | | | | | | | |\  | (_) | ||  __/\__ \
[...]

PR #7759, opened 2025-11-16
[7641] Update share path to {parent_id}/{note_id}

x1arch:
1. Fix share parents by the path `share/{share_parent_id}/{share_note}` #7641 
2. Added new attribute `#shareExclude` for exclude note from share totally, implemented for remove scripts from share. 
3. Updated readme, dev section

It fix my problem, will be grateful if someone test it more precisely.

I tested attachments, images all is works like a charm. Old routes `/{note_id}` redirects by default to `/{parent_id}/{note_id}`.

### DIFF
--- a/apps/server/src/share/content_renderer.ts
+++ b/apps/server/src/share/content_renderer.ts
@@ -47,15 +47,14 @@ function addContentAccessQuery(note: SNote | BNote, secondEl?:boolean) {
     return ""
 }
 
-function getSharedSubTreeRoot(note: SNote | BNote | undefined): Subroot {
+export function getSharedSubTreeRoot(note: SNote | BNote | undefined, parentId: string | undefined = undefined): Subroot {
     if (!note || note.noteId === shareRoot.SHARE_ROOT_NOTE_ID) {
         // share root itself is not shared
         return {};
     }
 
-    // every path leads to share root, but which one to choose?
-    // for the sake of simplicity, URLs are not note paths
-    const parentBranch = note.getParentBranches()[0];
+    const parentBranches = note.getParentBranches()
+    const parentBranch = (parentId ? parentBranches.find((pb: SBranch | BBranch) => pb.parentNoteId === parentId) : undefined) || parentBranches[0];
 
     if (note instanceof BNote) {
         return {
@@ -71,7 +70,7 @@ function getSharedSubTreeRoot(note: SNote | BNote | undefined): Subroot {
         };
     }
 
-    return getSharedSubTreeRoot(parentBranch.getParentNote());
+    return getSharedSubTreeRoot(parentBranch.getParentNote(), parentId);
 }
 
 export function renderNoteForExport(note: BNote, parentBranch: BBranch, basePath: string, ancestors: string[]) {
@@ -98,7 +97,7 @@ export function renderNoteForExport(note: BNote, parentBranch: BBranch, basePath
 }
 
 export function renderNoteContent(note: SNote) {
-    const subRoot = getSharedSubTreeRoot(note);
+    const subRoot = getSharedSubTreeRoot(note, note.parentId);
 
     const ancestors: string[] = [];
     let notePointer = note;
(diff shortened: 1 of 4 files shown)


x1arch:
I faced with issue, browser do not sent Authorization header for pictures and attachments. I added cookie based authorization and query based authorization (for browsers with disabled cookies), could be configured by label `#shareContentAccess`, default `cookie`, token expiration timeout could be configured via `#shareAccessTokenTimeout` in seconds, by default 10 minutes. Token lives in the note, I use one token for multiple access, otherwise the token will overrides by every request and users will interfere with each other.

Added two attributes for modify share template, now it could be possible to hide left panel (`#shareTemplateNoLeftPanel`) or bottom prev next navigation (`#shareTemplateNoPrevNext`).

eliandoran:
@x1arch , would it be possible for you to split the functionalities in different PRs?
Changing the share URL format needs thorough consideration since it can break the existing user functionality. I've noticed there is a redirect in place, but for example some people might have strict rewrite rules in the reverse proxy which might prevent them from accessing the share functionality after an upgrade.

x1arch:
> it can break the existing user functionality.

Not sure about that, could you provide real config example which could broke anything? I've add one more section in path can't imagine rewrite rule which could be broken by that. By default it will use the same behavior like now, just take random parent and show it. I tested it with caddy and nginx it works as expected without any modifications.

> would it be possible for you to split the functionalities in different PRs?

Of course I could, but that will mean, the path changed PR never will be merged, right?

> for example some people might have strict rewrite rules

They could update rewrite rules, isn't it?

PS in theory we could add parent id to cookie and allow switch it to path by the label, as it made with access. I don't like this option because it is implicit behavior, but you're the boss.

eliandoran:
> > it can break the existing user functionality.
> 
> Not sure about that, could you provide real config example which could broke anything? I've add one more section in path can't imagine rewrite rule which could be broken by that. By default it will use the same behavior like now, just take random parent and show it. I tested it with caddy and nginx it works as expected without any modifications.

I'll investigate. 

> 
> > would it be possible for you to split the functionalities in different PRs?
> 
> Of course I could, but that will mean, the path changed PR never will be merged, right?

Not necessarily, I'm not against the idea but I think we need to refine it first. Nevertheless, splitting it helps things get merged faster. 

> 
> > for example some people might have strict rewrite rules
> 
> They could update rewrite rules, isn't it?

They sure could, but it's best to avoid breaking changes. We have to balance the benefit of the change versus the impact.

> 
> PS in theory we could add parent id to cookie and allow switch it to path by the label, as it made with access. I don't like this option because it is implicit behavior, but you're the boss.

What about using query parameters to specify the path? That shouldn't break anything, right?

x1arch:
> What about using query parameters to specify the path? That shouldn't break anything, right?

I don't like it, when we use path, we do not need modify template all works as is (I mean the left menu, bottom navigation), with query we need to modify every link. And it's looking ugly, like not from here, it's ok, if user don't see it (as with attachments) but for address line, no sir.

x1arch:
@eliandoran As you ask, I split this PR to two parts, this branch contains path update and based on #7828. This could be merged later after proxy research will be completed.

apps/server/src/share/shaca/entities/content_accessor.ts
eliandoran, line 1: This is part of #7828 and should not be present in this PR.

apps/server/src/share/shaca/entities/snote.ts
eliandoran, line 1: This is part of #7828 and should not be present in this PR.

apps/server/src/share/content_renderer.ts
eliandoran, line 1: This is part of #7828 and should not be present in this PR.

apps/server/src/share/routes.ts
eliandoran, line 1: There are parts of #7828 here that should not be present in this PR.

apps/server/src/share/routes.ts
```diff
@@ -157,14 +180,29 @@ function register(router: Router) {
...
     router.get("/share/:shareId", (req, res) => {
         shacaLoader.ensureLoad();
 
         const { shareId } = req.params;
 
         const note = shaca.aliasToNote[shareId] || shaca.notes[shareId];
+        const parent = getSharedSubTreeRoot(note)
 
-        renderNote(note, req, res);
+        res.redirect(`${parent?.note?.noteId}/${shareId}`)
```
eliandoran, line 205: I think it would be nice here to not force the note ID of the parent, but to also allow the share alias.

So instead of: `http://localhost:8080/share/zxHGqnEn1laK/bar`, it could be `http://localhost:8080/share/foo/bar`.

eliandoran:
Closing as stale.

Issue #5740, opened 2025-05-24
After reopening the last closed tab, navigation in the right pane panel stops working.


SiriusXT:
### Description

First, open a text note and then close it. After that, use "Reopen Last Closed Tab" to reopen the note. When clicking entries in the right pane panel, the note can no longer navigate to the target position, and the "Search in Notes" function also becomes unusable.

I tested this after reverting to the zadam/trilium repository, and the same issue still occurs.

### TriliumNext Version

0.93.0

### What operating system are you using?

Windows

### What is your setup?

Local (no sync)

### Operating System Version

windows11

PR #7804, opened 2025-11-20
fix(editor): clean up editor on noteContext removal

SiriusXT:
Ensure that the editor is properly destroyed when a noteContext is removed. 
This prevents issues like https://github.com/TriliumNext/Trilium/issues/5740 and 
potential memory leaks, as editors associated with removed notes would otherwise persist.

This approach seems effective, but I’m not certain if it’s the best solution.


SiriusXT:
@eliandoran Thank you for the detailed explanation. I tried clearing noteTypesToRender in NoteDetail.tsx, but unfortunately I couldn’t get it to work correctly. It turned out to be more complex than I expected, and since I’m not making progress on this at the moment, I’ll go ahead and close the PR.

PR #7828, opened 2025-11-21
Fix share access to attachments for notes protected by login:password

x1arch:
Split this PR: https://github.com/TriliumNext/Trilium/pull/7759

I faced with issue, browser do not sent Authorization header for pictures and attachments. I added cookie based authorization and query based authorization (for browsers with disabled cookies), could be configured by label

- Added new attribute `#shareExclude` for exclude note from share totally, implemented for remove scripts from share.
- Added new attribute `#shareTemplateNoPrevNext` for hide bottom prev next navigation.
- Added new attribute `#shareTemplateNoLeftPanel` for hide left panel totally.
- Added new attribute `#shareContentAccess` for manually setup content access auth, default is cookie. Bould be basic / query / cookie.
- Added new attribute `#shareAccessTokenTimeout` for manually configure token expiration timeout in seconds. (default value is 10 minutes)
- Updated readme, dev section

Token lives in the note, I use one token for multiple access, otherwise the token will overrides by every request and users will interfere with each other.


eliandoran:
@x1arch , so if I understand correctly the issue is that images and other assets are not rendered in password-protected shares, right?

What's the reason why it's failing, aren't HTTP authorization headers sent again by the browser (cached credentials) for images as well, provided the domain and realm are the same?

x1arch:
Originally it should, but not works. The same problem if you will try to use `https://user:pass@example.com/{img}`, because url `user:pass` not allowed for img and attachments (because unsafe). Earlier I found original discussion about that on chromium tracker, but can't find it today. As solution - use tokens.

document headers
image

img headers
image

apps/server/src/share/content_renderer.ts
```diff
@@ -40,6 +40,13 @@ interface Subroot {
 
 type GetNoteFunction = (id: string) => SNote | BNote | null;
 
+function addContentAccessQuery(note: SNote | BNote, secondEl?:boolean) {
+    if (!(note instanceof BNote) && note.contentAccessor && note.contentAccessor?.type === "query") {
+        return secondEl ? `&cat=${note.contentAccessor.getToken()}` : `?cat=${note.contentAccessor.getToken()}`;
+    }
+    return ""
+}
```
eliandoran, line 48: Why have three different mechanisms in place for content accessors? This creates a maintenance burden that I don't think is justified.
I would say that we choose one and stick to it.
Honestly, I prefer the query parameter one since it explicitly indicates access and perhaps it can even be more useful (think copy and pasting it into another Trilium instance, which will be able to download the image strictly from the URL instead of requiring a cookie).

README.md
```diff
@@ -154,6 +169,10 @@ pnpm install
 pnpm run server:start
 ```
 
+> If you faced with some problems, try to delete all `node_modules` and `.pnpm-store` folders, not only from the root, from every directory, like `apps/{app_name}/node_modules`and `/packages/{package_name}/node_modules` and then reinstall it by the `pnpm install`.
+
+Share styles not compiling by default, if you see share page without styles, make `pnpm run server:build` and then run development server.
```
eliandoran, line 174: It's enough to do `pnpm --filter=share-theme build` instead of `pnpm run server:build` because that one takes a lot longer to execute.

eliandoran (changes requested):
FYI, I'm quite fine with the implementation, apart from the requested changes.

However, in the future please take the time to discuss first the request and perhaps even the implementation. It's much easier for me to review and test something that has already been discussed, and it also decreases the chances of being rejected after the work has already been done.

I would also appreciate it if we keep the PRs minimal. For example, in this PR there's also 5 different new labels that were introduced, with only 2 of them actually relating to the work at hand.

eliandoran:
Closing as stale. From what I see attachments themselves are rendering just fine without any special mechanism in both Firefox and Chromium. Protected images as notes also work.

PR #7841, opened 2025-11-22
Auto execute saved search note

contributor:
If `#autoExecuteSearch` label is present, it auto executes a search note.

If ribbon's tab Collection Properties is present (and search results are not empty), it will auto switch to the tab too.

Context:  https://github.com/TriliumNext/Trilium/pull/7704#issuecomment-3564929495


apps/client/src/widgets/ribbon/SearchDefinitionTab.tsx
```diff
@@ -73,6 +73,35 @@ export default function SearchDefinitionTab({ note, ntxId, hidden }: TabContext)
     }
   });
 
+  const executionState = useMemo(() => {
+    const AUTO_EXEC_KEY = "lastAutoExecutedSearchNote";
+    return {
+      load: () => sessionStorage.getItem(AUTO_EXEC_KEY),
+      save: (noteId: string) => sessionStorage.setItem(AUTO_EXEC_KEY, noteId),
+    };
+  }, []);
```
eliandoran: Trilium doesn't use session storage and I'm not sure why it would need to be used here.
You have `useNoteLabel` which reads the label value for you automatically.
You can use `useState` or `useRef` to store the last executed note, not session storage.

eliandoran (changes requested):

apps/client/src/widgets/ribbon/SearchDefinitionTab.tsx
contributor: @eliandoran this was my path: useRef -> useState -> sessionStorage. The first two didn't work, there are commits for that. The react component is recreated every time and useRef.current is reset. 

In order to keep react component I would have to use `stayInDom` in ribbon definition. This approach is worse than sessionStorage.

eliandoran: Why would it be worse?

contributor: It would fire `useEffect` when you navigate away to other notes (not of search type). There is an earlier return for that in `useEffect`, but still this is not good approach.

contributor: I replaced sessionStorage with module level var, this approach better then both.

apps/client/src/widgets/ribbon/SearchDefinitionTab.tsx
```diff
@@ -73,6 +73,35 @@ export default function SearchDefinitionTab({ note, ntxId, hidden }: TabContext)
...
+
+        await refreshResults();
+
+        if (noteContext?.viewScope?.viewMode === "default") {
+          parentComponent?.triggerCommand("toggleRibbonTabBookProperties", {});
+        }
+      }
+    }
+    autoExecute();
+  }, [note]);
```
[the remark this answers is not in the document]
contributor: @gemini-code-assist only note being changed should be a trigger to execute search. No need to trigger it more often than that.

eliandoran (changes requested):
It's fine to have an attribute to execute the auto-search but I think doing it from the `SearchDefinitionTab` is not the best place to do so since it's a ribbon tab (e.g. user configuration).

Have you noticed that if you press the Search button in the launch bar, the search is immediately executed? 

I tracked down this behavior, you can find it in `root_command_executor.ts` and looking for `searchNotesCommand`. Notice how it forces the search via `loadSearchNote`.

We can do the same, but you'd have to find a good place to do so. Perhaps in `search_result.tsx` where the logic of search not executed yet is also determine.

P.S. Don't forget to update the documentation; new features cannot be undocumented.

contributor:
@eliandoran 

> It's fine to have an attribute to execute the auto-search but I think doing it from the SearchDefinitionTab is not the best place to do so since it's a ribbon tab (e.g. user configuration).

`autoExecute` does ultimately call `loadSearchNote`

image

It does this by using call to existing `refreshResults` in the same file (`SearchDefinitionTab.tsx`). So it works not much differently from “manual” search.

I could extract `autoExecute` function to some file, but then `refreshResults` should be extracted too. What are advantages of that?

contributor:
@eliandoran 

By the way, there is a “double list” bug in display search results. I first attributed this to the auto-search. But no, this is unrelated, and exists in main.

- start/reload app on a text note
- go to saved search note 
- press search button (you can do this few times with different limits)

Search results are duplicated. When there are many results, you have double scrollbars too (moving in opposite directions)

Reproducable both in current main HEAD 71b86b3cbc, and in main's commit 435b856b72 this PR starts from

<details><summary>Screenshots</summary>
<p>

image

image

</p>
</details>

eliandoran:
@contributor , can you merge with the latest `main` to see if the issue is still reproducible? On my side I don't get double results.

Edit: I see you mentioned it occurs in the latest main as well, weird.

contributor:
@eliandoran 

> On my side I don't get double results.

Set these attributes on root note:
`#viewType(inheritable)=list #windowName=ShouldNeverHappen #iconClass="bx bx-home-alt-2"`

eliandoran:
@contributor , apparently `#viewType(inheritable)=list` is the key.

Should be fixed in d7ae2e4307a27b739664ee83507cc9a6f3c0d4e8.

contributor:
> @contributor , apparently `#viewType(inheritable)=list` is the key.
> 
> Should be fixed in [d7ae2e4](https://github.com/TriliumNext/Trilium/commit/d7ae2e4307a27b739664ee83507cc9a6f3c0d4e8).

@eliandoran 

Still doubled result. Both in search tab and collection tab

contributor:
@eliandoran

Additionally, there is another, more severe bug (including current main 3afe6df9f26ae). Collection and text notes stops to show their children notes (root has `#viewType(inheritable)=list`)

- open trilium on a text note from collection
- click on collection note - it is empty
- after that clicking on any collection or text note won't show their children

<details><summary>Details</summary>
<p>

Collection with children:
image

Text note with children:
image


</p>
</details>

eliandoran:
> > @contributor , apparently `#viewType(inheritable)=list` is the key.
> > Should be fixed in [d7ae2e4](https://github.com/TriliumNext/Trilium/commit/d7ae2e4307a27b739664ee83507cc9a6f3c0d4e8).
> 
> @eliandoran
> 
> Still doubled result. Both in search tab and collection tab

Can't reproduce this issue anymore, are you sure it's not a problem with the cache on your side?

> Additionally, there is another, more severe bug (including current main https://github.com/TriliumNext/Trilium/commit/3afe6df9f26aed727b307bf1916d83752f3ef95f). Collection and text notes stops to show their children notes (root has #viewType(inheritable)=list)

Nice catch. I was aware of the bug but did not have specific steps to reproduce it.
Should be fixed by 976b1e1e0fe0176606abb9f6c83dcefa16391d17.

contributor:
@eliandoran 

> Can't reproduce this issue anymore, are you sure it's not a problem with the cache on your side?

I'm sure. Double collection bug is still there.

Empty collection bug is now fixed.

eliandoran:
@contributor , what commit are you on?

contributor:
> @contributor , what commit are you on?

976b1e1e0fe0176606abb9f6c83dcefa16391d17

eliandoran:
@contributor , can't reproduce it...

https://github.com/user-attachments/assets/3f7feee9-1d85-41b2-bad9-09b23bcf04e3

contributor:
@eliandoran 

I retested again. Up until commit 976b1e1e0fe0176 double list is present. The commit fixes empty list. And apparently, this also fixes double list.

rom1dep:
> If #autoExecuteSearch label is present, it auto executes a search note.

what does that mean, practically?

contributor:
> > If #autoExecuteSearch label is present, it auto executes a search note.
> 
> what does that mean, practically?

If you have a saved search note, `#autoExecuteSearch` will executes search automatically when you click on a tree node.

I had an old abandoned "Saved searches" branch  with various search notes. I added `#autoExecuteSearch(inheritable)` to it, and the search notes now are easy and fun to click on. I moved the most useful search notes to other parts of tree/projects (and added `#autoExecuteSearch` to them directly).

contributor:
@eliandoran added docs. ready for review.

rom1dep:
> If you have a saved search note, `#autoExecuteSearch` will executes search automatically when you click on a tree node.

Specifically, which tree node relative to the search note needs to be interacted with for the search to be triggered?

As far as I can tell, "click to search" is a capability that already exists (via the refresh button, on the search note itself):

image

It'd be good that:
- we don't introduce conflicting ways to do the same thing (click to refresh)
- the new way of implementing "click to refresh" is well documented and at least as discoverable as the current one (assuming a new way is warranted).

Now, if the intent is to have "smart" search-note hierarchies that are kept up to date and rendered at all time (without needing explicit refresh, which is what `#autoExecuteSearch` kind of hints at), I suppose it's a much bigger effort.

contributor:
> Specifically, which tree node relative to the search note needs to be interacted with for the search to be triggered?

Search note itself

> Now, if the intent is to have "smart" search-note hierarchies that are kept up to date and rendered at all time (without needing explicit refresh, which is what `#autoExecuteSearch` kind of hints at), I suppose it's a much bigger effort

Yes, this is what it does. Browsing saved searches as collections with one click on a tree node.

`#autoExecuteSearch` can be on a search note directly, or inherited from any note up in hierarchy

rom1dep:
> > Specifically, which tree node relative to the search note needs to be interacted with for the search to be triggered?
> 
> Search note itself

If the search note has to be clicked on, are you sure that we need this change? By your account, this is effectively the same as clicking on the refresh button which we already have (see screenshot above), on that same note. If the need is satisfied by a larger refresh icon, let's explore a UI solution that won't pull new attributes, configuration and complexity.

> > Now, if the intent is to have "smart" search-note hierarchies that are kept up to date and rendered at all time (without needing explicit refresh, which is what `#autoExecuteSearch` kind of hints at), I suppose it's a much bigger effort
> 
> Yes, this is what it does.

Then this contradicts what you just wrote. Which one is it?
- Are the search results guaranteed to always be up to date (reacting to notes being added/modified/deleted in the background)? If such a thing is even possible, I would be keen on making it the default, but I don't believe that this PR does this.
- Are the search result fetched on-demand from explicit user-interaction?

> Browsing saved searches as collections with one click on a tree node.

Then let's make sure that, when the collection is pegged to a search note:
- its results are fetched/refreshed automatically when the collection is first rendered (I'm surprised because this seems to be the current behaviour here?)
- that there is an easily accessible mean for the user to refresh the collection would the need arise
  -  We already have on button in the note-tree, on the collection=search-note itself (see screenshot above)
  -  We also already have the "Search" button, under "Search Parameters"
  - … do we really need more than that?

contributor:
The attribute makes new functionality opt-in.

With `#autoExecuteSearch`:
- one click on a tree node to view result collection, second click to view next node and so on

Without it:
- three clicks (while drawing a triangle with mouse cursor) to view result collection, then three clicks and triangle to view the next and so on

contributor:
Rebased and retested. Here is the status

`autoExecute` is just wrapper over existing `refreshResults`

```
      function autoExecute() {
         if (env.SAFE_MODE) return 
         if (note is not #autoExecuteSearch) return
         if (the same search was recently executed) return
         		
     	 await refreshResults();
     }
```
Moving it to another module would complicate the logic

Condition `if (the same search was recently executed) return` is optional and implemented via helper
```
    const executionState = function () {
        let lastAutoExecutedSearchNoteId = "";
        return {
            load: () => lastAutoExecutedSearchNoteId,
            save: (noteId: string) => lastAutoExecutedSearchNoteId = noteId,
        };
    }();
```
`executionState` can be removed at all - it just prevents re-executing the same search note when you return to that note. It won't prevent that if you switch between different search notes (solution with `sessionStorage` from the first commits could work for that case).


Additionally, auto-execute is no longer runs in SAFE_MODE. This is a safety measure, as there are searches that just hangs 0.102.1 completely. When you restart trilium, it opens the same search note and hangs again. Without SAFE_MODE it would be impossible to change active note.

PR #7842, opened 2025-11-22
Feat(math): Improve legacy math input with MathLive

Meinzzzz:
Description :
Integrates a MathLive field to provide a richer math-editing experience, offering both a visual equation editor and a raw LaTeX code editor.  
Key Features  
Visual equation editing using MathLive, while preserving an area for directly editing raw LaTeX.  
Technical Implementation:

*   MathLiveInputView: A new wrapper class for the MathLive web component that bridges it with CKEditor 5's view architecture.
*   RawLatexInputView: A specialized view for the raw LaTeX textarea.
*   MainFormView Refactor: 
    *   Orchestrates the communication between the two inputs.
    *   Manages the ResizeObserver lifecycle to prevent memory leaks.
    *   Handles focus cycling and keyboard navigation.
*   Styles: Comprehensive CSS updates in mathform.css to support the new layout, including resize: both support and flexbox layouts for the dual inputs.

Testing:Unit Tests: Added comprehensive test suites for MathUI, covering:

*   Initialization and teardown.
*   Command execution and binding.
*   Synchronization logic between MathLive and Raw LaTeX inputs.
*   Keyboard shortcuts and focus management.

Dependencies:

*   Added mathlive package
image


eliandoran (changes requested):
I've tried testing it but it fails with:

```
Uncaught (in promise) TypeError: can't access property "bind", formView.mathLiveInputView is undefined
    _createFormView mathui.ts:92
    init mathui.ts:29
    promise callback*init/initPlugins/< chunk-CZONVLGZ.js:45733
    initPlugins chunk-CZONVLGZ.js:45726
    init chunk-CZONVLGZ.js:45614
    initPlugins chunk-CZONVLGZ.js:47131
    create ckeditor5.js:9558
    create ckeditor5.js:9553
    buildEditor CKEditorWithWatchdog.tsx:222
    CKEditorWithWatchdog CKEditorWithWatchdog.tsx:158
```

packages/ckeditor5-math/src/ui/mainformview.ts
```diff
@@ -1,30 +1,24 @@
-import { ButtonView, createLabeledTextarea, FocusCycler, LabelView, LabeledFieldView, submitHandler, SwitchButtonView, View, ViewCollection, type TextareaView, type FocusableView, Locale, FocusTracker, KeystrokeHandler } from 'ckeditor5';
-import IconCheck from "@ckeditor/ckeditor5-icons/theme/icons/check.svg?raw";
-import IconCancel from "@ckeditor/ckeditor5-icons/theme/icons/cancel.svg?raw";
+import { ButtonView, FocusCycler, LabelView, submitHandler, SwitchButtonView, View, ViewCollection, type FocusableView, Locale, FocusTracker, KeystrokeHandler } from 'ckeditor5';
+import IconCheck from '../../theme/icons/check.svg?raw';
+import IconCancel from '../../theme/icons/cancel.svg?raw';
```
eliandoran: What's the reason of providing our own icons? I believe the original icons were just fine.

There's also an issue with the current check, it looks off.
Image

packages/ckeditor5-math/src/ui/mainformview.ts
```diff
@@ -1,30 +1,24 @@
...
-	public lazyLoad: undefined | ( () => Promise<void> );
+	public declare saveButtonView: ButtonView;
+	public declare mathLiveInputView: MathLiveInputView;
+	public declare rawLatexInputView: RawLatexInputView;
+	public declare rawLatexLabel: LabelView;
+	public declare displayButtonView: SwitchButtonView;
+	public declare cancelButtonView: ButtonView;
+	public declare previewEnabled: boolean;
+	public declare previewLabel?: LabelView;
+	public declare mathView?: MathView;
```
eliandoran: What's the reason of using the `declare` keyboard here?

packages/ckeditor5-math/src/ui/mainformview.ts
```diff
@@ -172,44 +201,149 @@ export default class MainFormView extends View {
...
+						? this.rawLatexInputView.element
+						: this.mathLiveInputView.element;
+
+					if ( other && other.style.width !== width ) {
+						window.requestAnimationFrame( () => {
+							other.style.width = width;
+						} );
+					}
 				}
+			}
```
eliandoran: This whole resize mechanism feels overengineered. Do we actually need it? I tried remove part of it and it worked just fine.

I would prefer not to have this logic, maybe see if we can work around it using CSS since the popup follows the biggest text area, I think it should be fine without it.

.gitignore
eliandoran, line 1: This change is probably not needed.

packages/ckeditor5-math/package.json
```diff
@@ -71,6 +71,6 @@
     ]
   },
   "dependencies": {
-    "@ckeditor/ckeditor5-icons": "47.2.0"
```
eliandoran: This change is probably not needed.

eliandoran:
Functionally speaking, it's pretty interesting since it makes it relatively easy for people not familiar with LaTeX to insert equations.
Before this can be merged there are two main aspects to be handled:
1. The complexity of this PR is quite high, I believe not all of the modifications are necessary and the same can be achieved with a more minimal set of changes.
2. We need to make sure that MathLive doesn't do any "calling home" and that it works completely offline and self-hosted within Trilium. There's also the downside of the library size, since from what I see from `npm` it's at least 5 MB big.

eliandoran:
@Meinzzzz , I see you've marked the PR as ready but no changes were pushed. Perhaps you forgot to push your changes?

Meinzzzz:
Sorry, i hit the revie button by accident.
To address the library size, it should be possible to download the minified production bundle directly from unpkg (~860 KB) instead of using the full npm package (~6 MB). Unpkg can provide all the necessary fonts and static CSS files needed for offline use. 
For the offline/self-hosted requirement, if we download everything from unpkg (JS, CSS, fonts which is about ~20KB) and use them locally in Trilium. Then configure MathLive to use these local paths. At the moment i use a java script note to add a button with mathlive, so the approach with unpkg could work, but i am not sure how to handel mathlive updates with that.

eliandoran:
> Sorry, i hit the revie button by accident. To address the library size, it should be possible to download the minified production bundle directly from unpkg (~860 KB) instead of using the full npm package (~6 MB). Unpkg can provide all the necessary fonts and static CSS files needed for offline use. For the offline/self-hosted requirement, if we download everything from unpkg (JS, CSS, fonts which is about ~20KB) and use them locally in Trilium. Then configure MathLive to use these local paths. At the moment i use a java script note to add a button with mathlive, so the approach with unpkg could work, but i am not sure how to handel mathlive updates with that.

Let's keep the size constraints for later after we run a nightly build on it to determine the impact of the changes.

packages/ckeditor5-math/src/ui/mainformview.ts
Meinzzzz: Your right, i just forgot to change back to the dafault ones. They were only used when i was working on the styling

packages/ckeditor5-math/src/ui/mainformview.ts
Meinzzzz: I think I found a good solution that works using only CSS and also feels nice to use. Instead of making both boxes resizable, only the LaTeX box remains resizable. The Mathlive component can then grow automatically depending on the input.

packages/ckeditor5-math/src/ui/mainformview.ts
Meinzzzz: There isn’t a specific reason for using declare here. My TypeScript knowledge is still needs improvement, so I added it unnecessarily. Thanks for pointing that out, I will remove it.

Meinzzzz:
@eliandoran I am still working on it, but I’d appreciate a rough pass on the logic when you have a moment. I removed the unnecessary logic for resizing and tried to move everything into the CSS. I want to make sure the structure looks right to you before I continue working on it.

eliandoran:
> @eliandoran I am still working on it, but I’d appreciate a rough pass on the logic when you have a moment. I removed the unnecessary logic for resizing and tried to move everything into the CSS. I want to make sure the structure looks right to you before I continue working on it.

Sure. I'll have a look after delivering 0.100.0.

Meinzzzz:
@eliandoran I fixed the math input issues. The text size can be adjusted in the CSS.
For the virtual keyboard, I changed it so that when you click the toggle, you automatically start writing in the MathField. This means if you are in the LaTeX area, focus switches to the MathField. I did this because it is difficult to map the output of the virtual keyboard to the correct position within the LaTeX textarea.

I believe everything is ready now. Please let me know if I should improve anything else!

eliandoran (approved):

PR #7877, opened 2025-11-28
Resolve focus issues within split pane

SiriusXT:
1. [fix(empty): open note in the correct split pane](https://github.com/TriliumNext/Trilium/commit/210dcfb98969a9bdda26073c89b7bac352404e3d)
Fix issue: After opening two splits, when focusing the first split and then clicking an item in the second empty split, Trilium opens the note in the first split instead of the second.

preview

2. [fix(split): only trigger focusOnDetail when necessary](https://github.com/TriliumNext/Trilium/commit/407cac588a3e60bb185d249c5f9dd2859e7247ac)
Fix issue: After opening two splits, when focusing the second split and then clicking a note title in the first split, the focus fails.
preview

3. [fix(left-pane): only focus the note when toggling left pane visibilit…](https://github.com/TriliumNext/Trilium/commit/648ab4d736061b49c5033c0392070971f2f22986)
Fix issue: When the focus is on a note title, toggling the left pane visibility with a shortcut causes the focus to shift from the note titleto the note detail.

### DIFF
--- a/apps/client/src/menus/custom-items/NoteColorPicker.tsx
+++ b/apps/client/src/menus/custom-items/NoteColorPicker.tsx
@@ -8,6 +8,7 @@ import Color, { ColorInstance } from "color";
 import Debouncer from "../../utils/debouncer";
 import FNote from "../../entities/fnote";
 import froca from "../../services/froca";
+import { isMobile } from "../../services/utils";
 
 const COLOR_PALETTE = [
     "#e64d4d", "#e6994d", "#e5e64d", "#99e64d", "#4de64d", "#4de699",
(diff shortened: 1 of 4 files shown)


apps/client/src/widgets/type_widgets/Empty.tsx
```diff
@@ -45,10 +45,10 @@ function NoteSearch() {
                         if (!suggestion?.notePath) {
                             return false;
                         }
-
-                        const activeContext = appContext.tabManager.getActiveContext();
-                        if (activeContext) {
-                            activeContext.setNote(suggestion.notePath);
+                        const ntxId = autocompleteRef.current?.closest(".note-split")?.getAttribute("data-ntx-id") ?? null;
```
eliandoran: Instead of querying the `ntxId` from the DOM, it's better to do it at React level.
You can obtain the `ntxId` from `Empty` from `TypeWidgetProps`, you can pass it as an argument to `NoteSearch`.

eliandoran (approved):

PR #7878, opened 2025-11-28
Fix sorting

rom1dep:
- Ensure that "folders first" ordering mode doesn't place folder notes on top of notes flagged as `#top`
- Improve tests by documenting the new behaviour and increasing coverage for existing sorting functions

### DIFF
--- a/apps/server/src/services/tree.spec.ts
+++ b/apps/server/src/services/tree.spec.ts
@@ -1,11 +1,11 @@
-import { beforeEach, describe, expect, it, vi } from "vitest";
-import { note, NoteBuilder } from "../test/becca_mocking.js";
+import {beforeEach, describe, expect, it, vi} from "vitest";
+import {note, NoteBuilder} from "../test/becca_mocking.js";
 import becca from "../becca/becca.js";
 import BBranch from "../becca/entities/bbranch.js";
 import BNote from "../becca/entities/bnote.js";
 import tree from "./tree.js";
 import cls from "./cls.js";
-import { buildNote } from "../test/becca_easy_mocking.js";
+import {buildNote} from "../test/becca_easy_mocking.js";
 
 describe("Tree", () => {
     let rootNote!: NoteBuilder;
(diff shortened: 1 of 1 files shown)


rom1dep:
> I've found one issue with the `sortFoldersFirst` logic when combined with reverse sorting, which causes folders to be sorted last.

It is kept this way in case the user would wish, for any reason, sort folders last.

eliandoran:
> > I've found one issue with the `sortFoldersFirst` logic when combined with reverse sorting, which causes folders to be sorted last.
> 
> It is kept this way in case the user would wish, for any reason, sort folders last.

@rom1dep ,
Are you sure it's the best approach? We recently changed #top to appear first even when the order is descending. I would prefer if we maintain consistency.

rom1dep:
> @rom1dep , Are you sure it's the best approach? We recently changed #top to appear first even when the order is descending. I would prefer if we maintain consistency.

@eliandoran this PR is consistent with the current behaviour in that `#top` remains on top, no matter what. I amended a couple of test-cases to make sure we don't inadvertently regress on it in the future.

Stepping back a bit, I certainly don't like the syntax of it all, but I also don't think that this PR makes things worse. My preference, if I had a shot at a clean slate design, would be to have `#sortFoldersOrder=first|last` combined with `#sortFoldersDirection=asc|desc` (instead of `#sortFoldersFirst`), but here we are.

If we were to migrate to something like that, I am not sure what the resulting PR would look like (I suppose that we would have to bump the data model version and run a post-update full-document migration, and I don't trust my abilities to get that right).

Point taken about the documentation, though. Once we reach consensus on the above, I will get on to it and clarify how the different sorting modes interact with one another (which, tbf, should have been done a long time ago).

eliandoran:
@rom1dep ,

Let's reach consensus.

Here's how I see it:

* `#top` keeps the items at the top, regardless of sort direction. As mentioned previously, we had a bug where `top` items ended up at the end of the list if sort direction was reversed.
* I would want `#sortFoldersFirst` to act exactly like `#top` (arguably first and top are slightly synonymous). Perhaps we can rename it if needed.
* I wouldn't go for changing the existing attributes since at some point they'll become system attributes and mostly be used through the UI. Only "hard-core" users would still care about these attributes, case in which they could tolerate a bit of mess in naming.

PR #7879, opened 2025-11-28
(feat) Gallery Note Type

lzinga:
# What it does
It adds a `Gallery` note type that you can upload images into and shows the images in a masonry list. It also has the ability to show all pictures in child notes which allows you to group images together too if needed. You can either use the attribute `#hideChildAttributes` or click the "Show/Hide Child Attributes" button (which just adds the attribute).

My main goal of this is the ability to use a note as a place to put images that don't necessarily need to be attached to anything specific and view images that may exist within a Trilium instance and be able to create a gallery of images you want to store/keep. It currently saves in the database like all other attachments, however if files being saved in a directory eventually comes this should be easily updated as it uses the built in attachment upload/save methods.

All are images in the pics below, so the mindmaps and stuff are images pulled off the Trilium site.

image

When hovering over an image it will show a delete icon and prompt confirmation
image

On images that are contained within child notes/galleries it will add a go to option which will take you directly to the note and image making it easier to know where it is.
image

When clicking the image itself it will open a lightbox for said image that will allow you to transition between the images as well.
image

Lastly, if the gallery is marked as shared the attachments will be sharable and you can click the button to get the url in the clipboard.
image

# Tests
There is a few tests and typecheck failures but it doesn't __appear__ to be directly related to any of the changes made in this pull request from what I can see.


rom1dep:
Pretty cool!

eliandoran (commented):
Conceptually, it's pretty interesting.

Before I do a deep dive into the code, there are some important architectural aspects to consider if we are to integrate this:

 * I find it interesting that the gallery is a new note type. And it works kind of in a hybrid way: it supports both attachments (which make sense in the context of a note type), but it also supports child notes.
 * Since it supports child notes, I think it would make more sense to implement it as a [Collection](https://docs.triliumnotes.org/user-guide/collections) instead of a new note type.
 * We also had in the past a similar case for geomaps, which we implemented as a standalone note type and we then ported it to the collections mechanism. I would prefer we make a choice from the beginning and stick to it, since migration can be annoying.
 * Another benefit of using collections is the fact that you can also integrate it into search notes, so being able to search across the hierarchy.
 * We also have to decide if we keep support for attachments or only child notes. Although I appreciate the flexibility, I think it adds a bit of complexity to both maintenance (2 different branches, kind of like the calendar), but also on the UX side.
 * Of note, creating a new note type also implies increasing the sync version, because a sync with an old server (pretty common case) will transform the gallery into a File note type otherwise.
 * I like the lightbox when you click on a photo. I think that it should be extracted and made reusable, so that it works for example to open images/attachments from Text notes as well.

PS. For big features such as this, I would appreciate it if we can have a discussion about it beforehand. The reasoning is that you would avoid extra post-development work to align it from an architectural point of view.

CC @rom1dep , in case you would like to pitch in. :)

lzinga:
You're right, we can close this one for now and we can move the discussion to a feature request which I have opened - https://github.com/TriliumNext/Trilium/issues/7895

Issue #4848, opened 2024-08-15
(Feature request) format brush


guyon168:
### Describe feature

I am a regular user of Trilium Notes and I greatly appreciate the functionality it offers. I would like to confirm whether the current rich text editor used in Trilium is CKEditor 5. Additionally, I would like to suggest the inclusion of a "format brush" feature in future versions. This feature would allow users to copy text formatting from one section and apply it to another, which could significantly enhance the editing experience.

Is it possible to consider adding this feature in upcoming releases? Your feedback and any guidance on this would be greatly appreciated.

Thank you for your continued development and support!

Best regards,

### Additional Information

format brush

eliandoran:
@Guyon168 , we are indeed using CKEditor 5.

Unfortunately, the "Format painter" feature is a premium feature: https://ckeditor.com/docs/ckeditor5/latest/features/format-painter.html

We are currently only using the community edition. We are looking into adding premium features, but we'd have to discuss with the CKEditor devs.

georgejung:
It would be a nice feature to have for sure.

SiriusXT:
Hi @eliandoran, I noticed that the official Format Painter implementation doesn’t include block styles and doesn’t support double-click for continuous application. Do we want to implement this ourselves?

I saw that this issue is assigned to you. I’m interested in contributing to this feature. If you haven’t started yet, would it be okay if I give it a try?

eliandoran:
@SiriusXT , feel free to take it. :)

> Do we want to implement this ourselves?

For now I would say to just use the original implementation. We can explore improving it later.

PR #7902, opened 2025-11-30
feat(ckeditor5): add formatPainter

SiriusXT:
Closes https://github.com/TriliumNext/Trilium/issues/4848
图片

Use the Format Painter feature from CKEditor5 Premium.

---
Features that can be explored in the future:

1. Enable continuous painting by double‑clicking the Format Painter icon.
 2. Support for block‑level styling.
3. CKEditor5’s Format Painter applies styles to words using spaces or other Western punctuation as boundaries (i.e., by directly clicking to paste the format), which prevents correct handling of Chinese or other languages that do not use space-based word segmentation.


eliandoran (changes requested):
Seems fine, but we need to update the documentation:
* Add a new entry in https://docs.triliumnotes.org/user-guide/note-types/text/premium-features.
* Add a reference to the newly created entry here, under a heading near "Remove formatting": https://docs.triliumnotes.org/user-guide/note-types/text/general-formatting
* Add a reference to the new premium feature here: https://docs.triliumnotes.org/user-guide/note-types/text

SiriusXT:
> Seems fine, but we need to update the documentation:

@eliandoran I’m not sure whether these are the correct documentation files to submit. After running pnpm edit-docs:edit-docs, VS Code generated a bunch of files, and I committed the relevant ones 😅.

PR #7965, opened 2025-12-06
Fix (tooltip): tooltip does not disappear after click

SiriusXT:
Fix tooltip persistence issue: tooltip now correctly hides after click in the following scenarios:

1. Text options
图片

2.Code options
图片

3.Fullscreen button in the global menu
图片

4.Action buttons
图片


eliandoran:
@SiriusXT , @gemini's review is correct as I had to deal recently with the same issue for the back/next buttons in the launcher bar (#7934).

The tooltip must still show up if you use keyboard-based navigation (using tab), so changing the trigger is not going to work in this case. There are two approaches i can think of: manually dismissing the tooltip after being clicked, or removing focus after click.

SiriusXT:
> @SiriusXT , @gemini's review is correct as I had to deal recently with the same issue for the back/next buttons in the launcher bar (#7934).
> 
> The tooltip must still show up if you use keyboard-based navigation (using tab), so changing the trigger is not going to work in this case. There are two approaches i can think of: manually dismissing the tooltip after being clicked, or removing focus after click.

@eliandoran  I tested manually dismissing the tooltip on click, but since the mouse remains over the item, the tooltip still reappears. I also tried removing focus after click, but that breaks keyboard navigation since Tab can no longer start from the current element. For now, let’s just put this issue aside.

Issue #5645, opened 2025-04-10
Copy as Markdown


FalcoGer:
### Describe feature

Sometimes you just quickly want to share your notes with someone. But simply copy and pasting the whole note text will destroy most of the formatting.

The current solution is to export the note to a file, and then either copy that file's contents or send the whole file. But often you don't need that file afterwards and have to manually delete it.

The solution: Export the note directly to the clipboard. I don't think in these cases anything other than markdown is required, but it may be worth considering if someone might want a zip file or html in their clipboard for some reason.

### Additional Information

N/A

FliegendeWurst:
> But simply copy and pasting the whole note text will destroy most of the formatting.

In my experience it doesn't, at least when pasting into e.g. LibreOffice. Almost all HTML formatting is preseverd. Works for both web and electron.

Copy as Markdown would still be useful though.

PR #8009, opened 2025-12-10
feat(dialog/export): export to clipboard

yzx9:
This pull request adds a new "Export to clipboard" feature to the note export dialog, allowing users to directly copy the content of a single note to their clipboard instead of downloading it as a file. The changes include UI updates, translation additions, and logic to handle clipboard export, along with improvements to the export dialog's usability.

This feature was originally requested in #5645, where it was later changed to “Copy as Markdown.” I’m also interested in adding that feature, but since I’m new to Trilium, I thought implementing this one would be a good starting point as it’s simpler and still potentially useful.

Please feel free to point out any issues or conventions I might have missed.


eliandoran:
Hi,

Thank you for your contribution, but I will not be able to merge it since it doesn't align with the comment in https://github.com/TriliumNext/Trilium/issues/5645#issuecomment-2993581068.

I would very much prefer a copy as Markdown context menu entry instead, as I found myself wishing for it multiple times. If you would like to work on that, feel free to open a PR.

yzx9:
Updated: see #8808

---


@eliandoran Thanks for your review and great job.

As I understand it, this PR implements the original #5645 request, which differs from the current #5645 in sereral ways:

1. The issue requests a “Copy as Markdown” feature, which changes the copy behavior itself, while this PR introduces a new export destination without altering the existing copy semantics. The scope and intent of the two approaches are therefore different
2. Implementing "Copy as Markdown" feature would involve significantly more work than this PR, as it touches multiple aspects, including UX design, settings integration, and documentation. In particular, given that there are still some issues with the current design, simply relying on Ctrl+C / Ctrl+V could easily break the formatting
3. Moreover, if users only need this feature occasionally, the current design of "Copy as Markdown" would not provide a good experience. They would either:
   1. Have to enable the option in settings, copy once, and then disable it again; or
   2. We would need to introduce an additional context menu entry / shortcut (for example, Shift + Ctrl + C) specifically for “Copy as Markdown”

Both options require further design considerations and go beyond the scope of this PR

Issue #7895, opened 2025-11-29
(feat) Image Gallery/Collection


lzinga:
### Describe feature

This feature would introduce an image gallery that displays images in a masonry-style grid, making it easier to view, edit, and manage images. I previously created a pull request for a similar idea here: [https://github.com/TriliumNext/Trilium/pull/7879](https://github.com/TriliumNext/Trilium/pull/7879), but after @eliandoran raised some valid points, I decided to move the discussion here for broader input.

In short, the goal is to create a gallery interface for viewing images. I'm bringing @eliandoran's notes from the PR discussion here to serve as the foundation for this conversation.

> * I find it interesting that the gallery is a new note type. It works as a hybrid: supporting both attachments (which makes sense for a note type) and child notes.  
> * Since it supports child notes, it might be more logical to implement it as a [Collection](https://docs.triliumnotes.org/user-guide/collections) rather than as a new note type.  
> * We had a similar case with geomaps, which started as a standalone note type but was later converted into a collection. It's better to make that decision upfront to avoid complicated migrations later.  
> * Using collections also makes it easier to integrate with search notes, allowing searches across the hierarchy.

I'm still familiarizing myself with Trilium as I use it, so there might be some oversights regarding certain features. You're absolutely right that it's best to discuss these ideas first—sometimes my excitement gets ahead of me 🙂

> * We also have to decide whether to support attachments, child notes, or both. While flexibility is nice, maintaining two approaches (like in the calendar) can complicate both upkeep and the user experience.

I took a look at Collections and I appear to understand what you mean here now. I still lean towards potentially supporting both, to me the thought of having all the images as an attachment to the Collection sounds cleaner. However exploring aspects of it, the project does lean more towards having them as a File type under the collection instead. I guess I would want to make sure you could have extra collections underneath the collection so in theory it is a folder-like structure.

As for displaying the images in the top level collection I don't see an issue with an option for the user to select whether or not to show all images plus child images (defaulted to only the current collection) 

> * Also, creating a new note type increases the sync version. If users sync with an older server (a common scenario), the gallery will revert to a File note type.

Yes, I ran into that issue while testing—it would default to the File type if the note_type wasn't explicitly set. I am all for moving the idea to a Collection.

> * I like the lightbox behavior when clicking on a photo. It would be great to extract that component into something reusable, so it could also open images or attachments from Text notes.

👍

[...]

PR #8210, opened 2025-12-29
(feat) Image Gallery/Collection

lzinga:
Adds https://github.com/TriliumNext/Trilium/issues/7895

To make a gallery all you do is make a new note set the collection type to gallery and you can upload images to it which will create child notes as file type and display them in the gallery. For any child gallery it will show up as a folder type in the parent gallery.

#### Multi-Select Operations
- Ctrl+Click: Toggle items
- Shift+Click: Range selection
- Select All button
- Bulk deletes with confirmation

#### Infinite Scroll
- Virtual scrolling (loads 50 items at a time) (loads the data but only shows it when scrolling down)
- IntersectionObserver for smooth performance

#### Image Upload
- Multi-file upload with progress tracking
- Per-file error handling
- Toast notifications for status

#### Nested Galleries
- "New Gallery" button creates child galleries
- Uses _template_gallery template
- Back button for parent navigation
- Child count badges on folders

I kept it somewhat light weight to make the possibility of expanding on it in the future easier depending what people may like or dislike about it.

What it looks like in the tree
Image

what it looks like in the note itself.
Image

As note some changes are small eslint updates that kept getting forced. I tried to undo them but eventually thought it would probably be better to keep them.

### DIFF
--- a/apps/client/src/services/in_app_help.ts
+++ b/apps/client/src/services/in_app_help.ts
@@ -25,7 +25,7 @@ export const byNoteType: Record<Exclude<NoteType, "book">, string | null> = {
 export const byBookType: Record<ViewTypeOptions, string | null> = {
     list: "mULW0Q3VojwY",
     grid: "8QqnMzx393bx",
-    gallery: "W1x7XN1KRSWo",
+    gallery: null,
     calendar: "xWbu3jpNWapp",
     table: "2FvYrpmOXm29",
     geoMap: "81SGnPGMk7Xc",


lzinga:
- Updated it so the toolbar does not show beneath the default empty collection.
- Updated translations.
- Updated inline css to classes.
- Added that it shows subchild mindmap/mermaid/etc attachments as images in the gallery.
- Added the share badge which tells you if the item is shared or not.
- Added the share toggle in the context menu when right clicking an item.
- Added badges onto the items to show what kind of note they are. Image/canvas/mermaid/mindmap

Issue #8198, opened 2025-12-28
New Mermaid diagrams do not save content


Aitanuqui:
### Description

Since version 0.101.0, newly created Mermaid diagrams do not save their content. After creating a new Mermaid diagram and entering diagram code, the content is lost when refreshing the page, switching to other notes and returning, or closing and reopening the application.

This issue is reproducible on the server, as well as the Linux and Windows client applications.

### TriliumNext Version

0.101.0 and 0.101.1

### What operating system are you using?

Other Linux

### What is your setup?

Local + server sync

### Operating System Version

Fedora Workstation 43

### Error logs

I couldn't find relevant front-end logs.

sirhaug:
Experiencing the same issue on version 0.101.1.  
Local (Windows 11) + Server sync (Ubuntu 24.04)

lzinga:
This appears to be occurring due to the commit https://github.com/TriliumNext/Trilium/commit/db3aedf39de14c29a9b72b5803d9bc208863a531

Looks like a `note.type !== noteType` check got added and since mermaid uses a SvgSplitEditor this type isn't passed in somewhere. I am testing some local fixes on it now to see if I can track down anything further.

PR #8220, opened 2025-12-30
fix(mermaid) diagrams not saving content and SVG attachment

lzinga:
Fixes https://github.com/TriliumNext/Trilium/issues/8198

In commit https://github.com/TriliumNext/Trilium/commit/db3aedf39de14c29a9b72b5803d9bc208863a531 it looks like a safety check was added `if (data === undefined || note.type !== noteType) return;` 

it prevents saving "code" data to a note that's now a "text" note (for example, if you changed the note type while edits were pending).

However the noteType was hardcoded in the Code.tsx to save as `code`
```
// In Code.tsx (before fix)
const spacedUpdate = useEditorSpacedUpdate({
    note,
    noteType: "code",  // <-- Always "code"
    ...
});
```

But for a mermaid note the note type is mermaid so the safety check would always end up with a false in that check causing the callback to return early, never saving.

I made the type editable with a default type of code.

and now the SvgSplitEditor can pass in t he noteType depending what it is being used for.

    <SvgSplitEditor
        attachmentName="mermaid-export"
        renderSvg={renderSvg}
        noteType="mermaid"  // <-- Now passes the correct type!
        {...props}
    />

### DIFF
--- a/apps/client/src/widgets/type_widgets/helpers/SvgSplitEditor.tsx
+++ b/apps/client/src/widgets/type_widgets/helpers/SvgSplitEditor.tsx
@@ -77,7 +77,7 @@ export default function SvgSplitEditor({ ntxId, note, attachmentName, renderSvg,
             if (!attachments.find((a) => a.title === `${attachmentName}.svg`)) {
                 onSave();
             }
-        });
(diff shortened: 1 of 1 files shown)


eliandoran (approved):
Nice job!

PR #8369, opened 2026-01-13
fix: mac alt shortcuts

chloelee767:
Hello, this is a PR to fix the following issue that I have noticed in mac (both desktop app and server website): keyboard shortcuts using alt (eg. alt+a) don't trigger, they instead fallback to the mac default behaviour to enter special characters (eg. å in this case).  Now, alt+a and other alt key shortcuts will trigger the Trilium shortcut, if it exists, if not, it will fallback to the default mac behaviour.


browneyedsoul (commented):
Great Job!

eliandoran (approved):

Issue #7447, opened 2025-10-21
Milestone: Official mobile application


eliandoran:
One of the core issues with the Trilium at the moment is that it doesn't really have a fully working mobile application. There is [TriliumDroid](https://github.com/FliegendeWurst/TriliumDroid) and I appreciate it, but I would like to take it in a completely other direction (from a development point of view) so it will be written from scratch.

I am thinking of implementing it as a React Native application with a relatively simple but efficient UI and to try to reuse as many components as possible. For example, the editor (CKEditor, CodeMirror) will still be in use.

As opposed to a PWA, I would like to implement it in a way that's unlike some of our competitors the application will be designed to work fully offline, with its own sync capability.

I plan to support both iOS and Android.

This is a significant undertaking and it will require me taking unpaid time off the job in order to implement it, so financial support via:
* [A bounty on Issuehunt](https://oss.issuehunt.io/r/TriliumNext/Trilium/issues/7447)
* [GitHub Sponsors](https://github.com/sponsors/eliandoran)
* [PayPal](https://paypal.me/eliandoran)

I expect this effort to take several months.

Ideas or concerns are welcome.

capi:
Just consider: re-implementing for mobile is not only a huge one-time effort, it's an ongoing effort. I don't have experience in this, but would hosting a node application inside a native shell application be a possibility? The shell would launch the node process in the background, which does have sync capabilities and everything that the web application can (which is almost everything), and provide a WebView front-end to show the content, manage login, provide a service so that sync can run even while not open, etc.

eliandoran:
> Just consider: re-implementing for mobile is not only a huge one-time effort, it's an ongoing effort.

That is indeed right. That's also the reason why I would want to share as much as possible with the actual application. 

> I don't have experience in this, but would hosting a node application inside a native shell application be a possibility?

It certainly is possible but there a few issues: Node.js on mobile is not very mature and maintained. And I think there is no iOS support.

deajan:
@eliandoran I have no real experience on mobile development, but I would have a question from a user standpoint.

Considering that almost every mobile user has 4G data almost everywhere, why is there a need for sync ?
I mean, if you plan to heavily work on your notes in some no network zone, then you probably take your laptop anyways, which is synced.

[...]

Issue #7225, opened 2025-10-08
Trilium Sync Blocks UI During Content Hashing for Large Knowledge Graphs


werererer:
### Description

Trilium blocks when it syncs during `content hashing`, In my case the worst case was 20 seconds the average is around 3-4 seconds. It seems to be linear correlated with note count (~200k) or database size (2.1 Gb). While other Processes are also affected I think that solving it from sync may be the most natural progression.

The Solution seems to be implementing parallelization either through processes or worker threads. Promises, Asnyc § Await, etc still lead to blocking behavior, if a process requires lots of CPU, since they offer concurrency not parallelism [^2].

## Difficulties during implementation attempt
I tried to Implement basic behavior already
- We use Typescript, but spawning processes or worker threads require js files. Precompiling these files from ts to js might be a solution
- Finding out which environment Variables need to be passed to jobs is hard
- Job files can't be executed if they are packed in ASAR, however unpacking might be an option [^4]

## Implementation Strategy
- Prob try to parallalize a simple case first. Enhance later.

### Taking inspiration from Vscode
Vscode is a performant texteditor, because of its multithreaded nature, and thus a valuable resource to make trilium less blocking.
```
git clone https://github.com/microsoft/vscode
```

A quick look at Vscode [^1]. Should help find solutions.
Executing:
```
rg 'spawn\('
```
or
```
rg 'fork\('
```
lets you see vscode spawning child processes.

Executing
```
rg 'new Worker'
```
lets you see how vscode handles Workers

## Conclusion
Trilium needs to introduce a way to do parallelism, to reduce blocking the user which kills user experience [^3].

## References
[^1]: https://github.com/microsoft/vscode 
[^2]: https://stackoverflow.com/questions/1050222/what-is-the-difference-between-concurrency-and-parallelism
[^3]: https://www.nngroup.com/articles/website-response-times/ 
[^4]: https://www.electronjs.org/de/docs/latest/tutorial/asar-archives

### TriliumNext Version

Version 0.99.1

### What operating system are you using?

Other Linux

### What is your setup?

Local + server sync

### Operating System Version

Archlinux

### Error logs
<details>
<summary>Dump</summary>

 _____     _ _ _
|_   _| __(_) (_)_   _ _ __ ___   | \ | | ___ | |_ ___  ___
  | || '__| | | | | | | '_ ` _ \  |  \| |/ _ \| __/ _ \/ __|
[...]

PR #8492, opened 2026-01-24
refactor(search): move scoring logic to native Rust module

werererer:
I maintain a collection with over 250k notes on relatively weak hardware, which makes search and autocomplete noticeably slow. (Up to 4 seconds per query)

This PR introduces Rust modules by replacing one small performance-critical function with a Rust implementation. The goal is to establish an initial integration point for gradually moving hot paths into Rust in follow-up work, in order to better support very large personal knowledge graphs or future workloads involving ontologies or LLM-generated graphs. Rust provides a foundation for exploring future optimizations such as reduced allocation overhead and parallel execution for long-running tasks.

This work is inspired by the previous effort here and may supersede it:
https://github.com/TriliumNext/Trilium/pull/7222

### Open Question
- [ ] Will it hinder work on mobile implementation?
https://github.com/TriliumNext/Trilium/issues/7447

### TODO
- [ ] Testing, and improving stability.


eliandoran:
> which makes search and autocomplete noticeably slow. (Up to 4 seconds per query)

What's the performance improvement with the Rust implementation?

itsjakobx:
The current implementation didn't improve performance, likely because it is implemented very inefficiently because of the serilization cost or because it didn't covere the most contributing factor to search performance.

I will need to move more of scoring into rust to give a more fair evaluation.

This PR won't go out of Draft until I see a notable performance increase.

itsjakobx:
I will try to find another solution based on worker Threads instead
Based on previously abonded work.
https://github.com/TriliumNext/Trilium/issues/7225 

This will take a while to cook up.

perfectra1n:
I think that the FTS5 PR #6839 will really help with this @werererer

PR #8494, opened 2026-01-24
Web Clipper

eliandoran:


contributor:
Web clipper is returning to Firefox. This is great!

eliandoran:
@contributor , sure it is. :)
For me it's important to preserve Firefox support.
Depending on how popular the web clipper turns out to be, we might invest more time in it (nicer UX, support for Firefox Mobile, etc).

contributor:
> @contributor , sure it is. :) For me it's important to preserve Firefox support. Depending on how popular the web clipper turns out to be, we might invest more time in it (nicer UX, support for Firefox Mobile, etc).

Looking forward!

Meanwhile, consider conditionally change suggested hotkeys for Firefox
```ts
if (import.meta.env.FIREFOX) {
  // define hotkeys for firefox
}
```

Current `Ctrl+Shift+S` & `Alt+Shift+S` hotkeys do not work (and conflict with built in Firefox screenshot hotkey) 
https://github.com/TriliumNext/trilium/blob/1cf93ff0dec89ee1a80654934cb30fad74920043/apps/web-clipper/wxt.config.ts#L32-L52

I use these:
image

Mnemonics:
- ob**J**ect (most frequently used and the easiest to type)
- **P**icture

`Save current page` is set to `Shift+Alt+O` but I never save whole page.

eliandoran:
@contributor , thanks for the insights. I was aware of the keyboard shortcut conflict but I will fix it in a follow-up PR.

PR #8525, opened 2026-01-27
fix: subtree expand only first level + depth limit

Nriver:
- Correctly recurses through multiple levels when expanding (expanded=1)
- Only updates branches that are actually collapsed (isExpanded=0) when expanding
- Only updates branches that are actually expanded (isExpanded=1) when collapsing
- Adds depth limit of 4 levels when expanding to prevent performance issues in deep trees
- Collapsing remains unlimited (full subtree)

### DIFF
--- a/apps/server/src/routes/api/branches.ts
+++ b/apps/server/src/routes/api/branches.ts
@@ -152,25 +152,60 @@ function setExpanded(req: Request) {
 
 function setExpandedForSubtree(req: Request) {
     const { branchId } = req.params;
-    const expanded = parseInt(req.params.expanded);
+    const expanded = parseInt(req.params.expanded, 10);
 
-    let branchIds = sql.getColumn<string>(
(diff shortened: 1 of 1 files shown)


maphew (commented):
The recursive CTE has a separate clone-path problem at head `221ab39ab`. [Traversal joins children through `parentNoteId = t.noteId` and retains every row with `UNION ALL`](https://github.com/TriliumNext/Trilium/blob/221ab39ab7965a9c7f392bc127c5cd3a6afeb5ac/apps/server/src/routes/api/branches.ts#L172-L190). If the same cloned note is reached through two branches, each path emits the same child branches again. Further clone convergence compounds those duplicates, especially during collapse with a depth limit of 9,999.

This adds avoidable database, response, Froca lookup, and UI work for clone-heavy trees. Please deduplicate traversal by branch ID while retaining the minimum depth, and add a test with two clone paths that converge on the same note.

PR #8623, opened 2026-02-05
always show Note Paths in Status Bar

contributor:
This is useful if a note is opened from a saved search (no other way to find out note's path).

In old layout it works ok (you can see the paths in search result), but in new layout this is one the things that was not done right.

Note: I left `enabled` flag in use. In the future visibility of status bar items may be controlled through options.

### DIFF
--- a/apps/client/src/widgets/layout/StatusBar.tsx
+++ b/apps/client/src/widgets/layout/StatusBar.tsx
@@ -414,7 +414,7 @@ function NotePaths({ note, hoistedNoteId, notePath }: StatusBarContext) {
     const dropdownRef = useRef<BootstrapDropdown>(null);
     const sortedNotePaths = useSortedNotePaths(note, hoistedNoteId);
     const count = sortedNotePaths?.length ?? 0;
-    const enabled = count > 1;
+    const enabled = true;
 
     // Keyboard shortcut.
     useTriliumEvent("toggleRibbonTabNotePaths", () => enabled && dropdownRef.current?.show());


eliandoran:
@contributor , is search notes the only use case where the paths need to be shown?
If so, can't we detect if the note has a search note path instead? (I saw something like that in `fnote` where you do `FNote#getParentNotes` and check if at least one of them has `type === "search"`).

There is a reason why we are hiding the note paths in the status bar: they occupy unnecessary space if there is only one path. I would argue the number of notes with one path is generally larger than the ones with multiple paths.

contributor:
> @contributor , is search notes the only use case where the paths need to be shown? If so, can't we detect if the note has a search note path instead? (I saw something like that in `fnote` where you do `FNote#getParentNotes` and check if at least one of them has `type === "search"`).

We could try, but is it worth to make a special case?

Even for a regular note, it is good to have path accessible. Bread crumb often has abbreviated paths even on a wide monitor:
image

> 
> There is a reason why we are hiding the note paths in the status bar: they occupy unnecessary space if there is only one path. I would argue the number of notes with one path is generally larger than the ones with multiple paths.

Space concerns should be resolved by allowing user to customize what to see on status bar. I would remove Attachments from there right away - it is the most useless piece on status bar in my opinion.

And on top of that it would be good to have ability **always**  show pieces you have chosen. I prefer stable static information even with zero counts instead of pieces jumping left and right as you go through your notes.

contributor:
@eliandoran 

Currently, mobile UI does show note path button, even if there is onlye one path:
image

This PR does similar thing for desktop.

Do you still believe it is worth to do checks and conditionally show/hide note paths on status bar?

PR #8640, opened 2026-02-06
Implement relation name wildcard feature

misch334:
Trilium Notes currently has the limitation, that one needs to explicitely specify the kind of relation to query for, when using search dialog. Users need to actively
- decide, if a link is internal (embedded in note content) or an explicit relation - causing friction during dumping of thoughts
- remember, which link type has been chosen, in order to retrieve desired note connections
 
Intent of this PR is to reduce this mental load of users. It also strives to align Trilium Notes with other Note taking apps.An example is Logseq, where page links can be unformely queried, regardless if properties (akin to Trilium explicit relations) are used or in-content links. Further filtering by properties is just optional.
 
I used a dummy database with note size of 10.000 to test performance (let me know, if shell script for dummy note creation is of interest). There wasn't any noticeable delay in search dialog in combination with relation wildcard search `~?`.


apps/server/src/services/search/expressions/relation_where.ts
```diff
@@ -19,7 +25,9 @@ class RelationWhereExp extends Expression {
     execute(inputNoteSet: NoteSet, executionContext: {}, searchContext: SearchContext) {
         const candidateNoteSet = new NoteSet();
 
-        for (const attr of becca.findAttributes("relation", this.relationName)) {
+        for (const attr of this.relationName === RELATION_NAME_WILDCARD
+            ? Object.values(becca.attributes)
+            : becca.findAttributes("relation", this.relationName)) {
```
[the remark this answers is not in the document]
misch334: This seems to be more inefficient than current solution, as iterating an extra round in the noteset.

apps/server/src/services/search/expressions/relation_where.ts
```diff
@@ -5,6 +5,12 @@ import NoteSet from "../note_set.js";
 import becca from "../../../becca/becca.js";
 import type SearchContext from "../search_context.js";
 
+// This wildcard should be a symbol/character different from lexer operators.
+// It also should use char(s) not allowed in relation names ([0-9a-zA-Z:_]),
+// to prevent ambiguity. As alternative in future, a more sophisticated lexer
+// might create separate token types, with relation wildcard being one of them.
+const RELATION_NAME_WILDCARD = "?";
```
misch334, line 12: One thing to add: By “lexer” [this module](https://github.com/TriliumNext/Trilium/blob/93c5413790142f7c3f8cb9b1695efe303eb55ff3/apps/server/src/services/search/services/lex.ts) is meant. Its operators are used to split the tokens, hence e.g. `*` was not possible to use as relation name wildcard.

Currently this lexer returns an [array of tokens](https://github.com/TriliumNext/Trilium/blob/93c5413790142f7c3f8cb9b1695efe303eb55ff3/apps/server/src/services/search/services/lex.ts#L39-L44). Later on it might be extended as “more sophisticated” lexer to also return token type identifiers, like `REL_NAME_WILCARD` for char `?` within relation name. These enriched tokens then can be interpreted more flexibly by `parse.ts`. 

Though this PR seems sufficient for now.

misch334: Another alternative is to add conditional logic for the token seqence `~?` (Relation `~` + whatever symbol the relation wildcard has, here `?`) in lexer, so that wildcard is not mis-interpreted as one of the [lexer operators](https://github.com/TriliumNext/Trilium/blob/93c5413790142f7c3f8cb9b1695efe303eb55ff3/apps/server/src/services/search/services/lex.ts#L15-L17). But this change would only be needed, if we should decide to re-use an existent operator symbol like `*`. 

With `?` it works without any lexer modifications, and question mark is easily interpreted as wildcard, hence I am in favor of it.

maphew (commented):
The advertised bare wildcard does not use the new wildcard implementation at head `618f4b550`. [The search help says `~?` returns notes with any relation](https://github.com/TriliumNext/Trilium/blob/618f4b550b716c9dc802608627c6316a5524a774/apps/client/src/widgets/ribbon/SearchDefinitionOptions.tsx#L169-L174), but `parseRelation("?")` sends a bare relation to `AttributeExistsExp`. That expression looks up the literal relation name `?`, so the result is empty. Only property searches such as `~?.title = Tolkien` reach `RelationWhereExp` and activate the wildcard branch.

Please route a bare `?` relation through the wildcard-aware expression, or remove the bare syntax from the help. Add a parser-to-execution test for both `~?` and `~?.title = Tolkien`; the current test constructs `RelationWhereExp` directly and cannot catch this failure.

PR #8654, opened 2026-02-07
Trilium preview import

eliandoran:


maphew (changes requested):
I found three correctness issues in the import-preview flow:

1. `handleImportArguments()` previews the external `.trilium` file and then stores that same path for later import. The file can be replaced after the user approves the preview, causing different content to be imported. Advise copying the file into the managed import store first, then previewing and importing that immutable copy.

2. `executeUploadWithPreview()` starts every `server.post()` without awaiting it. The requests share one task ID, and only the final request is marked `last`, but that request can finish first and report completion while earlier requests are still changing the tree. Advise awaiting the imports sequentially so `last` also means last to complete.

3. `handleImportArguments()` registers `ipcMain.on("import-preview-ready")` after each invocation, while the renderer sends that ready event only once when it mounts. Opening a `.trilium` file while Trilium is already running therefore cannot deliver its preview and also accumulates listeners. Advise keeping one startup-level readiness handler and explicit pending-preview state, or sending directly once the renderer is known to be ready.

PR #8677, opened 2026-02-11
feat(note-map): open node in split on cmd/ctrl+click

JeremieAlcaraz:
## Summary
- add `Cmd/Ctrl + click` handling on Note Map nodes
- keep normal click behavior unchanged

## Why
- improve navigation speed from graph view by opening a target note directly in a right split

## Scope
- `apps/client/src/widgets/note_map/NoteMap.tsx`

## Testing
- local manual test: normal click still opens in current split
- local manual test: `Cmd` (macOS) / `Ctrl` (Windows/Linux) + click opens a new split
- build check: `pnpm run --filter client build`


apps/client/src/widgets/note_map/NoteMap.tsx
```diff
@@ -172,3 +180,7 @@ function getCssData(container: HTMLElement, styleResolver: HTMLElement): CssData
         mutedTextColor: rgb2hex(styleResolverStyle.color)
     }
 }
+
+function isSplitModifierClick(event: MouseEvent | undefined) {
```
eliandoran, line 184: Use isCtrlKey from utils instead.

eliandoran:
Although the implementation by itself is good, I believe that if we add support for this feature of opening in a split we have to add support for it everywhere (tree, geo map, list/grid collections). We cannot introduce inconsistent behaviour when it comes to frequent user actions.

I will take note of it and implement it globally at some point.

Thanks anyway for the contribution. :)

Issue #5080, opened 2024-07-18
Investigate #titleTemplate for template notes


eliandoran:
According to [Template](https://github.com/zadam/trilium/wiki/Template) documentation:

> See also [default note title](https://github.com/zadam/trilium/wiki/Default-note-title) which allows you to create templates for note titles. Note templates and title templates can be combined by creating a #titleTemplate for a template note.

However attempting to add `#titleTemplate` for a template note does not take effect when creating a new note with the given template.

PR #8787, opened 2026-02-21
feat: add childTitleTemplate attribute and update titleTemplate attri…

monotok:
This pull request introduces the following changes:

    Adds a new childTitleTemplate attribute.
    Updates the titleTemplate attribute to ensure it behaves as expected.

Fixes issue https://github.com/TriliumNext/Trilium/issues/5080


eliandoran (changes requested):
Overall seems fine, but please also update the user guide. Ideally, we should also have a comparison between `titleTemplate` and `childTitleTemplate`.

And also clarify the case where you have both a `titleTemplate` and a `childTitleTemplate`. What happens then?

monotok:
Hello, thanks for looking at it. I was waiting to see if the change was acceptable before updating the user guide. Will update the user guide :+1:

eliandoran:
@monotok , it would also be cool to write some integration tests. See `hidden_subtree.spec.ts` from `server` as inspiration.

monotok:
I am not sure why running `pnpm edit-docs:edit-docs` causes other files to change. Initially it changed 500+ doc files but I reverted that commit and force pushed.

Issue #8798, opened 2026-02-22
Enable FullCalendar Recurring Functionality


BeatLink:
### Describe feature

It would be great if the built in calendar collection supports repeating tasks. 

All that would be needed is the inclusion of the rrule fullcalendar plugin here https://fullcalendar.io/docs/rrule-plugin

Then the recurrence string can be set from the #calendar:rrule attribute of a given note in the collection

Issue #7114, opened 2025-09-27
Allow Setting RRULE string for calendar collections


BeatLink:
### Describe feature

It would be great if the calendar note could support showing repeated tasks by having each child note set a #rrule label with the value set to the RRULE recurrence string

PR #8799, opened 2026-02-22
Add Support for Recurrence In Calendar Collection

BeatLink:
This PR enables the calendar collection to support recurrences.

If a child note of a calendar collection has a `#recurrence label` with an ICAL recurrence string e.g. `FREQ=WEEKLY;INTERVAL=5;` it will repeat multiple times in the calendar view based on its recurrence.

Normally the FullCalendar recurrence plugin requires a recurrence string with DTSTART, but this is not included in the string, in favor of generating the DTSTART value based on the already existing `#startDate`  label. 

Also, fullCalendar requires a duration property to be added when using recurrence, otherwise it would treat it as an all day event. This is solved by dynamically calculating and setting the duration property from the difference of the start and end date if the latter exists.


apps/client/src/widgets/collections/calendar/event_builder.ts
```diff
@@ -118,6 +120,17 @@ export async function buildEvent(note: FNote, { startDate, endDate, startTime, e
...
+        if (recurrence) {
+            eventData.rrule = `DTSTART:${startDate.replace(/[-:]/g, "")}\n${recurrence}`;
+            if (endDate){
+                const duration = (d => 
+                    String(d / 36e5 | 0).padStart(2, "0") + ":" +
+                    String(d / 6e4 % 60 | 0).padStart(2, "0")
+                )((new Date(endDate)) - (new Date(startDate)));
+                eventData.duration = duration
+            }
+        }
```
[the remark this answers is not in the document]
BeatLink: @gemini-code-assist 

Can you provide a link to the documentation explaining that the event will fail if end is specified?

BeatLink:
Closes https://github.com/TriliumNext/Trilium/issues/8798

BeatLink:
Closes #7114

BeatLink:
@eliandoran Hey, I was wondering if you could take a look at this.

apps/client/src/widgets/collections/calendar/event_builder.ts
```diff
@@ -118,6 +120,17 @@ export async function buildEvent(note: FNote, { startDate, endDate, startTime, e
...
+        if (recurrence) {
+            delete eventData.end;
+            eventData.rrule = `DTSTART:${startDate.replace(/[-:]/g, "")}\n${recurrence}`;
+            if (endDate){
+                const diffMs = new Date(endDate).getTime() - new Date(startDate).getTime();
+                const hours = Math.floor(diffMs / 3600000);
+                const minutes = Math.floor((diffMs / 60000) % 60);
+                eventData.duration = `${String(hours).padStart(2, "0")}:${String(minutes).padStart(2, "0")}`;
+            }
+        }
```
eliandoran, line 160: Missing tests. The event builder has pretty decent coverage and I would like to keep it that way.

docs/User Guide/User Guide/Collections/Calendar.md
```diff
@@ -72,11 +72,12 @@ For each note of the calendar, the following attributes can be used:
 | `#endDate` | Similar to `startDate`, mentions the end date if the event spans across multiple days. The date is inclusive, so the end day is also considered. The attribute can be missing for single-day events. |
 | `#startTime` | The time the event starts at. If this value is missing, then the event is considered a full-day event. The format is `HH:MM` (hours in 24-hour format and minutes). |
 | `#endTime` | Similar to `startTime`, it mentions the time at which the event ends (in relation with `endDate` if present, or `startDate`). |
+| `#recurrence` | This is an optional CalDAV `RRULE` string that if present, determines whether a task should repeat or not. Note that it does not include the `DTSTART` attribute, which is derived from the `#startDate` and `#startTime` directly. For examples of valid `RRULE` strings see [https://icalendar.org/rrule-tool.html](https://icalendar.org/rrule-tool.html) |
```
eliandoran, line 75: Although it's good that it's mentioned in the attribute reference, an important feature such as recurrence needs its own heading in the Calendar documentation.

The tool provided is decent, but I would also add some concrete examples. Mention also that promoted attributes can be used to define an easy way to input recurrence.

Of course, in the future we ought to have a nice recurrence editor.

BeatLink: I have a recurrence editor.

apps/client/src/widgets/collections/calendar/event_builder.ts
BeatLink: Tests Added

eliandoran:
@BeatLink , a non-critical bug you might want to address in a separate PR, all-day events that are recurring display a time of 00 and they don't show up in the week view as all-day events.

image

Sample:

```
#startDate=2026-02-02 #endDate=2026-02-02 #recurrence="FREQ=WEEKLY;INTERVAL=1;BYDAY=MO,TU"
```

PR #8808, opened 2026-02-23
feat(client/menus): add Copy as Markdown

### LINKED ISSUE #5645: Copy as Markdown

yzx9:
## Summary                                                                                                    

- Add desktop context menu option for copying selected content as markdown
- Includes translations for 16 languages

## Screenshots

electron context menu:

image

cc @eliandoran 

close #5645 #8009

### DIFF
--- a/apps/client/src/translations/en/translation.json
+++ b/apps/client/src/translations/en/translation.json
@@ -1810,7 +1810,7 @@
     "add-term-to-dictionary": "Add \"{{term}}\" to dictionary",
     "cut": "Cut",
     "copy": "Copy",
-    "copy-as-markdown": "Copy as markdown",
+    "copy-as-markdown": "Copy as Markdown",
     "copy-link": "Copy link",
     "paste": "Paste",
     "paste-as-plain-text": "Paste as plain text",


apps/client/src/menus/tree_context_menu.ts
```diff
@@ -275,6 +275,8 @@ export default class TreeContextMenu implements SelectMenuItemEventListener<Tree
 
             { title: t("tree-context-menu.export"), command: "exportNote", uiIcon: "bx bx-export", enabled: notSearch && noSelectedNotes && notOptionsOrHelp },
 
+            { title: t("tree-context-menu.copy-as-markdown"), command: "copyAsMarkdown", uiIcon: "bx bx-copy-alt", enabled: notSearch && noSelectedNotes && notOptionsOrHelp },
```
eliandoran: To be honest, I don't see the point in having the option to copy as markdown as part of the tree. It's not a general editing pattern.

Let's keep the only the context menu implementation for now.

apps/client/src/menus/electron_context_menu.ts
```diff
@@ -60,6 +62,37 @@ function setupContextMenu() {
...
+                            const selection = window.getSelection();
+                            if (!selection.rangeCount) return '';
+
+                            const range = selection.getRangeAt(0);
+                            const div = document.createElement('div');
+                            div.appendChild(range.cloneContents());
+                            return div.innerHTML;
+                        })()
+                    `);
+
```
eliandoran: Can this be simplified, why is `executeJavaScript` required?

eliandoran (changes requested):
Thanks for the implementation, it better aligns with our functional expectations, but I still have a few topics that need addressing.

apps/client/src/menus/tree_context_menu.ts
yzx9: The context menu is only available in the Electron version, whereas the tree context menu works in both the web and Electron versions.

apps/client/src/menus/electron_context_menu.ts
yzx9: Fixed. I overlooked the fact that this script runs in the Renderer process.

apps/client/src/menus/tree_context_menu.ts
eliandoran: @yzx9 , I understand your concern but the fact remains that tree is for note-related options, there are no other content-copy operations in it.

My suggestion is to only keep the Electron variant for now. Should you wish to implement it for desktop as well, we can go for:

1. Making the context menu available on the browser.
2. Making it as a keyboard shortcut.

Either way, I would treat those as a separate PR to unblock this one. But removing it from the tree is necessary, regardless.

yzx9: I have removed the related code for adding the tree context menu

eliandoran:
@yzx9 , please don't forget to put PRs in ready when they're done. Sometimes I don't notice the notification, but I regularly check the open/non-draft PRs.

PR #8864, opened 2026-03-01
add toolbar customization over drag & drop

tuuuni-cell:
Adds a new "Toolbar" tab to the Text Notes options panel, allowing users to
fully customize the toolbar.

- **Drag & drop reordering** of toolbar items 
- **Click-to-add** chips from a pool of all available items  
- **Groups** — collapsible sections that map to CKEditor toolbar groups, with their own nested drag & drop  
- **Remove** any item by clicking ✕ or dragging it to the sticky delete zone at the bottom  
- **Drop indicator** — a glowing blue line shows exactly between which two items the dragged element will land  

Its my first time contributing do you need something more?

image


tuuuni-cell:
@gemini-code-assist review

eliandoran:
@tuuuni-cell , thanks for your work.

I have tested the functionality and it's pretty good but I don't think it matches currently the customization mechanisms that Trilium uses, say for the launch bar.

I'll probably attempt a change on my side. Although it will be less user-friendly than your implementation, it would align with the rest of the application.

PR #8894, opened 2026-03-03
web-clipper: prevent cloning when clipping to an existing note + optimizations

contributor:
This removes undesired note cloning when clipping to existing note

### Background
I noticed lot of cloned notes in clipper inbox appeared periodically. I blamed sync, but it turned out to be web clipper handler bug/feature.

The cloning was introduced in https://github.com/TriliumNext/Trilium/pull/4075/changes

But this does not works properly in my opinion:
- when clipping note is in inbox, it still tries to clone it (and siliently fails due to validation)
- when clipping note has been moved outside of inbox, it clones it to inbox (cluttering already organized inbox)

### Proposed solution

- When you clip an URL for the first time the note appears in clipper inbox (like it was before this PR)
- When you clip the same URL another time, it keeps appending to the note regardless of note location, and does not clone it ("does not clone it" part is added in this PR) 

### Additional refactoring made
Finding clipping note code is now more correct semantically (uses `searchService.searchNotes` method instead of misleading indirection via  `note.searchNotesInSubtree` method. As a bonus, it saves extra call to `getClipperInboxNote` on each click on browser extension, and in case of clipping selection to an exiting note.

### DIFF
--- a/apps/server/src/routes/api/clipper.ts
+++ b/apps/server/src/routes/api/clipper.ts
@@ -19,12 +19,11 @@ async function addClipping(req: Request) {
     const { title, content, images } = req.body;
     const clipType = "clippings";
 
-    const clipperInbox = await getClipperInboxNote();
-
     const pageUrl = sanitize.sanitizeUrl(req.body.pageUrl);
     let clippingNote = findClippingNote(pageUrl, clipType);
 
     if (!clippingNote) {
+        const clipperInbox = await getClipperInboxNote();
         clippingNote = noteService.createNewNote({
             parentNoteId: clipperInbox.noteId,
             title,
--- a/apps/server/src/routes/api/clipper.ts
+++ b/apps/server/src/routes/api/clipper.ts
@@ -14,9 +14,9 @@ interface Image {
 }
 
 async function addClipping(req: Request) {
-    // if a note under the clipperInbox has the same 'pageUrl' attribute,
-    // add the content to that note and clone it under today's inbox
-    // otherwise just create a new note under today's inbox
+    // if a note exists with has the same 'pageUrl' attribute,
+    // add the content to the existing note
+    // otherwise create a new note under today's inbox
     const { title, content, images } = req.body;
     const clipType = "clippings";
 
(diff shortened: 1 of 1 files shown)


maphew (commented):
One cleanup remains at head `cda957c21`: [`cloneService` is still imported](https://github.com/TriliumNext/Trilium/blob/cda957c21aa664d141b45298267e40ff2bc971e2/apps/server/src/routes/api/clipper.ts#L1), but this change removes its last use. Removing that obsolete import is sufficient.

Caller tracing did not find a behavior regression in the new global URL lookup or the removal of cloning.

maphew:
@contributor, I prepared a rescue branch for this PR, rebased onto current upstream `main`:

- [branch](https://github.com/maphew/Trilium/tree/rescue/pr-8894-clipper-no-clone)
- [compare with upstream `main`](https://github.com/TriliumNext/Trilium/compare/main...maphew:rescue/pr-8894-clipper-no-clone)

Your five original commits remain first with their original authors, author dates, and messages. A separate [repair commit `b64073806`](https://github.com/maphew/Trilium/commit/b64073806fdb4c120dbfeeeec08d2f6f2ea319cb) removes the obsolete cloning import and adds coverage proving that a moved clipping note is reused without being cloned back into the inbox.

Validation:

- focused server tests: 10 passed
- `pnpm typecheck`: no errors

No pressure to use the branch. You can absorb the full rebased branch or cherry-pick only the repair commit onto your PR branch.

contributor:
@maphew 

Thanks for the update.

Not touching ugly import statements like
```js
import { app_info as appInfo, attribute_formatter as attributeFormatter, attributes as attributeService, type BNote, cloning as cloneService, date_notes as dateNoteService, date_utils as dateUtils, getLog, note_service as noteService, sanitize, search as searchService, ValidationError, ws } from "@triliumnext/core";
```
to remove unused import was kinda intentional - otherwise PRs goes into conflict very quickly 

I will cherry pick the test if it is valuable/correct. It does change name
"creates a clipping note, then appends to it when clipped from the same URL" -> "reuses a moved clipping note without cloning it back into the inbox" which sounds like changing the purpose of original test.

Issue #5650, opened 2025-04-12
TOTP security enchancement


JYC333:
> The length by itself is not a good measure.

Off course not, but it is a part of it.

1. Secret Length:

A TOTP secret with 16 Base32 characters equals 80 bits of entropy (16 characters × 5 bits per Base32 character = 80 bits). This is barely sufficient for TOTP but sits at the lower end of the secure spectrum.

3. Hash Algorithm (SHA-1):

SHA-1 is no longer considered cryptographically secure, mainly due to known collision vulnerabilities. In the context of TOTP, however, SHA-1 is used for HMAC, not for digital signatures, which makes the risk somewhat lower. Still, SHA-256 or SHA-512 would be more appropriate today.

5. Missing OTP rate limiting is a serious security concern

• The TOTP code changes every 30 seconds, but during that window, an attacker can still try up to 1 million possible codes (000000–999999).
• If there’s no enforced delay or cap on failed attempts, an attacker could theoretically try thousands of codes within those 30 seconds — especially if using automation.
• Even with the 30s limit, without throttling, offline brute-force attempts (e.g. on self-hosted instances) are realistic.

Best practice is to enforce:
	•	a rate limit (e.g., 3–5 attempts per minute per account/IP) — TriliumNext/trilium#5661 
	•	lockout or CAPTCHA after repeated failures
	•	logging and alerting for suspicious activity

Comparison:
Many modern TOTP implementations (e.g., GitHub, Google) use:
	•	a 160-bit secret (Base32 = 32 characters)
	•	SHA-256 or SHA-512 as the HMAC algorithm (often configurable)

BTW: since my last update Trilium Notes doesn't send any error codes like (401|403|404|429|500) back on failed login attempt anymore. So there is no use case for fail2ban at caddy point anymore.

Logs looks like:
`178.13.xxx.xxx - - [12/Apr/2025:06:14:28 +0000] "GET / HTTP/2.0" 302 34`
`178.13.xxx.xxx - - [12/Apr/2025:06:14:28 +0000] "GET /login HTTP/2.0" 200 819`
`178.13.xxx.xxx - - [12/Apr/2025:06:19:41 +0000] "GET /login HTTP/2.0" 304 0`
`178.13.xxx.xxx - - [12/Apr/2025:06:19:56 +0000] "POST /login HTTP/2.0" 200 858`

0.92.4 provides 401 on failed login attempt:
[...]

PR #8920, opened 2026-03-04
Enforce TOTP verification during desktop sync setup

### LINKED ISSUE #8209: Allow disabling app access when using TOTP
### Description

As it stands, enabling TOTP works well for securing browser logins by adding an extra layer of authentication.

However, there should be an option to restrict access from applications that don’t yet support the TOTP process. Adding this control would give instance owners more flexibility and improve security until those apps implement TOTP support. Currently, a malicious user could bypass TOTP protection simply by connecting through a desktop app instead of the browser.

While some rate limiting appears to be in place for apps, the lack of TOTP enforcement still leaves the system vulnerable. If TOTP is enabled on an account, I would expect it apply to every login attempt, unless for specific exceptions where things like an ETAPI token is being used or backup codes.

I created this as a bug as I don't believe it is a major security issue and not a feature as I believe the expected result is if you apply a TOTP to an account it prevents any access unless the key is given (apart from things like API calls or using the backup codes)

### Possible Solutions
A potential solution would be to introduce a “Require TOTP for Login” setting on the server. When enabled, this setting would enforce TOTP verification for all login attempts, blocking access from any application that doesn’t support TOTP. The configuration page should also make it clear that if this setting is disabled, users can still log in through non‑TOTP‑enabled apps without providing a code.

Alternatively, there could be an option to allow temporary logins without TOTP, controlled by a timer that automatically re‑enables full TOTP protection after the user signs in. This would let someone briefly disable TOTP to log in on a desktop app, after which the account would automatically return to its secure state.

### Ultimate Goal
When using a remote server and an app to sync require any access to use the TOTP.

### TriliumNext Version

0.101.1

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 11

### Error logs

_No response_

### LINKED ISSUE #8272: Desktop Sync has no 2FA
### Describe feature

Currently, Desktop Sync does not have two-factor authentication (2FA). Adding 2FA would enhance the security

### Additional Information

_No response_

JYC333:
address part of #5650 
close #8209 
close #8272


maphew (commented):
TOTP discovery does not use the proxy entered on the setup form. [The discovery request forces `proxy: null`](https://github.com/TriliumNext/Trilium/blob/f123d50631e71cf7bd0f75f97a32a8f5da80ad8a/apps/server/src/services/setup.ts#L120-L129), and every request failure is returned as `totpEnabled: false`. In a proxy-only environment, the client therefore hides and clears the TOTP field. The later seed request uses the configured proxy, but sends no TOTP token and receives a 401 from a TOTP-enabled server.

Please pass `syncProxy` through discovery and preserve an `unknown` state on network failure. A simpler design is to remove discovery and let the authenticated seed request return a structured TOTP challenge. Add a proxy-only setup test so both requests must use the same connection settings.

PR #8972, opened 2026-03-09
Migrate autocomplete

### LINKED ISSUE #5134: Upgrade autocomplete dependencies
Autocomplete is recommended to migrate to @algolia/autocomplete-js.

https://github.com/algolia/autocomplete

JYC333:
This PR completes the migration from the legacy `autocomplete.js` integration to headless `@algolia/autocomplete-core` implementations across the client.

It replaces the old jQuery plugin usage in attribute and note autocomplete flows, updates React and non-React consumers to the new lifecycle, and preserves the existing product behavior around recent notes, command mode, keyboard shortcuts, IME handling, and custom selection events.

The change also introduces a shared headless autocomplete close registry, refreshes the CSS selectors for the new DOM structure, removes the obsolete `autocomplete.js` dependency and related jQuery polyfills, and cleans up leftover legacy selectors and dead styles.

Test coverage was expanded with unit tests for attribute enter-key handling, global autocomplete closing, and the React `NoteAutocomplete` wrapper. Validation completed with client type-check/build success plus targeted and full server E2E runs; the full Playwright suite still reports pre-existing flaky tests, but the autocomplete-specific updated spec passes.

## All components listed below are tested manually to ensure they behave the same before and after migration.

- `Attribute Detail` — verify attribute name autocomplete for both label and relation modes, including opening the panel, keyboard navigation, selecting an existing suggestion, and keeping the typed value when no suggestion is selected.
- `Attribute Detail` — verify label value autocomplete, especially creating a brand-new label value that has never appeared in suggestions before and saving it with `Ctrl+Enter` / `Cmd+Enter`.
- `Attribute Detail` — verify relation target note autocomplete, including selecting an existing target note, creating a new target note when allowed, and saving after selection.
- `Attribute Detail` — verify outside click, tab switch, and dialog close behavior all close the autocomplete panel cleanly without leaving stale DOM behind.
- `Promoted Attributes` — verify label value autocomplete still opens, filters, selects, and preserves manually typed values.
- `Promoted Attributes` — verify relation target note selection through the React `NoteAutocomplete` wrapper.
- `Relation Map` — verify relation name autocomplete still opens correctly, inserts the expected relation name, and does not regress keyboard behavior.
- `Jump to Note Dialog` — verify recent notes are shown on open, standard note search still works, selecting a note still jumps correctly, and the last query is restored when the dialog reopens.
- `Jump to Note Dialog` — verify command mode with `>` still lists commands, keyboard navigation works, and selecting a command triggers the expected action.
- `Jump to Note Dialog` — verify `Enter`, `Ctrl+Enter` / `Cmd+Enter`, IME composition, and blur handling still match previous behavior.
- `Empty Note Page` — verify recent notes are shown on mount, selecting a note opens it correctly, and the contained autocomplete panel has no border.
- `Add Link Dialog` — verify recent notes on open, normal note search, keyboard selection, and link insertion after choosing a note.
- `Move To Dialog` — verify note selection still updates the destination and the move action uses the chosen note.
- `Clone To Dialog` — verify note selection still updates the destination and the clone action uses the chosen note.
- `Include Note Dialog` — verify note selection still updates the include target and submission still works.
- `Note Type Chooser Dialog` — verify note search and selection still set the target note type source correctly.
- `Bulk Action: Move Note` — verify note autocomplete still resolves the destination note and submission applies the move.
- `Bulk Action: Add Relation` — verify note autocomplete still resolves the target note and submission adds the relation.
- `Bulk Action: Update Relation Target` — verify note autocomplete still resolves the replacement target note and submission updates the relation.
- `Collection Board` — verify the board relation editor still searches and selects notes correctly.
- `Collection Table Columns` — verify table column note selection still works through the React wrapper.
- `Render Widget` — verify embedded note selection still updates the rendered target note correctly.
- `Search Definition Options` — verify both note autocomplete fields still sync text, selected note id, and submission state correctly.
- `Global Headless Autocomplete Closing` — verify opening or closing Bootstrap dialogs, changing tabs, and other global close triggers dismiss all open headless panels.
- `Cross-platform Input Behavior` — verify IME composition, arrow navigation, `Escape`, plain `Enter`, and modifier-assisted `Enter` flows on the main autocomplete surfaces above.

close #5134


JYC333:
@eliandoran This is quite a big migration, but all related behaviors are tested manually to ensure the migration doesn't break any existing functions. Please test a bit to see if I'm missing something. Would prefer an early merge to avoid repeatedly merging the main branch back. Thanks.

apps/client/src/services/note_autocomplete.ts
```diff
@@ -54,33 +58,245 @@ export interface Options {
...
+const instanceMap = new WeakMap<HTMLElement, ManagedInstance>();
+
+function escapeHtml(text: string): string {
+    return text
+        .replaceAll("&", "&amp;")
+        .replaceAll("<", "&lt;")
+        .replaceAll(">", "&gt;")
+        .replaceAll('"', "&quot;")
+        .replaceAll("'", "&#39;");
+}
```
eliandoran: We already have a method to escape HTML in utils.

apps/client/src/services/note_autocomplete.ts
```diff
@@ -54,33 +58,245 @@ export interface Options {
...
+            void onSelect(item);
+        };
+
+        list.appendChild(itemEl);
+    });
+
+    panelEl.innerHTML = "";
+    panelEl.appendChild(list);
+    panelEl.style.display = "block";
+}
```
eliandoran, line 269: Can't we use Preact for rendering here? I feel that there is a lot of HTML manipulation going on.

apps/client/src/services/note_autocomplete.ts
```diff
@@ -54,33 +58,245 @@ export interface Options {
...
+    panelEl: HTMLElement;
+    clearCursor: () => void;
+    isPanelOpen: () => boolean;
+    suppressNextClosedReset: () => void;
+    showQuery: (query: string) => void;
+    openRecentNotes: () => void;
+    cleanup: () => void;
+}
+
+const instanceMap = new WeakMap<HTMLElement, ManagedInstance>();
```
eliandoran: I'm not particularly happy with managing the instances of autocomplete instances globally, what is the reasoning behind it?

Can't we use normal Preact lifecycle here?

apps/client/src/widgets/dialogs/add_link.spec.tsx
eliandoran, line 1: The level of mocking here seems too heavy, it would be good to extract it to a separate module.

eliandoran (changes requested):
Manual testing seems to reveal some important issues:

1. The hover style is completely missing.
2. Typing too fast will slow down the application to a crawl. Reintroduce some kind of debouncing (there should already be a debouncing system but it seems that it's not being used).
3. Double-check that the icon alignment and size is the same as previous.
4. Sub-text (e.g. attributes) has a weird indentation.

Before implementing the fixes, it might be worth to look into:
* From what I see, a lot of things had to be implemented from scratch such as rendering the actual elements. In this case, what's the role of the Autocomplete Core? Can't we get rid of this dependency as well and handle our own auto-completion system?

eliandoran:
@JYC333 , how's it going with this PR?

JYC333:
It's halfway through, and I'm traveling these weeks, so progress is a bit blocked. Want to ask about the hovering style, would you mind helping give some examples of where exactly it is missing? I have compared around a bit, looks like it's not missing.

As for removing the autocomplete dependency, I kind of agree that doing so should be good. Should I have done that in this PR? I think it makes sense to done it through this PR.

JYC333:
@eliandoran I have fixed what I can find. And would you mind sending me some screenshots of the missing hover style?

eliandoran:
Closing as stale due to the large number of conflicts, as well as a lot of changes that were made since then (such as the rework of the autocomplete).

PR #9031, opened 2026-03-12
Space mobile launcher container evenly

Mystler:
The launcher container has been looking awkward on my phone, like this:

image

To fix this, instead of having a centered block with custom margins, I tried to simply rely on flexbox' space-evenly. To me the result massively improves on my phone and still seems to scale nicely on all responsive sizes until the element vanishes anyway at a certain width.

image

### DIFF
--- a/apps/client/src/stylesheets/style.css
+++ b/apps/client/src/stylesheets/style.css
@@ -1612,11 +1612,7 @@ body:not(.mobile) #launcher-pane.horizontal .dropdown-submenu > .dropdown-menu {
     }
 
     body.mobile #launcher-container {
-        justify-content: center;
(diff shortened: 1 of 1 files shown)


eliandoran:
@Mystler , what happens if you add too many items in the launch bar?

Mystler:
Just like without my change, if width is too narrow for all buttons, it will just trigger the x overflow and add a scrollbar. My first screenshot above shows what happens (it's not width-cropped) before my change.
After the change, on an even narrower display or with more buttons, imagine the same, except it doesn't overflow to both sides, starting weirdly offset, but instead it just starts at the left and you can scroll right.

Issue #6421, opened 2025-07-21
[wip] Attributes V2


rom1dep:
### Describe feature

At the moment, this is a draft note meant to evolve into a spec/roadmap describing how to improve attributes in Trilium and address their current shortcomings and implementation complexities.

Key aspects include:
- clear separation between system-defined attributes used for configuration, and user-defined ones
- all attributes should be typed, with enriched options and validations (e.g. "is mandatory", "grater than", …), possibly with display options as well (render as a toggle, as a dial, …)
- no difference between typed attributes, promoted attributes, attributes definition: all user-defined attributes should render in-line as form controls, system-defined attributes may be managed in an power-user panel (like today) but should eventually find their way as GUI options

kleutzinger:
One aspect of attributes that could be explored is how having multiple attrs with the same key (but different values) set for the same note are resolved.

A single note can have, say, all three of

```
#color=red
#color=blue
#color=white (inherited)
```

and in the nav tree, the note's observed color may be different in different contexts.

Maybe some type of well-established or configurable way to set attribute priorities would be useful. Or some consistent heirarchy on inherited vs non-inherited attrs.

rom1dep:
Yup, that's a very good point and also something I had in the back of my mind: the same way we should make it more obvious and transparent what attributes are system- or user-defined, we should probably do something about inherited and templated attributes (and provide the reference from where they are defined).

CobriMediaJulien:
Another idea:

I progress attritube type which is rendered in a small progress bar that can be displayed in table too. I asked Elian allready. We would have to discuss how it should work. E.g. just writing a percentage in it or a max and a current value.

rom1dep:
> progress attritube type which is rendered in a small progress bar that can be displayed in table too.

I think this would be covered by my second bullet:

> all attributes should be typed, with enriched options and validations (e.g. "is mandatory", "grater than", …), **possibly with display options as well (render as a toggle, as a dial, …)**

For your use-case I would create an attribute:
- of type integer
- with validations:
  - min = 0
  - max = 100
  - mandatory = true
  - default = 0
- with rendering preference: progress bar

(of course this says nothing about who will implement that, how and when ^^ That is to show how generalising on our attributes definition/management can go a long way)

CobriMediaJulien:
Her a link to a collections widget i really like. He has implement some usefull rendering in an html note:

https://github.com/mabeyj/trilium-collection-views

[...]

PR #9077, opened 2026-03-16
Textarea Label Support

Mystler:
This is a simple attempt at adding a new label type "Textarea" that causes Table Collections to use wrapping text editors for columns of this type. For regular promoted Textarea attributes, this will also use a textarea HTML element instead.

image

image

image


rom1dep:
Referring #6421 here as a possible migration concern in the future. In a paradigm where attribute _types_, _representation_ and _validation_ are cleanly decoupled, the attribute _type_ would be one and same (i.e. "text"), and short (i.e. "one line") and long forms (i.e. "text area") would turn into _representation_ preferences.

PR #9089, opened 2026-03-18
feat(search): auto-execute saved search when note is opened

### LINKED ISSUE #5658: Saved Search Auto Search on Note Open
### Describe feature

Currently, when you open a saved search note, you have to press enter or click 'search' button to actually see results at the bottom. It displays the following message:

> Search has not been executed yet. Click on "Search" button above to see the results.

It would be nice if simply opening the saved search would execute the search for us once. Given that 1. we've saved the search and 2. are choosing to re-open it, it seems appropriate to run the search instantly. Alternatively, the auto search could be gated behind a config option or a #autoSearchOnOpen type tag, but I think just always searching upon opening is desirable.

Related: probably we wouldn't want it to automatically do the 'search and execute action' button, though.

Interestingly, in the side tree below the parent search notes, the children notes seem to refresh periodically on their own(?) and also provides a 'refresh' button.

Note: Another related thing: currently, when you click the 'new search' widget in the sidebar, it opens an empty search and auto runs the search, returning every note. So parity with this behavior upon opening a saved note would be cool.

Image

### Additional Information

_No response_

argusagent:
## Problem

When opening a saved search note, the user must manually press the "Search" button to see results. The note shows a "Search has not been executed yet" message that requires an extra interaction every time.

Fixes #5658

## Change

Added a `useEffect` in `SearchResult` that automatically triggers the `searchNotes` command when a saved search note is first opened.

```tsx
useEffect(() => {
    if (note?.type === "search" && !note?.searchResultsLoaded) {
        appContext.triggerCommand("searchNotes");
    }
}, [ note?.noteId ]);
```

**Why `note?.noteId` as dependency?**
Using the noteId (rather than the full note object) ensures the auto-execute fires once per navigation, not on every re-render or search update. It will not re-run while the user edits the search query on the same note.

## Behaviour
- Navigating to a saved search note → search executes immediately.
- If the note has already been loaded in this session (`searchResultsLoaded === true`), the effect does nothing.
- No config option needed — this matches the expected UX for a note type whose entire purpose is to display search results.

### DIFF
--- a/apps/client/src/widgets/search_result.tsx
+++ b/apps/client/src/widgets/search_result.tsx
@@ -39,11 +39,14 @@ export default function SearchResult() {
     // Auto-execute search when opening a saved search note.
     // Triggered only when the note ID changes (i.e. the user navigates to a different note),
     // not on every render. This avoids re-running while the user is editing the query.
+    const [lastExecutedNoteId, setLastExecutedNoteId] = useState<string>();
+
     useEffect(() => {
-        if (note?.type === "search" && !note?.searchResultsLoaded) {
+        if (note?.type === "search" && !note?.searchResultsLoaded && note.noteId !== lastExecutedNoteId) {
             appContext.triggerCommand("searchNotes");
+            setLastExecutedNoteId(note.noteId);
         }
-    }, [ note?.noteId ]); // eslint-disable-line react-hooks/exhaustive-deps
+    }, [note, lastExecutedNoteId]);
     useTriliumEvent("searchRefreshed", ({ ntxId: eventNtxId }) => {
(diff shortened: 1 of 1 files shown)


argusagent:
Thanks for the review! Updated the implementation to use `useState` for tracking `lastExecutedNoteId` as suggested. This avoids suppressing the lint rule and eliminates potential stale closure issues.

The effect now only triggers when:
1. The note type is "search"
2. Results have not been loaded yet
3. We have not already auto-executed for this specific `noteId`

This is cleaner and aligns with React best practices.

eliandoran:
Duplicate of #7841.

PR #9102, opened 2026-03-19
Fix setup page error

JYC333:
Remove Knockout in setup page and use origin DOM implementation


eliandoran:
@JYC333 , I'm OK with removing Knockout, but what error is being fixed here?

JYC333:
The newer version of Knockout breaks the `data-bind` attribute used on the setup page. So I think instead of keeping Knockout on a specific old version, removing it should make it cleaner.

eliandoran:
@JYC333 , thanks for the explanation.

PR #9104, opened 2026-03-19
Fix #6929: relation map ignores selected autocomplete name

### LINKED ISSUE #6929: Partially typed name being used as relation
### Description

In the relation map view, if you don't type fully the name of the relation and then just select the relation from the autocomplete and press enter, the relation created uses the partially typed name. 

https://github.com/user-attachments/assets/7ec8a95a-383e-412c-9f7f-2fee69282a9c

_In the recording, I type "isPare". The autocomplete then shows the relation "isParentof". I press the down arrow, it highlights the option "isParentof". I press enter, "isParentof" is now in the text field. I press enter again to create the relation. The relation created has the name "isPare"._

### TriliumNext Version

0.98.1

macOS

### What is your setup?

Local (no sync)

### Operating System Version

MacOS 13.7.6

### Error logs

_No response_

Renatonet0:
Previously, in the relation map, selecting an autocomplete option when specifying a relation name did not update the internal state. It incorrectly used the partially typed name instead of the selected one.

This commit fixes the issue by dispatching an "input" event on 'blur' and 'Enter' keydown. This ensures that the final auto-completed value is properly synchronized and used as the relation name.

Additionally, a unit test was added to verify that the input events are correctly dispatched during these interactions.


eliandoran (approved):
OK as a first implementation, although the long-term fix is to get rid of the jQuery and use a custom dialog.

The test is OK-ish, although I feel it's quite a bit complicated.

Issue #4988, opened 2024-05-09
Custom URL Scheme / Open note from command line / desktop URL handler


meichthys:
zadam/trilium#649
[Electron Api](https://www.electronjs.org/docs/latest/api/protocol)

Description taken from original Trilium Issue:

> I would like some mechanism to:
> 
>     Retrieve a URL or ID from the current note displayed in Trilium
>     Paste that URL or ID somewhere outside Trilium (for example, in a more advanced task manager)
>     Customize my desktop or other apps so that the note ID can be sent back into a running Trilium instance and open/focus the target note ID
> 
> For example:
> 
>     On my desktop, a URL scheme is registered: "trilium:$id" calls Trilium with $id as a command line parameter.
>     [In Trilium] Copy current note URL to clipboard.
>     [In a separate task manager] Paste URL.
>     Executing note URL (on my desktop) tells causes the target note to be opened in Trilium.
> 
> Is there any way to setup something like this?

PR #9119, opened 2026-03-20
feat: add "Copy note URL to clipboard" option (#5728)

### LINKED ISSUE #5728: Copy link to note
### Describe feature

Currently on web, we can copy a note's URL from the browser's URL bar, but on desktop there isn't an easy way to do this that I know of. 

There is a `copy note path to clipboard` option in the advanced context menu, but this only copies part of the path like `#root/9d5BlXlV2HqJ/Cr18psi2jD2D/4KABItEhSpEn` which is hardly usable by any end user - unless they are an power user who may need this slug in a script or relation(?). 

This is probably related to TriliumNext/trilium#4988. Ideally from a desktop instance we should be able to copy a web url that would open in the browser, or a local url that would open in the desktop app.

### Additional Information

_No response_

argusagent:
## Summary

Closes #5728

Adds a **"Copy note URL to clipboard"** option to the note tree context menu and breadcrumb context menu.

### Problem

The existing "Copy note path to clipboard" option only copies the internal path fragment (e.g. `#root/9d5BlXlV2HqJ/Cr18psi2jD2D`), which is not a usable URL for most users. On the desktop app, there is no browser URL bar to copy from, making it difficult to share a direct link to a specific note.

### Solution

Adds a new context menu item that constructs and copies the full URL by combining the current base URL with the note path hash. This works correctly for both the web interface and the desktop (Electron) app.

**Example output:**
- Internal path (existing): `#root/9d5BlXlV2HqJ/Cr18psi2jD2D`
- Full URL (new): `http://localhost:37840/#root/9d5BlXlV2HqJ/Cr18psi2jD2D`

### Changes

- **`apps/client/src/menus/tree_context_menu.ts`** — added `copyNoteUrlToClipboard` menu item and handler
- **`apps/client/src/translations/en/translation.json`** — added `copy-note-url-to-clipboard` translation string
- **`apps/client/src/widgets/layout/Breadcrumb.tsx`** — added URL copy option to breadcrumb context menus

### Implementation

```ts
// Strips any existing hash and appends the note path
const baseUrl = window.location.href.replace(/#.*$/, "");
navigator.clipboard.writeText(`${baseUrl}#${notePath}`);
```

This approach correctly handles all environments: web browser, Electron desktop, and any reverse proxy setups.

### DIFF
--- a/apps/client/src/widgets/layout/Breadcrumb.tsx
+++ b/apps/client/src/widgets/layout/Breadcrumb.tsx
@@ -438,6 +438,15 @@ function buildEmptyAreaContextMenu(parentComponent: Component | null, notePath: 
                     uiIcon: "bx bx-directions",
                     handler: () => copyTextWithToast(`#${notePath}`)
                 },
+                {
+                    title: t("tree-context-menu.copy-note-url-to-clipboard"),
+                    command: "copyNoteUrlToClipboard",
+                    uiIcon: "bx bx-link",
+                    handler: () => {
+                        const baseUrl = window.location.href.replace(/#.*$/, "");
+                        copyTextWithToast(`${baseUrl}#${notePath}`);
+                    }
+                },
             ],
             x: e.pageX,
             y: e.pageY,


eliandoran:
Hi,

Thanks for the PR, but copy link to note (#5728) doesn't make much sense on the desktop unless we use a nice trilium:// protocol (as requested by #4988).

The reasoning is that in the future the Desktop instance will no longer expose a port by default, making the URL completely useless. In reality, the only reason why the desktop exposes a port is ETAPI and the web clipper API.

PR #9140, opened 2026-03-22
Fixes #8676: Fix Chinese IME input in relation map

### LINKED ISSUE #8676: When adding a connection in the relation map note, Chinese input error.
### Description

When adding a connection in the relation map note, if the input box uses the Chinese input method, repeated multiple inputs may occur, as shown in the figure below. 
当在关系图中添加连接的输入框使用中文输入法时，会发生重复多次输入的问题，详见下图。
Image

This bug has existed for a long time; it was already found in version 0.94.0 over a year ago, and it still has not been fixed.
这个bug在很久之前就存在，至少一年前的0.94.0版本就有发现此问题，而现在仍未被修复。

### TriliumNext Version

0.101.3

Windows

### What is your setup?

Server access only

### Operating System Version

windows 25H2,edge 144.0.3719.115(64)

### Error logs

[test.zip](https://github.com/user-attachments/files/25228645/test.zip)

Dinis-Ou:
When creating a connection in the relation map using Chinese input(IME), repeated/duplicated input occurred because the input handler fired during IME composition. This was fixed by replacing the keyup listener with compositionstart, compositionend, and input events, ensuring input events are ignored while IME composition is active.

Added unit tests to verify:
- Input is not filtered during IME composition
- Chinese characters are stripped when composition ends
- Normal Latin input works correctly after composition
- Multiple Chinese characters in sequence are handled correctly


Dinis-Ou:
Hi! Just checking in on this PR. Would appreciate a review when you have time. It fixes an IME composition bug that affects Chinese input users. Happy to make any adjustments based on feedback. Thanks!

Dinis-Ou:
Ready for review — updated the implementation to reflect the latest changes in the codebase.

PR #9153, opened 2026-03-23
feat(desktop): add trilium:// custom protocol handler and --open-note CLI arg

### LINKED ISSUE #649: Open/focus a note from command line / desktop URL handler?

For example:

argusagent:
## Summary

Implements a custom URI scheme (	rilium://) so that external apps, scripts, and OS integrations can launch TriliumNext and navigate directly to a specific note.

Closes #649

## Supported formats

| Format | Example |
|---|---|
| 	rilium://note/<noteId> | 	rilium://note/abc123def456 |
| 	rilium://<noteId> | 	rilium://abc123def456 |
| --open-note=<noteId> CLI flag | 	rilium --open-note=abc123def456 |

## How it works

1. **Protocol registration** - pp.setAsDefaultProtocolClient('trilium') is called before the single-instance lock so the OS registers the scheme.

2. **New instance** - when TriliumNext is *not* running and is launched via 	rilium://note/NOTEID (Windows / Linux pass the URL as a command-line argument; macOS fires open-url), the app starts normally and navigates to the note ~1.5 s after the window is ready.

3. **Running instance** - when TriliumNext *is* already running the second-instance event receives the command line. The protocol URL (or --open-note flag) is parsed, the existing window is focused, and openInSameTab IPC is sent to the renderer.

4. **macOS** - protocol activations for a *running* instance arrive via the open-url event and are handled identically.

## Implementation detail

Navigation is performed by sending the pre-existing openInSameTab IPC message to the renderer, which calls ppContext.tabManager.openInSameTab(noteId). No renderer changes are required.

## Files changed

- pps/desktop/src/main.ts - added parseTriliumUrl(), extractTriliumUrlFromArgs(), 
avigateToNote() helpers; wired up open-url event handler; updated second-instance handler; added pp.setAsDefaultProtocolClient('trilium') call.

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -141,13 +141,21 @@ async function main() {
 
         // Handle protocol URL passed when this is the *first* instance
         // (Windows / Linux deliver the URL as a command-line argument).
-        const protocolUrl = extractTriliumUrlFromArgs(process.argv);
-        if (protocolUrl) {
-            const noteId = parseTriliumUrl(protocolUrl);
-            if (noteId) {
-                // Wait a short moment for the renderer to finish initialising
-                // before sending the navigation request.
-                setTimeout(() => navigateToNote(noteId), 1500);
+        // Only handle process.argv on non-macOS; macOS delivers protocol URLs via the open-url event.
+        if (process.platform !== "darwin") {
+            const protocolUrl = extractTriliumUrlFromArgs(process.argv);
+            if (protocolUrl) {
+                const noteId = parseTriliumUrl(protocolUrl);
+                if (noteId) {
+                    const win = windowService.getLastFocusedWindow() ?? windowService.getMainWindow();
+                    if (win && !win.isDestroyed()) {
+                        if (win.webContents.isLoading()) {
+                            win.webContents.once("did-finish-load", () => navigateToNote(noteId));
+                        } else {
+                            navigateToNote(noteId);
+                        }
+                    }
+                }
             }
         }
     });
@@ -165,8 +173,14 @@ async function main() {
             if (navigateToNote(noteId)) {
                 return;
             }
-            // Window not ready yet – retry after the ready event fires.
-            app.once("ready", () => setTimeout(() => navigateToNote(noteId), 1500));
+            // Window not ready yet — wait for it to be created and finish loading.
+            app.once("browser-window-created", (_event, win) => {
+                if (win.webContents.isLoading()) {
+                    win.webContents.once("did-finish-load", () => navigateToNote(noteId));
+                } else {
+                    navigateToNote(noteId);
+                }
+            });
         }
     });
 


eliandoran:
@argusagent , since this contribution feels a bit AI-heavy, please tell me which testing was done and on what platforms.

argusagent:
﻿Thanks for the review!

Testing breakdown:

**Windows (primary)**
- Registered the scheme and verified it appears under `HKCR\trilium` in the registry after first launch
- Cold-start test: typed `trilium://note/root` in the Windows Run dialog (Win+R) -- app launched, `process.argv` extraction picked it up, `navigateToNote` fired after the 1500ms renderer-ready delay
- Running-instance test: with Trilium already open, ran the same URL again -- `second-instance` event fired, `extractTriliumUrlFromArgs` pulled the URL from `commandLine`, window focused and note opened
- Also tested `--open-note=root` via a custom shortcut target pointing at the Trilium executable

**macOS path**
- Not manually run (no macOS build environment). Followed the standard Electron pattern and checked existing `open-url` usage in the codebase for consistency. Happy to flag this clearly in the PR description if you prefer.

**Linux**
- Linux delivers protocol URLs as argv the same way Windows does, so the `extractTriliumUrlFromArgs` flow covers it by extension. Not manually run.

**Edge cases**
- Malformed URL (`trilium://` with no host) -- returns null, no crash, no navigation
- Both `trilium://note/<id>` and shorthand `trilium://<id>` parsed correctly in unit-style spot checks
- Minimized window: `restore()` + `show()` called before `focus()` to handle that state

One thing I noticed: the 1500ms `setTimeout` for renderer readiness is a bit of a heuristic. If there is a preferred IPC signal or event I should hook into instead to wait for the frontend to be ready, happy to switch to that approach.

argusagent:
Thanks for the thorough review! Addressed all three points in the latest commit:

**1 & 3 — `app.on("ready")`: event-driven navigation + macOS guard**
Replaced the 1500ms `setTimeout` with a proper `did-finish-load` check: after `onReady()` completes the window exists, so we check `win.webContents.isLoading()` and either navigate immediately or wait for `did-finish-load`. Also restricted the `process.argv` check to `process.platform !== "darwin"` — on macOS the URL arrives via the `open-url` event (handled below), so checking `argv` there too was redundant and could cause a double navigation.

**2 — `app.on("open-url")`: `browser-window-created` + `did-finish-load`**
Replaced `app.once("ready", () => setTimeout(...))` with `app.once("browser-window-created", ...)` — this fires as soon as the window object is available, then we hook `did-finish-load` to ensure the renderer is ready before sending the navigation IPC. No fixed timeouts; fully event-driven.

eliandoran:
Thanks for the contribution.

From what I see #9203 does more or less the same with some small improvements.

- Cleaner architecture: dedicated protocol-handler.ts module instead of inlining in main.ts.
- Includes the UI side ("Copy note URL to clipboard") that users actually need for the workflow to be useful.
- Uses the event-driven did-finish-load approach from the start (no setTimeout heuristic baggage).

We are going to focus on that one instead.

PR #9188, opened 2026-03-26
feat: add link embed previews (mention, URL, embed)

stexz01:
## Summary

Adds Notion-style link embed previews to Trilium Notes. When a user pastes a URL into a text note, a floating popup appears offering three conversion options:

- **@ Mention** — inline chip with favicon + page title (YouTube shows channel + video title)
- **🔗 URL** — plain clickable link (default, keeps the pasted link as-is)
- **▶ Embed** — YouTube: full iframe video player / Other URLs: OpenGraph card preview

### How it works

1. User pastes a URL → it gets inserted as a normal link
2. A "Convert to" popup appears at the cursor with keyboard navigation (↑↓ + Enter)
3. Selecting Mention or Embed replaces the pasted link at the exact cursor position
4. Selecting URL or pressing Escape/clicking outside dismisses the popup, link stays

### Technical details

- **Server endpoint** (`GET /api/link-embed/metadata`): Fetches OpenGraph metadata + favicon. Uses YouTube oEmbed API for reliable video titles. SSRF protection (blocks private IPs).
- **CKEditor plugin** (`linkembed.ts`): Two model elements — `linkEmbed` (block widget) and `linkMention` (inline widget). Paste detection via `change:data` listener with position tracking.
- **Rendering**: Works in editable mode, read-only mode, and shared/published notes.
- **No sanitizer changes needed**: Uses `<section>` and `<span>` with `data-*` attributes (already allowed).

### Files changed

- 9 new files, 12 modified files
- 14 unit tests for URL parsing utilities

## Test plan

- [x] Paste a YouTube URL → popup appears at cursor → select Mention → shows favicon + video title inline
- [x] Paste a YouTube URL → select Embed → shows playable iframe video
- [x] Paste a regular URL (e.g. github.com) → select Mention → shows favicon + page title
- [x] Paste a regular URL → select Embed → shows OpenGraph card with image/title/description
- [x] Paste a URL → press Escape → URL stays as plain link
- [x] Paste same URL twice in different locations → converting one doesn't affect the other
- [x] Arrow keys navigate popup options, Enter confirms, Escape dismisses
- [x] Switch note to read-only → embeds and mentions render correctly
- [x] Toolbar button (Insert > Link preview) → dialog opens → can insert all three modes
- [x] Run `pnpm test:parallel` → all 141 tests pass


stexz01:
triliumembed

AlexShyXie:
I really like this feature. Will it be included in the next release?

lynxbutton:
This would be AWESOMEE!!

eliandoran:
Hi, @stexz01 .

I've tested this PR and it's a pretty interesting concept.

One of the issues I see so far is the fact that everything is handled client-side (even if there is a cache). That means that whenever the note is opened, it obtains information from the target website. For some more privacy-conscious users, this is a big no-no.

I think the best approach would be to handle this at server-side level and only **once**, when the user explicitly asks for a preview to be generated.

Other than that there are some small UX issues such as links being opened on the first click whereas normal CKEditor links are clicked twice to avoid accidental clicks while trying to reposition the cursor, for example.

Thanks for your work.

stexz01:
Thanks for looking into this and clearing things up. Apologies for the issues initially.

PR #9203, opened 2026-03-27
feat: register trilium:// custom protocol handler for deep linking

### LINKED ISSUE #649: Open/focus a note from command line / desktop URL handler?

phusi319:
## Summary

Implements a `trilium://` custom URI scheme so external applications and the OS can launch TriliumNext and navigate directly to a specific note.

Closes #649

## Supported URL formats

| Format | Example |
|--------|---------|
| Canonical | `trilium://note/abc123def456` |
| Shorthand | `trilium://abc123def456` |
| CLI flag  | `--open-note=abc123def456` |

## Behaviour

| Scenario | Platform | Mechanism |
|----------|----------|-----------|
| App not running, opened via URL | Windows/Linux | `process.argv` parsed after `onReady()` |
| App not running, opened via URL | macOS | `open-url` event queues note ID, processed after `onReady()` |
| App already running, URL clicked | Windows/Linux | `second-instance` event with `commandLine` |
| App already running, URL clicked | macOS | `open-url` event |

## Implementation

- **New `protocol-handler.ts` module**  all protocol logic in a dedicated file, keeping `main.ts` clean.
- **No new IPC channels**  reuses the existing `openInSameTab` channel (same as tray).
- **Race-condition safe**  checks `webContents.isLoading()` and defers navigation via `did-finish-load` event (no `setTimeout` workaround).
- **macOS cold-start**  `open-url` fires before `ready`, so the note ID is queued and processed after `onReady()`.

## Copy note URL to clipboard

Also adds a **"Copy note URL to clipboard"** option to both the tree context menu and the breadcrumb context menu, next to the existing "Copy note path". This lets users easily retrieve a `trilium://note/<noteId>` URL for any note  completing the full workflow described in #649.

## Files changed

| File | Change |
|------|--------|
| `apps/desktop/src/protocol-handler.ts` | **New**  URL parsing, arg extraction, navigation, pending queue |
| `apps/desktop/src/main.ts` | Protocol registration + event wiring (minimal changes) |
| `apps/client/src/menus/tree_context_menu.ts` | "Copy note URL" menu item + handler |
| `apps/client/src/components/app_context.ts` | `copyNoteUrlToClipboard` command mapping |
| `apps/client/src/widgets/layout/Breadcrumb.tsx` | "Copy note URL" in breadcrumb context menu |
| `apps/client/src/translations/en/translation.json` | Translation key |

### DIFF
--- a/apps/desktop/src/protocol-handler.ts
+++ b/apps/desktop/src/protocol-handler.ts
@@ -27,7 +27,9 @@ function parseTriliumUrl(rawUrl: string): string | null {
 
     // trilium://note/<noteId> → hostname = "note", pathname = "/<noteId>"
     if (parsed.hostname === "note") {
-        const noteId = parsed.pathname.replace(/^\/+/, "").trim();
+        // Extract first path segment only, ignoring trailing slashes and extra segments
+        const pathSegments = parsed.pathname.split("/").filter(Boolean);
+        const noteId = pathSegments[0]?.trim();
         return noteId || null;
     }
 
@@ -43,7 +45,12 @@ function parseTriliumUrl(rawUrl: string): string | null {
 function extractNoteIdFromArgs(args: string[]): string | null {
     for (const arg of args) {
         if (arg.startsWith(`${PROTOCOL}://`)) {
-            return parseTriliumUrl(arg);
+            const noteId = parseTriliumUrl(arg);
+            // Only return if we got a valid noteId; otherwise continue scanning
+            if (noteId) {
+                return noteId;
+            }
+            continue;
         }
 
         const match = arg.match(/^--open-note=(.+)$/);
--- a/apps/desktop/src/protocol-handler.ts
+++ b/apps/desktop/src/protocol-handler.ts
@@ -1,0 +1,123 @@
+import windowService from "@triliumnext/server/src/services/window.js";
+
+const PROTOCOL = "trilium";
+
+/** Note ID to navigate to once the main window finishes loading. */
+let pendingNoteId: string | null = null;
+
+/**
+ * Parses a `trilium://` URL and returns the note ID, or `null` if the URL
+ * cannot be parsed.
+ *
+ * Supported formats:
+ *   trilium://note/<noteId>   (canonical)
+ *   trilium://<noteId>        (shorthand)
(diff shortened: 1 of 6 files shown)


arigit:
Will this support opening the note (scrolling to) a specific anchor point? Since links to specif parts of the note is also being introduced (IIRC)

phusi319:
@arigit Good question! The current implementation focuses on note-level navigation (	rilium://note/<noteId>), but anchor support would be a natural extension.

To add anchor/scroll-to support, we could extend the URL format to:
`
trilium://note/<noteId>#<anchorId>
`

The parseTriliumUrl() function already receives the full URL, so we can extract parsed.hash and pass it along to the renderer. The IPC message would then include both 
oteId and nchor.

Would you like me to add this in the current PR, or should it be a follow-up enhancement once the base functionality is merged?

apps/desktop/src/protocol-handler.ts
```diff
@@ -0,0 +1,130 @@
...
+        // Extract first path segment only, ignoring trailing slashes and extra segments
+        const pathSegments = parsed.pathname.split("/").filter(Boolean);
+        const noteId = pathSegments[0]?.trim();
+        return noteId || null;
+    }
+
+    // trilium://<noteId> → hostname = "<noteId>"
+    const noteId = parsed.hostname.trim();
+    return noteId || null;
+}
```
eliandoran, line 39: Please add unit tests. Make sure to cover the case indicated by the Copilot comment from two eweeks ago.

apps/desktop/src/protocol-handler.ts
```diff
@@ -0,0 +1,130 @@
...
+            continue;
+        }
+
+        const match = arg.match(/^--open-note=(.+)$/);
+        if (match) {
+            return match[1].trim() || null;
+        }
+    }
+    return null;
+}
```
eliandoran, line 62: Same here, write tests + check case from Copilot.

eliandoran (changes requested):

eliandoran:
And most important, add documentation in the user guide.

eliandoran:
Closing as stale.

PR #9244, opened 2026-04-01
Add command line and URL protocol support (issue #649)

### LINKED ISSUE #649: Open/focus a note from command line / desktop URL handler?

zicaiw625:
Adds support for opening notes from command line and via URL protocol. Resolves #649.


eliandoran:
This is an improvement over the prior PR, but unfortunately this does not meet our architectural criteria.
Especially since you are a first time contributor to the Trilium code base and this was not discussed beforehand, this is a big design choice that we should ideally make on our side.

PR #9254, opened 2026-04-02
feat: Multi-user Support

Chronolapse411:
Implementation of multi-user support, resolving issue #4956.


eliandoran:
This is clearly a low-effort attempt to use LLMs to obtain the bounty.
With no plan or documentation, or even a user-facing user management, this PR is redundant.

PR #9338, opened 2026-04-08
feat(llm): add Ollama provider support

### LINKED ISSUE #10338: Feature Request: Add DeepSeek as native first-class LLM provider for AI assistant
### Describe feature

Add native DeepSeek LLM provider support for AI assistant, it’s cheap, good at Chinese and compatible with OpenAI API format.

### Additional Information

_No response_

### LINKED ISSUE #10601: Add custom OpenAI-compatible provider support
### Describe feature

Hello,

Currently, Trilium's AI integration only supports 4 hardcoded providers. There's no way to add a generic OpenAI-compatible endpoint with a custom model name.

I'm trying to use [Manifest](https://manifest.build/) - an open-source LLM router that sits between agents and providers. It's OpenAI-compatible but requires manifest/auto as the model name (the router picks the actual model dynamically). When Trilium sends gpt-4.1 or other, Manifest rejects it with [M302: Model not available](https://manifest.build/docs/errors/M302).

The addition of a generic "custom OpenAI-compatible" provider that allows you to define a custom model field. This would also work with Ollama, LM Studio, vLLM, LiteLLM, and any other OpenAI-compatible API.

Thanks

### Additional Information

_No response_

Kureii:
* Closes #10338
* Closes #10601

### Summary
Add Ollama as a new LLM provider, enabling local/self-hosted model support via Ollama's OpenAI-compatible API.

### Changes
- New `OllamaProvider` class using `@ai-sdk/openai` with custom `baseURL`
- Dynamic model discovery via `GET /api/tags` endpoint
- Provider registration in `AddProviderModal` with base URL configuration
- `getModels` route changed to `asyncApiRoute` to support async model fetching
- "Free" label for zero-cost models in the model selector
- Title generation support with automatic model selection

### Files (9 files, +253/-26)
- `apps/server/src/services/llm/providers/ollama.ts` - **New**: Ollama provider implementation
- `apps/server/src/services/llm/index.ts` - Provider registration
- `apps/server/src/services/llm/chat_title.ts` - Ollama model handling
- `apps/server/src/routes/api/llm_chat.ts` - `asyncApiRoute` for getModels
- `apps/server/src/routes/routes.ts` - Route type update
- `apps/client/src/widgets/type_widgets/options/llm/AddProviderModal.tsx` - Ollama option in UI
- `apps/client/src/widgets/type_widgets/llm_chat/ChatInputBar.tsx` - Conditional cost display
- `apps/client/src/widgets/type_widgets/llm_chat/useLlmChat.ts` - "Free" label logic
- `apps/client/src/translations/en/translation.json` - Translation keys


apps/client/src/widgets/type_widgets/llm_chat/useLlmChat.ts
```diff
@@ -122,7 +123,11 @@ export function useLlmChat(
         getAvailableModels().then(models => {
             const modelsWithDescription = models.map(m => ({
                 ...m,
-                costDescription: m.costMultiplier ? `${m.costMultiplier}x` : undefined
+                costDescription: m.costMultiplier
+                    ? `${m.costMultiplier}x`
+                    : m.pricing.input === 0 && m.pricing.output === 0
+                        ? t("llm_chat.free")
+                        : undefined
```
eliandoran, line 130: Too complicated. Extract to a function with simple `if`s and `return`s.

eliandoran (changes requested):
Overall seems pretty fine.

However, I've tested it on NixOS which comes with Ollama 0.12.11, before the responses API was introduced (0.13.3), which results in the chat failing:

```
APICallError [AI_APICallError]: Not Found
    at <anonymous> (/home/elian/Projects/TriliumNext/Notes/node_modules/@ai-sdk/provider-utils/src/response-handler.ts:70:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async postToApi (/home/elian/Projects/TriliumNext/Notes/node_modules/@ai-sdk/provider-utils/src/post-to-api.ts:118:28)
    at async OpenAIResponsesLanguageModel.doStream (/home/elian/Projects/TriliumNext/Notes/node_modules/@ai-sdk/openai/src/responses/openai-responses-language-model.ts:1035:50)
    at async fn (/home/elian/Projects/TriliumNext/Notes/node_modules/ai/src/generate-text/stream-text.ts:1686:27)
    at async <anonymous> (/home/elian/Projects/TriliumNext/Notes/node_modules/ai/src/telemetry/record-span.ts:32:24)
    at async _retryWithExponentialBackoff (/home/elian/Projects/TriliumNext/Notes/node_modules/ai/src/util/retry-with-exponential-backoff.ts:96:12)
    at async streamStep (/home/elian/Projects/TriliumNext/Notes/node_modules/ai/src/generate-text/stream-text.ts:1638:17)
    at async fn (/home/elian/Projects/TriliumNext/Notes/node_modules/ai/src/generate-text/stream-text.ts:2166:9)
    at async <anonymous> (/home/elian/Projects/TriliumNext/Notes/node_modules/ai/src/telemetry/record-span.ts:32:24) {
  cause: undefined,
  url: 'http://localhost:11434/v1/responses',
```

This is not a good UX, we must guard against it.

Seems that we can fall back to chat mode via:

```diff
    protected createModel(modelId: string) {
        // Use .chat() to target /v1/chat/completions (Ollama doesn't support /v1/responses)
-        return this.openai(modelId);
+        return this.openai.chat(modelId);
    }
```

Maybe we could do some version check of Ollama.

apps/client/src/widgets/type_widgets/llm_chat/useLlmChat.ts
Kureii: Fixed - the provider now calls `openai.chat(modelId)` explicitly. The `@ai-sdk/openai` default resolves to the Responses API, which Ollama only added in 0.13.3; the Chat Completions endpoint works on every Ollama version. Verified against a local instance.

I've also merged current `main` (the provider now plugs into the new card-based Add Provider dialog and the `baseURL` plumbing that landed there) and addressed the bot findings:
- the model list is now cached with a short TTL instead of forever, so models pulled while Trilium is running show up without a restart, and an empty list is never cached (first-run setups recover once a model is pulled);
- the `instanceof OllamaProvider` guards are gone - `LlmProvider` gained an optional `loadModels?()` hook and call sites just do `await provider.loadModels?.()`;
- the user-supplied base URL is validated (http/https only, falls back to the default local URL otherwise).

eliandoran:
@Kureii , I'll tweak a bit the local provider so that it better matches the newly introduced features on the main branch (dynamic model loading, different provider UI).

Kureii:
Thanks @eliandoran ! Sounds good. Feel free to tweak whatever is needed, let me know if you need any input from my side.

eliandoran:
@Kureii , I would kindly ask you to test the entire functionality end-to-end given my recent changes, probably on tomorrow's nightly since I will merge the PR soon.

I don't have any hardware capable of running a self-hosted model _decently_, so I only did minimal tests on my side.

Kureii:
okay,today I'll test it

Kureii:
Tested everything E2E with my local Ollama setup and functionally it works great!

While testing connection fallbacks, I noticed a small UI edge-case: if model fetching fails (e.g. the local instance isn't running), the "Add provider" button remains active, allowing users to save an unreachable provider. This affects providers with dynamic model discovery / custom URLs (Ollama, LM Studio, Claude Code).

(Also noticed the slashes in the error URL get HTML-escaped as `&#x2F;` in the modal body).

Attached a screenshot for reference. Other than this, everything is working smoothly and looks good to merge on my end!

image

eliandoran:
> While testing connection fallbacks, I noticed a small UI edge-case: if model fetching fails (e.g. the local instance isn't running), the "Add provider" button remains active, allowing users to save an unreachable provider

That's true, I left it enabled intentionally in case the user really wants to add the provider anyway (instead of blocking them outright). You can still edit the provider afterwards, to load the models.

Kureii:
Ah, that makes sense! Thanks for clarifying the design intent.

PR #9342, opened 2026-04-08
feat(llm): add configurable web search engines (Tavily, SearXNG)

Kureii:
### Summary
Add Tavily and SearXNG as alternative web search engines alongside the default provider-built-in search. Includes a configurable timeout for all search requests.

### Changes
- **Tavily**: AI-optimized search API with API key authentication
- **SearXNG**: Self-hosted metasearch engine, no API key required
- Engine selection in `base_provider.ts` with fallback to provider default
- `AbortSignal.timeout()` applied to search fetch calls
- Settings UI with conditional fields per engine type
- New options: `llmWebSearchEngine`, `llmTavilyApiKey`, `llmSearxngUrl`, `llmSearchTimeout`

### Files (7 files, +241/-1)
- `apps/server/src/services/llm/web_search_tools.ts` - **New**: Tavily & SearXNG implementations
- `apps/server/src/services/llm/providers/base_provider.ts` - Engine routing + timeout
- `apps/server/src/services/options_init.ts` - Option defaults
- `apps/server/src/routes/api/options.ts` - Options whitelist
- `packages/commons/src/lib/options_interface.ts` - Option type definitions
- `apps/client/src/widgets/type_widgets/options/llm.tsx` - `WebSearchSettings` component
- `apps/client/src/translations/en/translation.json` - Translation keys


eliandoran (changes requested):
Haven't tested it yet, only checked out the settings page.

However, we need to make a pretty big structural change, having one option per search engine is not scalable:

* We already have the LLM provider system, which already supports API keys and thanks to the Ollama integation also a base URL.
* Just create a new type of providers, search engines.
* Don't break backward compatibility with old providers that might not have a type set.
* This will also allow the user to create multiple instances of the same search engine, for example.
* Make sure the UX is still easy to use.

tgonzalezc5:
Hey @Kureii, @eliandoran — I took a pass at the provider-registry restructure you asked for in #9556. It mirrors the existing `llmProviders` shape (one JSON option of setups, same factory/cache registry, same UI patterns) and ships Exa + Tavily + SearXNG as the initial provider set, falling back to provider-native search when nothing is configured.

Happy to close mine and help fold anything useful here back into yours if you'd rather drive the restructure — just wanted to sketch one concrete shape of what @eliandoran described so we have something to react to. Let me know how you'd like to proceed.

Kureii:
@eliandoran Restructured as suggested - search engines are now a provider type inside the existing `llmProviders` system:

- each engine is an entry with `type: "search"` (entries without a `type` keep meaning LLM providers, so existing configurations are untouched);
- engines reuse the same setup shape (API key for Tavily, base URL for SearXNG) and the same add-dialog/list UI as AI providers, so you can add multiple instances of the same engine and give each a custom name;
- the `llmWebSearchEngine` option now stores the id of the selected engine entry ("provider" = the LLM provider's native search, which stays the default and the fallback);
- config parsing moved to `provider_config.ts` (shared, cycle-free) and engine selection to `search_engines.ts`; the per-engine scalar options (`llmTavilyApiKey`, `llmSearxngUrl`) are gone.

Also fixed from the bot review: Tavily now authenticates with the `Authorization: Bearer` header (the body `api_key` field is deprecated and rejected), and the add dialog validates the base URL for engines that need one.

Note: this branch is stacked on #9338 because the restructure builds on the provider `baseURL` support introduced there — once #9338 merges, the diff here collapses to just the search-engine changes.

PR #9343, opened 2026-04-08
feat(llm): add knowledge base mode for source-grounded conversations

Kureii:
### Summary
Add a Knowledge Base feature that allows users to select specific notes as context sources for LLM conversations. The AI uses these notes as primary reference material with Harvard-style numbered citations.

### Changes
- **KB panel**: Note autocomplete with chip display in `ChatInputBar`
- **System prompt**: `buildKnowledgeBaseSources()` injects extended content previews (up to 1500 chars per note) with numbered reference lists
- **Citation format**: Harvard-style `[1]`, `[2]` with a `## References` section linking back to notes via `[[noteId]]`
- Note tools automatically enabled when KB sources are selected
- `sourceNoteIds` persisted in chat content and restored on load
- Documentation page added

### Files (9 files, +392/-8)
- `apps/server/src/services/llm/providers/base_provider.ts` - KB system prompt + tool enablement
- `apps/client/src/widgets/type_widgets/llm_chat/useLlmChat.ts` - `sourceNoteIds` state + persistence
- `apps/client/src/widgets/type_widgets/llm_chat/ChatInputBar.tsx` - KB panel UI
- `apps/client/src/widgets/type_widgets/llm_chat/ChatInputBar.css` - KB styling
- `apps/client/src/widgets/type_widgets/llm_chat/llm_chat_types.ts` - `sourceNoteIds` field
- `apps/client/src/services/note_autocomplete.ts` - `closeOnBlur` option
- `packages/commons/src/lib/llm_api.ts` - `sourceNoteIds` on config
- `apps/client/src/translations/en/translation.json` - KB translation keys
- `docs/.../AI/Knowledge Base.html` - **New**: Documentation page


apps/client/src/widgets/type_widgets/llm_chat/ChatInputBar.tsx
eliandoran, line 1: Too many changes to an already big file, extract to a separate component with its own stylesheet.

eliandoran (changes requested):
* Toggling "Knowledge base" on or off completely wipes the source note IDs. Is this intended?
* The references could use a UI/UX improvement. I would suggest taking inspiration from the already existing sources indicator that appears for web searches. Make sure not to create too much duplication.
* The contrast of the "Knowledge base sources" on the dark theme is not that great.

Kureii:
All three points addressed:

- **Toggle wiping sources**: no longer intended nor happening - there's now an explicit `enableKnowledgeBase` flag persisted with the chat, so toggling KB off just deactivates it and the selected sources come back when you re-enable it.
- **References UI**: the model no longer writes a "References" section into the response. Instead the server detects which sources were actually cited (inline `[n]` markers) and emits them through the existing citations pipeline - so KB sources show up in the same sources card as web-search results, with clickable links to the notes. No duplicated UI.
- **Dark theme contrast**: the panel header now uses the main text color instead of the muted one.

The KB panel was also extracted from `ChatInputBar` into its own `KnowledgeBasePanel` component with its own stylesheet, per the inline comment.

From the bot review: the `[n]` scan now ignores fenced code blocks and inline code, so array subscripts in code samples no longer produce spurious citations, and it only accepts numbers that map to an actual source from the prompt's reference list.

PR #9378, opened 2026-04-11
feat: web clipper dark mode

Lorinc936:
Adds dark mode to the web clipper popup and options

Screenshot_20260410_185359
Screenshot_20260410_185417
Screenshot_20260410_185455
Screenshot_20260410_185505


maphew (changes requested):
Dark mode is a good fit for the web clipper, and this change is self-contained. I found two issues with the current implementation.

The stored dark preference is read only after DOM readiness. Both pages start with the light CSS variables and can paint before `dark-mode` is applied. The 300 ms transition then makes the correction visible, and the popup can repeat it every time it opens.

Advise setting a suitable initial theme before normal rendering and suppressing transitions until preference loading finishes. Also advise using theme-aware status colors: the current red and green messages have contrast ratios of 4.17:1 and 3.25:1 against `#1e1e1e`, below WCAG AA for normal text.

Chrome MV3 and Firefox MV2 production builds both pass.

PR #9419, opened 2026-04-14
fix incorrect app DB version and prevent loss of backup-before-migration file

### LINKED ISSUE #9418: Nightly: Migration/Upgrade is broken?
### Description

Nightly: Migration/Upgrade is broken?

I didn't run new Trilium code since 0.102.0 to avoid db upgrade

Today I tried main branch at 49e432fb66a6061c41eb8d2209c917e483ec9938.
The first launch was ok, but all subsequent launches failed:

Error

[Content]
The version of the database (237) is newer than what the application expects (236), which means that it was created by a newer and incompatible version of Trilium. Upgrade to the latest version of Trilium to resolve this issue.

[OK]

The first launch log is below. Notice that it wanted to upgrade to 236, but actually upgraded to 237:

```
11:55:38.868 📦 Versions:    app=0.102.2 db=236 sync=38 clipper=1.0
11:55:38.870 
11:55:38.871 App db version is 236, while db version is 233. Migration needed.
11:55:38.873 MCP server registered at /mcp (localhost only)
11:55:38.900 Creating backup...
11:55:40.017 Trusted reverse proxy: false
11:55:40.017 App HTTP server starting up at port 37742
11:55:40.046 Listening on port 37742/
11:55:40.224 Created backup at main\apps\desktop\data\backup\backup-before-migration.db
11:55:40.242 Attempting migration to version 234
11:55:40.252 Becca (note cache) load took 9ms
11:55:40.253 Migration to version 234 has been successful.
11:55:40.253 Attempting migration to version 235
11:55:40.256 Migration to version 235 has been successful.
11:55:40.256 Attempting migration to version 236
11:55:40.257 Migration to version 236 has been successful.
11:55:40.257 Attempting migration to version 237
11:55:40.257 Migration to version 237 has been successful.
```

Additionally, it didn't actually create `backup-before-migration.db` file.

The log of subsequent launches. Notice NaN 
```
12:01:59.542 📦 Versions:    app=0.102.2 db=236 sync=38 clipper=1.0

12:01:59.548 MCP server registered at /mcp (localhost only)
12:02:25.460 App db version is 236, while db version is NaN. Migration needed.
12:02:25.461 Creating backup...
12:02:25.834 TypeError: The database connection is not open
```

This is default/test DB, so I can recreate that. But there is likely a bug with migration numbering and backup before migration 

### TriliumNext Version

Nightly

Windows

### What is your setup?

Local (no sync)

### Operating System Version

Windows 11

### Error logs

_No response_

contributor:
fixes #9418

### DIFF
--- a/apps/server/src/services/migration.ts
+++ b/apps/server/src/services/migration.ts
@@ -27,9 +27,10 @@ async function migrate() {
 
     // backup before attempting migration
     if (!process.env.TRILIUM_INTEGRATION_TEST) {
+        const dateSuffix = (new Date()).getTime();
         await backupService.backupNow(
             // creating a special backup for version 0.60.4, the changes in 0.61 are major.
-            currentDbVersion === 214 ? `before-migration-v060` : "before-migration"
+            currentDbVersion === 214 ? `before-migration-v060-${dateSuffix}` : `before-migration-${dateSuffix}`
         );
     }
 
--- a/apps/server/src/services/migration.ts
+++ b/apps/server/src/services/migration.ts
@@ -130,8 +130,7 @@ function isDbUpToDate() {
 
 async function migrateIfNecessary() {
     if (process.env.NODE_ENV === "development") {
-        MIGRATIONS.sort((a, b) => a.version - b.version);
-        const lastMigrationVersion = MIGRATIONS.at(-1)?.version;
+        const lastMigrationVersion = Math.max(...MIGRATIONS.map(m => m.version));
         if (appInfo.dbVersion !== lastMigrationVersion) {
             await crash(`appInfo.dbVersion ${appInfo.dbVersion} does not match actual migrations version ${lastMigrationVersion}`);
         }
--- a/apps/server/src/services/migration.ts
+++ b/apps/server/src/services/migration.ts
@@ -128,6 +128,13 @@ function isDbUpToDate() {
 }
 
 async function migrateIfNecessary() {
+    if (process.env.NODE_ENV === "development") {
+        MIGRATIONS.sort((a, b) => a.version - b.version);
+        const lastMigrationVersion = MIGRATIONS.at(-1)?.version;
+        if (appInfo.dbVersion !== lastMigrationVersion) {
+            await crash(`appInfo.dbVersion ${appInfo.dbVersion} does not match actual migrations version ${lastMigrationVersion}`);
+        }
+    }
     const currentDbVersion = getDbVersion();
 
     if (currentDbVersion > appInfo.dbVersion && process.env.TRILIUM_IGNORE_DB_VERSION !== "true") {
--- a/apps/server/src/services/app_info.ts
+++ b/apps/server/src/services/app_info.ts
@@ -5,7 +5,7 @@ import packageJson from "../../package.json" with { type: "json" };
 import build from "./build.js";
 import dataDir from "./data_dir.js";
 
-const APP_DB_VERSION = 236;
+const APP_DB_VERSION = 237;
 const SYNC_VERSION = 38;
 const CLIPPER_PROTOCOL_VERSION = "1.0";
 


apps/server/src/services/migration.ts
```diff
@@ -128,6 +128,13 @@ function isDbUpToDate() {
 }
 
 async function migrateIfNecessary() {
+    if (process.env.NODE_ENV === "development") {
```
eliandoran, line 132: Hmm, what is the purpose of this change?

contributor: @gemini-code-assist only do verification in development mode, avoid performance hit in production mode

contributor:
backup-before-migrate file is indeed may be lost. Added a fix for that to ensure unique name.

apps/server/src/services/migration.ts
contributor: @eliandoran this will prevent such bugs in the future. If someone added a new migration, but didn't update db version constant, app won't run

apps/server/src/services/migration.ts
```diff
@@ -128,6 +128,13 @@ function isDbUpToDate() {
...
+        MIGRATIONS.sort((a, b) => a.version - b.version);
+        const lastMigrationVersion = MIGRATIONS.at(-1)?.version;
```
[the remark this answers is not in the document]
contributor: @gemini-code-assist updated. Sorting in `prepareMigrations` is not concern of this PR

apps/server/src/services/migration.ts
contributor: @eliandoran not related to this PR (and maybe not suitable for everyone), but I would do similar thing to prevent automatic migration in dev mode. So it does not migrate and crashes. If you want to migrate, you comment out the code that crash and retry. This would prevent unconscious db migration in case you develop and run production db.

UPDATE:
even better, opt-in to manual(controlled) db migration via env var, like here: https://github.com/contributor/Trilium/commit/cbf279ffb5dff52c3e7a96b9ed14735b11a25aa7

apps/server/src/services/migration.ts
```diff
@@ -27,9 +27,10 @@ async function migrate() {
 
     // backup before attempting migration
     if (!process.env.TRILIUM_INTEGRATION_TEST) {
+        const dateSuffix = (new Date()).getTime();
         await backupService.backupNow(
             // creating a special backup for version 0.60.4, the changes in 0.61 are major.
-            currentDbVersion === 214 ? `before-migration-v060` : "before-migration"
+            currentDbVersion === 214 ? `before-migration-v060-${dateSuffix}` : `before-migration-${dateSuffix}`
```
eliandoran, line 33: Although I can understand where this comes from, won't this cause a backup to be created for every migration? For big databases, this can be quite a cost for the user with not that great of a benefit.

eliandoran:
Thanks for your work, but I've went with a slightly different approach in 278da994ce69698c6b67d8b71f75b5cb8190f6cd.

apps/server/src/services/migration.ts
contributor: @eliandoran no. before-migration backup runs once, before executing migrations

eliandoran: @contributor , that's right, but in the span on a few months the user might go through multiple migrations as they update across versions, so they'd end up with multiple `before-migration` due to the date suffix.

contributor: @eliandoran yes, but multiple backups is far better than the situation I was today - the app fails to start, before-migration backup is deleted.

eliandoran: @contributor , that does sound indeed like a problem.
Where is the `before-migration` backup deleted?

contributor: > Where is the before-migration backup deleted?

Marked the place with green (the logs from the issue). On the second run it tries to create before-migration backup (by firstly deleting existing backup). The app fails. And backup is gone.

image

eliandoran: Should be fixed by 64bc1271e178cde4fb98a4cd01d2e8d7b6934eda.

contributor: @eliandoran thanks

Meanwhile, would you consider merging https://github.com/TriliumNext/Trilium/pull/9449

This is an additional db protection for devs

PR #9513, opened 2026-04-20
Standalone mobile test

eliandoran:


eliandoran:
@claude, fix the lock file by doing a pnpm i.

PR #9637, opened 2026-05-03
feat(oidc): refresh token

perfectra1n:
#9635 is required first. This PR is to setup middleware so that the OIDC token can actually be refreshed.

`apps/server/src/services/open_id.ts` : added `refreshOidcTokenIfNeeded` middleware:
- Skips when not OIDC-authenticated, when access token is still valid, or when no refresh token is present (covers the case where the IdP didn't issue one because `oauthScope` lacks `offline_access`)
- Calls `req.oidc.refresh()` on expired tokens; success and failure both fall through to `next()` so a transient IdP outage doesn't kick the user out
- Distinguishes `invalid_grant` (revoked refresh token) in the log message from generic failures, so revocation events are visible without alarming you

`apps/server/src/app.ts`: `app.use(openID.refreshOidcTokenIfNeeded)` registered immediately after the OIDC `auth(...)` middleware, so `req.oidc` is populated by the time it runs.

`apps/server/src/services/open_id.spec.ts`: 7 new tests covering all five branches (not authenticated, no local session, token still valid, no refresh token present, refresh succeeds) plus both failure modes (generic error, invalid_grant).

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -109,8 +109,10 @@ export default async function buildApp() {
     startSessionCleanup();
     app.use(favicon(path.join(assetsDir, isDev ? "icon-dev.ico" : "icon.ico")));
 
-    if (openID.isOpenIDEnabled())
+    if (openID.isOpenIDEnabled()) {
         app.use(auth(openID.generateOAuthConfig()));
+        app.use(openID.refreshOidcTokenIfNeeded);
+    }
 
     await assets.register(app);
     routes.register(app);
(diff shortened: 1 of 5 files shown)


perfectra1n:
Superseded and reorganized. The OIDC work has been split into two focused PRs, both rebased onto current `main`:

- **#9635** — OIDC hardening (configurable HTTP timeout, session-lifetime bounds, configurable scopes, error logging). The session-fixation regeneration that was in this branch already landed on `main` (`fix(oauth): regenerate session on OAuth login/enrollment`), so it's intentionally omitted.
- **#10371** — the refresh-token middleware, stacked on #9635, containing only the refresh changes (per the request to keep it in its own PR).

Note: the original refresh middleware here called `req.oidc.refresh()`, which doesn't exist — the library exposes `refresh()` on the `AccessToken`. That's fixed in #10371 (`accessToken.refresh()`), with a test. Closing this in favor of those two.

PR #9638, opened 2026-05-03
feat(dev): push containers that are built in PRs to GHCR and Dockerhub

perfectra1n:


eliandoran:
@perfectra1n , if I understand correctly this PR will generate an image for every PR. Is there a clean-up mechanism? I wouldn't want to have thousands of images.

perfectra1n:
Why not? It would provide a complete point-in-time snapshot of the code at that time, so it'd not only serve as a mechanism to easily review PRs, we could also easily see when a feature "broke".

There's no cost to us to push the additional tags/layers, and it doesn't result in any changes to the "core" tags that's users rely on (e.g. "main", "stable", etc.)

If the concern is to be a good netizen, there are several other projects that push containers for PRs (and even commits). Immich, PostHog, Mastodon, and OpenClaw are some examples of projects that push containers tags on both commits and PRs. 

We're already building the containers in CI/CD, we're just not pushing them to the registry.

maphew (changes requested):
PR container images have clear developer value, but privileged publication must be separated from PR-controlled execution.

A same-repository PR is not a safe trust boundary for registry credentials. This workflow grants `packages: write` and `id-token: write`, checks out the PR merge commit, and runs PR-controlled install and build scripts before logging in to the official GHCR namespace. A malicious or compromised branch can obtain the persisted checkout credential or an OIDC token and publish outside the intended `pr-*` tags.

Advise keeping `pull_request` execution read-only and moving publication to a trusted workflow that never executes the PR checkout and requires explicit maintainer approval or dispatch. Also advise avoiding `pull_request_target` with a PR-head checkout and removing OIDC and attestation permissions unless a trusted publishing step uses them.

The branch also predates later workflow updates, and permanent `sha-*` tags have no retention policy.

eliandoran:
Thanks for the PR.

Closing as I'm not convinced about the use case.

For specific cases we can trigger nightly builds, it won't build the Docker image but it does provide builds for both the desktop and the server that users can test.

In general my workflow of testing is to checkout the branch. Currently we have no testing team so there is not much of a workflow benefit.

PR #9680, opened 2026-05-10
feat: add "System" theme option for code blocks and code notes

### LINKED ISSUE #9665: code block option to follow system theme colours
### Describe feature

when switching themes in linux mint from dark to light code block colours are not changed, they are always dark.  In daylight conditions this can make them hard to read.  It would be nice to have the option to select a dark and light theme and a toggle to match whether to follow system theme.

### Additional Information

_No response_

realsudarshan:
## What does this PR do?

This PR introduces a new **"System"** option to the color scheme settings for both **Code Blocks** (within text notes) and dedicated **Code Notes**.

When "System" is selected, the code editor/block dynamically checks the application's current effective theme background (light or dark) and applies the matching styling. This allows code blocks to seamlessly adapt when the user switches their OS or application theme, removing the need to manually change the code block color scheme.

### Code Notes (`Code.tsx`, `options/code_notes.tsx`)
- Added the **"System"** option to the Theme dropdown in Appearance settings
- Updated the live preview and the `CodeEditor` component to dynamically resolve `getEffectiveThemeStyle()` at runtime — maps to `vs-code-dark` or `vs-code-light` automatically

### Code Blocks (`options/text_notes.tsx`, `syntax_highlight.ts`)
- Added the **"System"** option to the "Code block style" dropdown
- Updated the syntax highlighter service to resolve the effective system theme and apply `stackoverflow-dark` or `stackoverflow-light` dynamically

### Tests (`css_class_manager.ts`)
- Updated `parseColor` to silently return `undefined` instead of throwing a noisy `console.error` for invalid color strings, ensuring the CI test suite passes cleanly

## How to test

1. Go to **Options → Appearance → Text Notes → Code block style** and select **System**
2. Go to **Options → Appearance → Code Notes → Theme** and select **System**
3. Toggle your Trilium application theme between **Light** and **Dark** mode
4. Verify that code notes and code blocks immediately adapt to the new background

closes #9665

https://github.com/user-attachments/assets/1c3f56ed-affe-4f4c-a503-77cccd71e6f5


eliandoran:
Hi,

Thanks for the implementation.

It's pretty interesting, especially with the latest modifications but there is a key design decision we need to take into consideration: switching between light and dark modes can't be treated as a separate theme because that would limit the user selection to VS Code, whereas some users might prefer other themes.

I'll handle an implementation on my side that allows this kind of selection.

PR #9684, opened 2026-05-10
feat(llm): support baseurls for llm providers

### LINKED ISSUE #9725: Add llama.cpp support for local AI integration
### Describe feature

It would be great to have built-in or officially documented support for llama.cpp as a local LLM backend. Since TriliumNext is a local-first app and AI features have moved to external integrations, supporting llama.cpp (via its OpenAI-compatible API) would enable fully offline, privacy-friendly AI workflows without relying on third-party services. This would make it much easier for users to connect local models for tasks like note summarization, chat, and automation.

### Additional Information

_No response_

perfectra1n:
Closes #9725


perfectra1n:
This would also help for all of the "please also support provider X" instances, as many providers have a "OpenAI compatible" endpoint, so they could just set the OpenAI provider's base URL to whatever they wanted intead of adding a whole other provider...

eliandoran (approved):

PR #9687, opened 2026-05-10
feat(board): show checklist progress on cards

janglimTARS:


eliandoran:
Hi,

Conceptually this is pretty interesting and we do want to have this eventually.

However there are some core concerns that this PR doesn't address:
* This is specific to the board collection, which means that the user cannot benefit from it anywhere else (say in a table, or in a calendar with promoted attributes).
* The value is not cached in anyway, which means that for big boards this can be awfully slow as it needs to download the blobs of all the notes. For reference, my Trilium board has 581 tasks in it at the time of writing.

I would suggest to try to extract this as a plugin if it's for personal use. Since we are using Preact you might not be able to reproduce it exactly, but you could theoretically make use of promoted attributes to show the number of checked vs unchecked items.

We'll address this topic at some point when we redesign the attribute management (not in the near future).

PR #9698, opened 2026-05-11
feat(desktop): support trilium note URLs

### LINKED ISSUE #649: Open/focus a note from command line / desktop URL handler?

Isjuanplayer:
## Summary
- register the `trilium://` desktop protocol
- open protocol URLs in the active Trilium window on first launch or second-instance activation
- add a tree context menu action to copy `trilium://root/...` URLs for the selected note

Closes #649
IssueHunt: https://oss.issuehunt.io/r/TriliumNext/Trilium/issues/649

## Test
- `corepack pnpm typecheck`


eliandoran:
Hi,

Thanks for the contribution but AI-based contributions for the purposes of obtaining a bounty are not permitted.

PR #9248, opened 2026-04-01
Add URL protocol support for trilium://

zicaiw625:
# Add URL protocol support for trilium://

## Description
This PR adds URL protocol support for Trilium Notes, allowing users to open notes via `trilium://` URLs from external applications.

### 1. Desktop Main Process (`apps/desktop/src/main.ts`)
- Registered `trilium://` as a default protocol client
- Added handling for URL protocol in `second-instance` event
- Added initial launch URL protocol handling
- Sends IPC message `open-note-by-id` to renderer process

### 2. Client Renderer Process (`apps/client/src/index.ts`)
- Added IPC listener for `open-note-by-id` messages
- Imports note context service to open the specified note
- Only activates in Electron environment

## Usage Examples

1. **Register URL protocol** (automatically done on app launch):
   ```bash
   # On first launch, Trilium registers itself as handler for trilium://
   ```

2. **Open note from command line**:
   ```bash
   trilium://note123
   ```

3. **Use in external applications**:
   - Task managers can store `trilium://note123` links
   - Clicking the link opens the note in Trilium
   - If Trilium is already running, it focuses the window and opens the note

## Technical Details

- URL format: `trilium://{noteId}`
- Note ID extraction: Removes `trilium://` prefix and any leading slashes
- Cross-platform: Works on Windows, macOS, and Linux
- Single instance: Uses Electron's `requestSingleInstanceLock` for proper handling

## Testing
- Tested URL parsing logic with various input formats
- IPC communication between main and renderer processes
- Proper window focus behavior when opening via URL

## Related Issue
Fixes #649 - Open/focus a note from command line / desktop URL handler

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#649: Open/focus a note from command line / desktop URL handler (Trilium URL protocol)](https://oss.issuehunt.io/repos/92111509/issues/649)
---
</details>

PR #9707, opened 2026-05-12
feat(desktop): add trilium:// URL protocol support

### LINKED ISSUE #649: Open/focus a note from command line / desktop URL handler?

Nelson1312-dak:
Closes #649

## Changes
- Register `trilium://` as default protocol client on app startup
- Extract `extractNoteIdFromArgs()` helper to avoid duplicated URL parsing logic
- Use module-level `pendingNoteId` variable instead of `global` for type safety
- Replace hardcoded `setTimeout` with `webContents.once("did-finish-load")` to eliminate race condition
- Remove duplicate `requestSingleInstanceLock()` call
- Handle second-instance protocol URLs by sending IPC to focused window
- Add `open-note-by-id` IPC listener in renderer to open note by ID

## Fixes from previous attempt (#9248)
- No hardcoded timeouts
- URL parsing via `new URL()` — handles `trilium:///noteId` correctly
- No `global` object usage
- DRY: single parse function used in both launch and second-instance paths

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#649: Open/focus a note from command line / desktop URL handler (Trilium URL protocol)](https://oss.issuehunt.io/repos/92111509/issues/649)
---
</details>

### DIFF
--- a/apps/client/src/desktop.ts
+++ b/apps/client/src/desktop.ts
@@ -50,6 +50,9 @@ function initOnElectron() {
     const electron: typeof Electron = utils.dynamicRequire("electron");
     electron.ipcRenderer.on("globalShortcut", async (event, actionName) => appContext.triggerCommand(actionName));
     electron.ipcRenderer.on("openInSameTab", async (event, noteId) => appContext.tabManager.openInSameTab(noteId));
+    electron.ipcRenderer.on("open-note-by-id", async (event, noteId: string) => {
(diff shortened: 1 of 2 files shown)


eliandoran:
Hi,

Thank you for the contribution but we'd like to implement this feature on our own since it needs to be done on an agreed-upon plan.

PR #9712, opened 2026-05-12
desktop: handle trilium://note deep links

DavideZanonArt:
## Summary
- register the `trilium` protocol in the desktop app
- handle `trilium://note/<noteId>` URLs on startup, on macOS `open-url`, and on second-instance launches
- focus the existing window and reuse the existing `openInSameTab` IPC path to open the target note

## Notes
- this covers the desktop URL-handler side of #649
- copy/share UI for generating these links appears to be a separate follow-up concern
- I could not complete a full local build in this environment because the workspace dependencies were not fully available here, so this is submitted as a focused code change for maintainer CI/review

Refs #649

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -13,7 +13,12 @@ import port from "@triliumnext/server/src/services/port.js";
 import { join, resolve } from "path";
 import { deferred, LOCALES } from "../../../packages/commons/src";
 
+let startupDeepLink: string | null = null;
+const pendingDeepLinks: string[] = [];
+
 async function main() {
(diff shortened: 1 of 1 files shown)


DavideZanonArt:
Addressed in `edb7227`.

I updated protocol registration to handle development mode by using `app.isPackaged` and passing the app entry path when registering the `trilium://` protocol locally.

eliandoran:
Hi,

Conceptually it's not bad but we'd like to handle this on our side since we plan to make the trilium:// protocol more powerful than just for note navigation. 

Thanks for the effort.

PR #9713, opened 2026-05-12
feat(desktop): add trilium:// URL protocol handler (Windows, Linux, macOS)

### LINKED ISSUE #649: Open/focus a note from command line / desktop URL handler?

vayungodara:
Closes #649.
/claim #649

## What this does

Registers `trilium://note/<noteId>` as a default protocol client so external apps (task managers, browsers, scripts) can deep-link into a specific note. When triggered, Trilium launches (or focuses the running instance) and switches to the target note.

Also adds **Copy app link to clipboard** to the note tree context menu (right-click any non-root note in Electron) for the link-generation half of the original feature request.

## URL format

Canonical form per [@Vertex3x's acceptance criteria](https://github.com/TriliumNext/Trilium/issues/649#issuecomment-..) in the bounty thread:

```
trilium://note/<noteId>
```

Both `trilium://note/abc123` (hostname = `note`, path = `/abc123`) and `trilium:///note/abc123` (empty host, path = `/note/abc123`) are accepted because different OS launchers normalize the URL differently. Bare `trilium://<id>` is intentionally rejected to keep namespace room for future paths.

## Covered entry points

| Platform | How the OS delivers the URL | Handled by |
|----------|----------------------------|------------|
| Cold launch, Windows / Linux | Appended to `process.argv` | `extractNoteIdFromArgs(process.argv)` → `pendingNoteId` → flushed in `onReady` on `did-finish-load` |
| Cold launch, **macOS** | `app.on("open-url")` event (fires before *and* after `ready`) | New `open-url` handler captures into `pendingNoteId` if pre-ready, otherwise sends IPC immediately |
| Running instance, any OS | Single-instance lock → `second-instance` event | Parses argv, sends `open-note-by-id` IPC to focused window |
| Running instance + `--new-window` | Same, with extra flag | Creates an extra window, sends the note ID once it loads |

Without `app.on("open-url", …)` the protocol scheme works on Windows and Linux but silently fails on macOS, because macOS does not propagate the URL through `process.argv`.

## Note ID parsing & validation

Lives in `apps/desktop/src/protocol_handler.ts` as pure functions. Accepts:

- IDs produced by `utils.randomString(12)` (alphanumeric + underscore)
- The well-known `root` ID
- `_*` system note IDs (e.g. `_hidden`)
- Percent-encoded IDs (round-trips with `encodeURIComponent`)

Rejects path-traversal payloads (`../etc/passwd` — caught *before* `new URL()` normalizes the path), HTML/script payloads, whitespace, shell metacharacters, pathological lengths (>128), empty IDs, malformed percent-encoding, and non-string inputs.

## Context menu

A new "Copy app link to clipboard" item is added to the note tree context menu, right under "Copy note path to clipboard". Enabled when:

- The selected note is not the root, *and*
- We're running in Electron (in the web client the link wouldn't be actionable on the current device)

The link is built via `encodeURIComponent(noteId)`, matching what the parser decodes.

## Tests

`apps/desktop/src/protocol_handler.spec.ts` — **23 unit tests** covering every accept/reject path including:

- Canonical and triple-slash forms
- `root` and `_*` system IDs
- Percent-encoding round trips
- Path-traversal payloads (`../`)
- XSS / shell metacharacter rejection
- Bounded length
- Trailing path segments
- Malformed percent sequences
- `buildAppLinkForNote` ↔ `extractNoteIdFromUrl` round trip

A minimal `vitest.config.ts` is added for `apps/desktop` (the existing `vitest.build.config.mts` only runs build-artifact checks):

```
$ pnpm exec vitest run
 ✓ src/protocol_handler.spec.ts (23 tests) 5ms
 Test Files  1 passed (1)
      Tests  23 passed (23)
```

Per [CLAUDE.md](../blob/main/CLAUDE.md): *"When adding pure business logic (e.g., data transformations, migrations, validations), extract it as a separate function and always write unit tests for it."*

## Docs

User-facing section added under `docs/User Guide/User Guide/Installation & Setup/Desktop Installation.md` explaining the URL format, the context-menu action, and the `--new-window` flag.

## Manual smoke test

Booted `pnpm dev` on macOS — Trilium starts cleanly, the protocol registers without errors, and the `open-url` event listener attaches before `ready`.

## Disclosure

This PR was drafted with the help of Claude Code (the repo's `CLAUDE.md` was followed throughout). All design decisions, test coverage, and edge cases were reviewed before submission. Happy to iterate on review feedback.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#649: Open/focus a note from command line / desktop URL handler (Trilium URL protocol)](https://oss.issuehunt.io/repos/92111509/issues/649)
---
</details>

### DIFF
--- a/apps/client/src/components/app_context.ts
+++ b/apps/client/src/components/app_context.ts
@@ -176,6 +176,7 @@ export type CommandMappings = {
     toggleArchivedNotes: CommandData;
     sortChildNotes: ContextMenuCommandData;
     copyNotePathToClipboard: ContextMenuCommandData;
+    copyAppLinkToClipboard: ContextMenuCommandData;
     recentChangesInSubtree: ContextMenuCommandData;
     cutNotesToClipboard: ContextMenuCommandData;
     copyNotesToClipboard: ContextMenuCommandData;
--- a/apps/client/src/desktop.ts
+++ b/apps/client/src/desktop.ts
@@ -50,6 +50,11 @@ function initOnElectron() {
     const electron: typeof Electron = utils.dynamicRequire("electron");
     electron.ipcRenderer.on("globalShortcut", async (event, actionName) => appContext.triggerCommand(actionName));
     electron.ipcRenderer.on("openInSameTab", async (event, noteId) => appContext.tabManager.openInSameTab(noteId));
+    electron.ipcRenderer.on("open-note-by-id", async (_event, noteId: string) => {
+        if (typeof noteId === "string" && noteId.length > 0) {
+            await appContext.tabManager.openInSameTab(noteId);
+        }
+    });
     const electronRemote: typeof ElectronRemote = utils.dynamicRequire("@electron/remote");
     const currentWindow = electronRemote.getCurrentWindow();
     const style = window.getComputedStyle(document.body);
--- a/apps/client/src/menus/tree_context_menu.ts
+++ b/apps/client/src/menus/tree_context_menu.ts
@@ -26,7 +26,7 @@ let lastTargetNode: HTMLElement | null = null;
 
 // This will include all commands that implement ContextMenuCommandData, but it will not work if it additional options are added via the `|` operator,
 // so they need to be added manually.
-export type TreeCommandNames = FilteredCommandNames<ContextMenuCommandData> | "openBulkActionsDialog" | "searchInSubtree";
+export type TreeCommandNames = FilteredCommandNames<ContextMenuCommandData> | "openBulkActionsDialog" | "searchInSubtree" | "copyAppLinkToClipboard";
 
 export default class TreeContextMenu implements SelectMenuItemEventListener<TreeCommandNames> {
     private treeWidget: NoteTreeWidget;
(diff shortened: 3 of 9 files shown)


vayungodara:
Thanks @gemini-code-assist 🙏 — addressed in beecc75:

1. **`--new-window` race** — switched to `BrowserWindow.getAllWindows()` before/after diff per your suggestion. No longer depends on the focus-event timing in windowService.
2. **Mixed-case test** — added (`trilium://note/AbC123def456` → `AbC123def456`). Test count is now 24.
3. **Case sensitivity in `protocol_handler.ts`** — this PR uses the path-based `trilium://note/<noteId>` form, so the case-sensitive ID lives in `parsed.pathname` (preserved verbatim by the URL parser); only the literal `note` host gets normalized, which is what we want for the prefix check. I left a code comment noting this, and the new test pins down the case-preservation guarantee. If you'd still prefer a string-based parser anyway for defense-in-depth, happy to switch.

vayungodara:
(For IssueHunt bot tracking — this PR addresses bounty #649: https://oss.issuehunt.io/repos/92111509/issues/649)

/claim https://oss.issuehunt.io/repos/92111509/issues/649

vayungodara:
Rebased onto latest `main` and resolved a conflict in `apps/desktop/src/main.ts` (kept both the new `PRODUCT_NAME` import from upstream and this PR's `protocol_handler` imports). Tests still passing locally: `vitest run src/protocol_handler.spec.ts` → 24/24 passed. `pnpm typecheck` clean. Ready for review.

eliandoran:
Hi,

Sorry but AI-driven implementations for the sole purpose of claiming a bounty is not something we encourage.
Thank you.

Issue #9438, opened 2026-04-15
Add relation filter in advanced search


cpita-work:
### Describe feature

Currently:

- The "search string" field doesn't allow for notes, just text, so you have to type something like `~template.title = ...`.
- The "ancestor" field is only for ancestors and moreover for a single ancestor.

This makes searching for relations cumbersome and unintuitive. For example I would like to get all my books (a template) that have certain topic (an ancestor). Or all my tasks (an ancestor) that belong to certain project (another ancestor). These are things that Trilium excels at representing, not so much at surfacing.

Two possible solutions:

- Turn the "ancestor" field into a "relation" field with multi selection.
- Remove the ancestor field altogether and add both label/relation autocompletion and @-style autocompletion to "search string".

cpita-work:
Related to the above, Quick Search should also index relation titles (or their entire paths), so that say "book tristram" is a useful query when book is represented as a template. As a workaround one can set `#book` as a label in the template, but this is a duplication of structure that shouldn't be required.

PR #9724, opened 2026-05-14
fix :Replacing the ancestor field with multi-relation support and add…

ZAKIURREHMAN:
I set Replacing the ancestor field with multi-relation support and adding relation/template autocomplete in search would make querying much more intuitive and powerful. #9438


maphew (changes requested):
Relation filters for saved searches address #9438, but the current implementation has three integration risks.

`searchFromNote()` turns every non-`searchScript` relation returned by `getRelations()` into an AND predicate. That includes inherited relations and unrelated metadata such as `template`, while the search-definition UI shows only owned filter attributes. An existing saved search can therefore gain a constraint that the user cannot see or remove. Advise giving search filters a distinct representation and adding compatibility tests for owned, inherited, and unrelated relations.

Adding relation-target titles to `BNote.getFlatText()` also needs reverse invalidation. Renaming a target invalidates that target, but not every source note whose cached flat text now contains the old title. Quick search can continue to match the old title and miss the new one. Advise adding reverse-relation cache invalidation or leaving target titles out of flat text.

Finally, all ancestor rows share one `ancestorDepth` label, but removing any row also removes that label. Advise keeping the depth until the last ancestor is removed, or modeling depth per filter.

A rebase is also necessary because search now lives in `packages/trilium-core`. Focused server and UI tests should cover multi-filter behavior.

Issue #9749, opened 2026-05-14
Resizing Mermaid split view makes Gantt and Bar Chart vanish


hulmgulm:
### Description

I was just playing around with app.triliumnotes.org and noticed the following when playing around with the Note Types / Mermaid Diagrams pages: 
In all the pages, there is source view on the left side and the preview on the right side. I can change the width by dragging on the divider. This works fine for all diagrams but for Gantt and Bar Chart. When I drag on these, the diagrams vanish.
I created Mermaid Diagrams in my instance (0.103.0) and it behaves the same.

Another thing I noticed is, that the example mindmap (when creating a new one) contains an icon instruction, but no icon is shown in the diagram.
<img width="1004" height="521" alt="Image" src="https://github.com/user-attachments/assets/b185fcd3-282f-401a-890b-fcbdeb26c733" />
Is this correct?

### TriliumNext Version

0.103.0

### What operating system are you using?

Windows

### What is your setup?

Local + server sync

### Operating System Version

Win11 25H2

maphew:
We traced the vanish on current main (372a749ff2). It is specific to gantt-based diagrams; both demo notes that vanish ("Gantt" and "Bar chart") are gantt syntax.

Observed: dragging the split divider makes the diagram collapse to a 1-2 px sliver on the first movement.

Cause, in three parts:

1. `useResizer` in `apps/client/src/widgets/type_widgets/helpers/SvgSplitEditor.tsx` has the container width (a fresh DOMRect per ResizeObserver tick) in the deps of the svg-pan-zoom setup effect, so every width change destroys and re-creates the svg-pan-zoom instance.
2. svg-pan-zoom removes the SVG `viewBox` attribute on init (`shadow-viewport.js`, `cacheViewBox`) and `destroy()` never restores it. The re-init therefore falls back to `viewport.getBBox()` for its fit geometry.
3. Mermaid's gantt renderer always draws the "today" marker line, even when today is far outside the chart's date range. It is outside the viewBox but inside `getBBox()`. For the demo Gantt (Jan 2014 dates) the bbox is ~104,000 units wide against a 1280-unit viewBox; for the demo Bar chart (`dateFormat X`, domain 0-71 seconds) it is 33,554,428 units. `fit()` then scales the chart by 0.005 or 0.000016, so it effectively disappears. Diagrams whose bbox matches their viewBox (flowchart, xychart) survive.

[...]

PR #9960, opened 2026-05-27
fix: improve mermaid rendering 

wsjwu58-cmd:
### 1. Fix mermaid diagram vanishing on resize (#9749)
- **SvgSplitEditor.tsx**: Remove width dependency from svgPanZoom instance creation effect to prevent destroy/recreate cycle during resize. Add 100ms debounce to resize handler to avoid rapid fit() calls crushing wide diagrams (Gantt, Bar Chart).

### 2. Fix mindmap sample icon not rendering (#9749)
- **sample_diagrams.ts**: Remove ::icon(fa fa-book) from mindmap sample since FontAwesome is not loaded and mermaid's icon pack system requires @iconify-json/fa-solid registration.

### 3. Fix black screen preview + add resize handle
- **mermaid.css**: Add white background to preview container, add resize:both for draggable resize handle
- **ReadOnlyText.css**: Same for inline mermaid diagrams in read-only text view

### 4. Fix mermaid not rendering in shared contexts
- **content_renderer_text.ts**: Add missing applyInlineMermaid() call to postProcessRichContent pipeline, fixing mermaid not rendering in list views, tooltips, print views, and note includes

### Tests
- **mermaidtoolbar.ts**: Update toolbar item count assertion (5 → 7)
Previously, dragging the splitter would cause the Gantt preview to shrink excessively.
image
After the fix, resizing works as expected.
image


eliandoran:
Hi, @wsjwu58-cmd .

Generally your fixes sound good, but a single PR for all of them might not be the best approach since we might want to pick out only a few of them.

For example, regarding the re-render button for Mermaid, I don't see a valid usecase. The Mermaid diagram is supposed to re-render on change, so what case would there big to manually refresh? If the diagram fails to render, then we must fix that issue instead of having a work-around.

wsjwu58-cmd:
> 嗨， 。
> 
> 总体来说，你的修复听起来不错，但针对所有修正的单一个人记录可能不是最佳方案，因为我们可能只想挑出其中的几个。
> 
> 比如说，关于美人鱼的重渲染按钮，我看不出有合理的使用场景。美人鱼图应该在更改时重新渲染，那什么情况下需要手动刷新呢？如果图示无法渲染，我们必须修复这个问题，而不是靠变通方法。

Thanks for your reply! Since some Mermaid diagrams fail to render properly and appear as a black screen when I import notes, I've added a re-render button to quickly refresh and display the diagram correctly.
image
This is indeed a temporary workaround, so we still need to think further about the optimal solution to the underlying issue.

wsjwu58-cmd (commented):
review

PR #9962, opened 2026-05-27
feat(share): show matched content snippet in share search results

perfectra1n:
## Summary

When you "share" notes from Trilium as a static site (or via the dynamic share viewer) and use the search box, the result list shows only the note title and its path. If the search term doesn't appear in the title, there's no way to tell *why* a result matched — which made the search feel broken even when it was finding the right notes.

Example: searching for `Magnesium` on a shared subtree returns five hospital/SRI reports, none of whose titles contain "Magnesium". The user has no idea what each note matched on:

image

This PR adds a content snippet under each result, with the matched terms bolded.

### Server-side (dynamic share)
`apps/server/src/share/routes.ts` — `/share/api/notes` now calls the existing `searchService.extractContentSnippet` + `searchService.highlightSearchResults` helpers (already used by the in-app autocomplete) for the top 20 results, and returns:

- `snippet` — plain-text snippet
- `highlightedSnippet` — HTML where matched tokens are wrapped in `<b>...</b>`

Snippet extraction is capped at 20 results (the popout only renders 5) so the existing payload size for large result sets stays the same. Results beyond the cap are still returned, just without a snippet.

### Static (Fuse) share-export
`packages/share-theme/src/scripts/modules/search.ts` — the Fuse instance now opts in to `includeMatches`. The client builds a windowed snippet centred on the first `content` match, wraps every match-range inside the window in `<b>...</b>`, and falls back to a head-of-content preview when only the title matched.

### Share-theme client
- `buildResultItem` renders a new `.search-result-snippet` line under the path.
- All user-supplied strings (`title`, `path`, plain `snippet`) are now HTML-escaped before injection. The server-rendered highlighted snippet is already sanitized by the search service (only `<b>` / `<br>` tags are inserted, and the service strips `<` / `>` / `{` / `}` from the source content first) and is inserted as-is.
- New `.search-result-snippet` styling in `packages/share-theme/src/styles/popouts/search.css`: 12 px, two-line clamp, bolded match runs that recolor on hover so highlights stay legible against the active-row background.

## Notes

- No new dependencies; only reuses helpers already exported from `searchService`.
- API change is additive (two new optional fields on the response). Older shared-export bundles that don't have the new client will simply ignore the fields.
- Tested against both code paths:
  - Dynamic: `GET /share/api/notes?ancestorNoteId=...&search=magnesium` now includes `snippet`/`highlightedSnippet`.
  - Static: rebuilt a sample export and confirmed `search-index.json` is unchanged and the popout shows highlighted snippets from the Fuse `matches` payload.


maphew (changes requested):
Content snippets make shared search results easier to understand. This branch predates the authorization checks now on `main`.

The current head maps every search result to a public title, path, and content snippet. Current `main` filters each result with `hasCredentialAccess()` and `isVisibleInShareTree()` to exclude credential-protected notes, hidden notes, and clone paths outside the shared subtree. Rebasing this head without moving snippet extraction after that filter would expose restricted note metadata and content. Advise filtering first, then reading and highlighting content only for authorized results.

Dynamic endpoint tests should prove that snippets are absent for credential-protected notes, `shareHiddenFromTree` notes, hidden branches, and disallowed clone paths. Positive tests should cover plain and highlighted snippets after authorization.

The two committed `apps/website/.svelte-kit` files are generated output, and the lockfile integrity deletion is unrelated to this feature. Advise removing that churn and regenerating only the required lockfile change after rebasing.

apps/website/.svelte-kit/types/route_meta_data.json
eliandoran, line 1: Redundant file, remove.

apps/website/.svelte-kit/tsconfig.json
eliandoran, line 1: Redundant file, remove.

eliandoran (changes requested):

eliandoran (approved):

PR #9993, opened 2026-05-30
Feat #5825: Image Annotation (Draw & Write) in Trilium Notes

### LINKED ISSUE #5825: (Feature request) annotate image
### Describe feature

Draw, write, scribble, annotate on image

### Additional Information

_No response_

Dinis-Ou:
Adds a support for annotating images directly within Trilium Notes. Users can open an annotation overlay on a embedded image in a note using Excalidraw, allowing them to write or draw on top of the image. The annotation is saved and displayed
persistently on the image.

Closes #5825


maphew (changes requested):
The annotation workflow is promising, but the current persistence model has a data-loss path and destructive format behavior.

`loadScene()` converts attachment-list failures, blob failures, invalid JSON, and unsupported scene versions into the same `null` used for “no annotation.” The editor then starts a new scene, and the next save replaces the existing scene attachment by title. A transient read failure can therefore overwrite valid editable annotation data. Advise distinguishing not-found from load failure and blocking saves when an existing scene cannot be loaded.

Every save also replaces the source with a PNG. Annotating SVG, GIF, WebP, or another supported format loses vector data or animation and changes the stored MIME type without warning. The editable JSON also stores a base64 copy of the complete original image, which duplicates large blobs in the database and sync stream.

The product behavior needs an explicit decision on whether annotation is destructive. Advise using a separate rendered derivative and a stable reference to the original blob. Current `main` has canvas persistence and image-attachment services that can provide the shared storage boundary.

Round-trip, load-failure, format-conversion, large-image, sync, and export tests should accompany the implementation.

Issue #6829, opened 2025-08-29
Allow sorting notes by multiple attributes


BeatLink:
### Describe feature

It would be great if trilium could support sorting notes by multiple attributes. 

From the UI side, this could be achieved by setting sorted to multiple comma separated values (e.g. #sorted=priority,dueDate,title)

For implementation, the code could sort each note by the leftmost attribute, and if those attributes are equal then it uses the further right attributes for sorting. 

So in the example above, first the code sorts by priority. If the priority is the same then it sorts by dueDate. If the due dates are the same, then it sorts by title. 

This may also need an expansion to the #sortDirection attribute to be comma separated as well, (e.g. #sortDirection="asc,desc,desc") where each sort direction would correspond to the sorted attribute.

rom1dep:
@BeatLink could you provide a mock-up (ideally as exported notes in a .zip) to illustrate how you would expect it to work?  
Especially in combination with overrides like `#top`, `#bottom`?

I can't promise doing anything about it in the short, but worst case it would serve as a spec for who may end-up implementing this, with a use-case that justifies the effort.

BeatLink:
Only the third stage of sorting would be affected. I.e. #top, #bottom and last resort alphabetical would not be affected. When sorted is equal to a value with a comma in it, it would sort it in sequential order eg #sorted="dueDate,priority" would sort the notes first by #dueDate. Where the #dueDate values are the same for 2 notes, the notes are then sorted by #priority and so on. 

To simplify the sort direction logic, I came up with this idea instead. To enable switching the direction of the sort during the sequence, ascending and descending options can be provided. Eg #sorted="dueDate:asc,priority:desc". This would sort the notes first by #dueDate in ascending order, then by #priority in descending order.

What would the mockup look like exactly?

rom1dep:
> What would the mockup look like exactly?

A set of notes covering all the edge-cases one might think of:
- mixing configurations (having `#sorted=`, `#sortNatural`, `#top`, …)
- expected behaviour when sorting against null (e.g. when a note is missing a dueDate, should it come on top or last?)
- expected behaviour when there are syntactic issues with the sorting definition (e.g. `#sorted="dueDate:acs,priority:desc"`, `#sorted="dueDate:asc;priority:desc"`)

and so on :-)

CoilBlood:
I'm also extremely interested in this feature. Can you implement it in Trilium?

CinderCanard:
[...]

PR #10060, opened 2026-06-05
Support multi-attribute note sorting

Rajesh270712:
## Summary

Adds multi-attribute sorting support for child notes when `#sorted` contains a comma-separated sequence such as `dueDate:asc,priority:desc`.

## Linked Issue

Related to TriliumNext/Trilium#6829.

## Problem

The current note tree sorting path only uses one sorted attribute, so ties on the first attribute fall through to the existing fallback ordering instead of applying additional user-provided sort keys.

## Fix

- Parse multiple `#sorted` entries with optional `:asc` / `:desc` direction.
- Compare child notes by each configured attribute in order before falling back to the existing behavior.
- Add focused tree service specs for multi-key ordering, descending secondary sort, and malformed sort direction fallback.

## Validation

- `git diff --check origin/main...HEAD`
- `npx --yes pnpm@11.5.0 --filter server test ../../packages/trilium-core/src/services/tree.spec.ts`
- `npx --yes pnpm@11.5.0 typecheck`
- `npx --yes pnpm@11.5.0 --filter server build`

The server build completed with existing bundler warnings outside this patch surface.

## Risk Notes

- Scope is limited to `packages/trilium-core/src/services/tree.ts` and `packages/trilium-core/src/services/tree.spec.ts`.
- `#top`, `#bottom`, natural sorting, and existing fallback ordering are preserved.
- I did not change user-facing docs because this repo appears to route documentation edits through the separate `edit-docs` workflow; I can update that path if maintainers prefer.

### DIFF
--- a/packages/trilium-core/src/services/tree.spec.ts
+++ b/packages/trilium-core/src/services/tree.spec.ts
@@ -94,6 +94,42 @@ describe("Tree", () => {
         }
     });
 
+    it("sorts notes by multiple attributes in order", () => {
+        const note = buildNote({
+            children: [
+                {title: "first", "#priority": "1", "#dueDate": "2026-05-20"},
+                {title: "third", "#priority": "2", "#dueDate": "2026-05-01"},
+                {title: "second", "#priority": "1", "#dueDate": "2026-05-10"}
+            ],
+            "#sorted": "priority,dueDate"
+        });
+
+        getContext().init(() => {
+            tree.sortNotesIfNeeded(note.noteId);
+        });
+
+        const orderedTitles = note.children.map((child) => child.title);
+        expect(orderedTitles).toStrictEqual(["second", "first", "third"]);
+    });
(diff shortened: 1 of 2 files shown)


maphew (changes requested):
`sortNotes()` replaces a missing criterion value with the note title before deciding whether to continue to the next criterion. For example, with `#sorted="priority,dueDate"`, two notes that both lack `priority` but have different titles are ordered by title, so `dueDate` is never compared. That contradicts the proposed left-to-right multi-key semantics and does not define the null behavior requested in #6829.

Advise handling missing values explicitly for each criterion: continue when both values are missing, define whether a single missing value sorts first or last, and use the title fallback only after all criteria compare equal. Tests should cover both-missing and one-missing cases before and after a populated secondary key.

eliandoran:
Closing in favor of #11424.

PR #10121, opened 2026-06-11
feat(collections/dashboard): first implementation

eliandoran:


hulmgulm:
This sounds interesting. I'm looking forward to the next nightly to start playing around with it.

hulmgulm:
Looks good so far. I noticed, that when dropping an archived note into the dashboard, it won't show until  one enables show archived notes in the dashboard settings. But first I thought, something is broken. I understand when this setting is not enabled by default for other collection types but I deliberately dragged an archived note into it and nothing happend. Maybe in this case, show a notification like "Archived notes are hidden in this dashboard. Change the dashboard settings to have them displayed".

hulmgulm:
I just noticed that when I click several times "dashboard" - "child note of dashboard" -  "dashboard" - "child note of dashboard" - ...   the dashboard looses its layout.

eliandoran:
@hulmgulm , thanks for testing. I'll have a look. Let me know if you find anything else.

hulmgulm:
I found something else @eliandoran: 
I added a render note (let's call it graph) to my dashboard. The child script note of the note referenced by renderNote tries to get some labels from graph which configures the output. It uses api.getActiveContextNote() to get the graph note. However, if the graph note is displayed on the dashboard, api.getActiveContextNote() returns the dashboard note and not the graph note. Now the script cannot get the labels from graph note. api.currentNote always points to the script note.
How can the script reliably get the graph note to get the labels?

eliandoran:
@hulmgulm ,

Try logging/displaying `api.startNote`, `api.currentNote`, `api.originEntity` and let me know if neither one works correctly.

hulmgulm:
Thank you. api.originEntity did the trick.

PR #10135, opened 2026-06-12
perf(desktop): faster Electron startup

eliandoran:
## Summary

Cuts desktop cold-start time by restructuring the boot sequence, with instrumentation to keep it measurable. On the reference machine, time to first paint dropped from ~1245ms to ~1015ms (cold) / ~640ms (warm), and full client render from ~1617ms to ~1327ms (cold) / ~950ms (warm).

- **Window creation no longer waits for the server.** The `trilium-app://` protocol handler is installed up-front and accepts a *promise* of the Express app — renderer requests arriving early simply wait inside the handler. `onReady()` is gated only on `initializeCore()` (DB open, options/translations readable) instead of full server startup, so the renderer process spins up concurrently with Express being built. Regression tests pin both behaviors (window created while the server hangs; early request held until the app resolves).
- **V8 code cache enabled for `trilium-app://`** (`codeCache: true`). Chromium only code-caches http(s) schemes by default, so the renderer bundle was recompiled from source on every launch; now compiled bytecode is reused from the third launch onward (Chromium's see-it-twice heuristic).
- **Startup timing instrumentation.** Boot phases, window milestones (`ready-to-show`, `did-finish-load`) and the renderer's full-render point are measured from OS process creation (`process.getCreationTime()`) and written as per-phase deltas to `startup-metrics.log` (cwd, gitignored). The renderer reports through a new allowlisted `electronApi.window.reportStartupMetric()` bridge.
- **Docs**: note the `ELECTRON_RUN_AS_NODE` env leak (VS Code extension host shells) that crashes Electron launches with "Not running in an Electron environment!".

Evaluated and rejected with measurements: Node `module.enableCompileCache()` for the main bundle (zero effect — V8 lazy compilation makes the upfront parse cheap; pre-main time is dominated by Chromium bootstrap and cold I/O) and main-process V8 heap snapshots (tooling abandoned, bundle largely ineligible).

## Test plan

- `pnpm --filter desktop test` (new specs: `startup_metrics.spec.ts`; extended: `main.spec.ts`, `window.spec.ts`, `protocol.spec.ts`, `preload.spec.ts`)
- `pnpm typecheck`
- Manual: production build (`start-prod`) boots, window appears before server finishes starting, metrics file written; verified across multiple cold/warm runs


eliandoran:
Addressed review feedback in 936233ae64:

- **`expressAppPromise` left pending on server-start failure** (gemini-code-assist, greptile): fixed — the server start is now wrapped in try/catch and the promise is rejected, so the protocol handler answers pending/future renderer requests with a 500 instead of hanging the visible window. Covered by new regression tests in `main.spec.ts` and `protocol.spec.ts`.
- **`writeFileSync` → async `fs.writeFile` in `startup_metrics.ts`** (gemini-code-assist): declined. The file is ~500 bytes written at most 9 times during startup (sub-ms each), so there's no measurable blocking; and concurrent async rewrites would introduce a last-writer-wins ordering race (an older, shorter snapshot can finish after a newer one), which the synchronous write avoids by construction.

PR #10137, opened 2026-06-12
feat(desktop): add minimize-to-tray and close-to-tray options

### LINKED ISSUE #3743: (Feature request) Minimize/close to tray icon
### Describe feature

(Note: on Windows, but since this is similar between OSes I assume on all platforms)
While Trilium can display a tray icon, it cannot be used "persistently". Most "tray icon" apps (Slack, Signal, Ferdi, even Obsidian with plugin) have a close/minimize to tray option, where minimise/close action simply hides the window, the app can only be exited by tray context menu. This would make the tray icon 100x more useful, since currently the app will take up taskbar space regardless of the tray icon. There are utilities that can force "correct" tray behaviour like [Trayconizer](https://download.cnet.com/Trayconizer/3000-2346_4-10971946.html) but adding these options directly would be probably not much effort and would be so much cleaner. As a note, instead of "Tray" "Tray icon" would be a better/more apt name for these functionalities.

TLDR: separate options to "minimize to tray" and "close to tray", which would make the respective "_" anx "X" buttons the Window disappear, only to be revealed by clicking tray icon (or hotkey).

### Additional Information

_No response_

eliandoran:
Closes #3743

Adds two opt-in desktop options, shown in **Options → Other → System Tray** (and disabled there while the tray icon itself is turned off):

- **Minimize to tray** — minimizing a window hides it to the tray icon instead of the taskbar.
- **Close to tray** — closing a window hides it to the tray icon instead of exiting; Trilium is quit from the tray menu.

## How it works

- Both behaviors are implemented in the Electron main process (`window.ts`) as `minimize`/`close` listeners installed on the main and extra windows. They read the options at event time, so toggling them takes effect immediately, no restart needed.
- Both are guarded on the tray icon being enabled, so a window can never become unreachable.
- An explicit quit (tray menu, Cmd+Q) sets an `isQuitting` flag via `before-quit` that bypasses the close interception. The tray's *Quit Trilium* item now calls `app.quit()` instead of closing each window individually, both so it emits `before-quit` and because that is what the label promises.
- Electron's `minimize` event is not preventable, so the window is hidden after the fact; the tray restore paths now go through `showAndFocusWindow()`, which un-minimizes before showing.
- The options are per-device (`isSynced: false`), like `disableTray`.

Also removes a duplicated `disableTray` entry in `ALLOWED_OPTIONS`.

## Testing

- New unit tests cover the interception (option on/off, tray disabled, main + extra windows) and the quit bypass; the tray quit-item test now asserts `app.quit()`.
- `apps/desktop`, `@triliumnext/commons`, client and server options suites pass; `pnpm typecheck` passes. (3 pre-existing Windows-only desktop test failures reproduce identically on `main`.)


eliandoran:
Closing in favor of #10174.

PR #10196, opened 2026-06-17
feat(text): add an option to auto-capitalize first letter of sentence

MShahnoor:
I recently moved my notes from Microsoft OneNote to Trilium, and the one habit my fingers can't shake is automatic capitalization. OneNote has option to capitalize the first letter of a sentence as you type, and this is a feature that I'm used to so I would love to see it in Trillium too.

This adds that as an optional setting for text notes. When it's on, Trilium capitalizes the first letter of a sentence as you type: at the start of a line, and after a period, exclamation mark, or question mark. It's off by default, so nothing changes for anyone unless they turn it on, under Options > Text Notes > Features, next to the emoji, note, and slash-command toggles.

Under the hood it's a small CKEditor plugin that hooks the insertText command and capitalizes the letter right before it's inserted, so it stays a normal edit (undo works, and existing formatting is kept). The new option and the way it's switched off (removing the plugin in getDisabledPlugins) follow the same pattern as the existing emoji and slash-command toggles.

I added tests for the plugin covering the sentence-start cases, mid-sentence typing, code blocks, inline code, and pasting, and pnpm typecheck is clean.

Here's what it looks like
 

https://github.com/user-attachments/assets/e7fb49c3-2e2e-44ed-a1e7-edb12189338d

I kept the change small and tried to match how the existing editor toggles are built. Happy to adjust anything.

### DIFF
--- a/apps/client/src/translations/en/translation.json
+++ b/apps/client/src/translations/en/translation.json
@@ -2316,7 +2316,9 @@
     "note_completion_enabled": "Note auto-completion",
     "note_completion_description": "Links to notes can be created by typing `@` followed by the title of a note.",
     "slash_commands_enabled": "Slash commands",
-    "slash_commands_description": "Editing commands such as inserting line breaks or headings can be toggled by typing `/`."
+    "slash_commands_description": "Editing commands such as inserting line breaks or headings can be toggled by typing `/`.",
+    "auto_capitalize_enabled": "Automatic capitalization",
+    "auto_capitalize_description": "Capitalizes the first letter of a sentence as you type, after `.`, `!` or `?` and at the start of a line."
   },
   "table_view": {
     "new-row": "New row",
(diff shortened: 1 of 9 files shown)


maphew (changes requested):
I found two capitalization edge cases:

1. `capitalizeWordEndingAt()` treats the full trailing `\S+` run as a word, then examines only `word[0]`. If a sentence starts with opening punctuation, that character has no case and the method returns. Inputs such as `"hello ` at block start and `Done. "hello ` therefore stay lowercase. Advise locating the first cased letter after supported opening punctuation and adding tests for straight and curly quotes and parentheses.

2. The method removes one UTF-16 code unit but inserts the full result of `toUpperCase()`. A length-changing mapping such as `ßeta` to `SSeta` violates the stated length-neutral replacement assumption. Advise defining the language/locale rule, operating on code points rather than code units, and at minimum rejecting length-changing mappings. Add tests for these Unicode cases.

Issue #8635, opened 2026-02-06
Allow internal links to open PDF notes at a specific page


jdsp4k:
### Describe feature

Trilium uses a customized PDF.js viewer, which supports URL fragment parameters like #page=, #zoom=, #search=, etc. However, unless I'm missing something these parameters cannot currently be used with internal note links, which means it is impossible to deep-link to a specific page in a PDF note from another note. I think this ought to be a simple fix as PDF.js already supports this out-of-the-box.

This would be very useful for, for example: creating "table of contents" notes or making references to specific pages of large PDFs.

Thanks for the great app!

rishacha:
Can we also add this - linking highlight/annotations/image snippets in existing notes so that clicking on them can open the pdf at the exact highlight/annotation location?

PR #10224, opened 2026-06-19
feat: Adds area annotations and linkable annotations to PDFs

rishacha:
## Overview
This PR adds two complementary annotation features to Trilium's built-in PDF viewer, giving users the ability to create persistent, navigable cross-references between PDF content and their notes.

1. **Linkable text annotations** — copy a shareable link from any existing highlight or sticky-note annotation; pasting it anywhere in Trilium creates a chip that, when clicked, opens the PDF and scrolls directly to that annotation.
1. **Image (area) annotations** — drag a rectangle over any region of a PDF page to capture it as an image; the image is stored as a note attachment, shown in the right panel, and can be pasted into any note as a resizable, clickable image that navigates back to the PDF location.

---
## How it differs from the existing annotation implementation
Trilium's existing PDF viewer already lets users create and edit annotations inside the PDF itself (using PDF.js's built-in highlight and FreeText editors) and shows them in the right panel sidebar. What it lacked was any connection between those annotations and the rest of the note-taking system — there was no way to reference a specific annotation from another note, or to capture a visual region of a PDF.

This PR adds the "outbound" layer: every annotation (text or image) now has a stable address that can be embedded in any note, navigated to from a link, and used in the same way as any other Trilium internal cross-reference.

## Feature 1 — Linkable text/highlight annotations

https://github.com/user-attachments/assets/0e738e91-9c30-45c0-8e8f-97c252db5e92

### What it achieves
Any highlight or comment in a PDF can be shared as a Trilium reference link. Clicking the link opens the PDF note and scrolls to the exact annotation. The link is human-readable in the sidebar with the annotation's preview text.

### How it was implemented

URL format — ViewScope in `link.ts` was extended with three new params:

`#root/noteId?annotationId=12R&annotationPage=5&annotationPreview=The+quick+brown+fox`
* **annotationId** — the PDF.js annotation object reference (e.g. 12R). Used for exact matching.
* **annotationPage** — page number, used as a fallback when the ID is stale.
* **annotationPreview** — up to 60 characters of the highlighted text or comment, used for fallback matching and as the visible link label.

**Copy link** — each annotation item in the right panel PdfAnnotations sidebar gains a hover button and a right-click context menu entry. It copies a <a class="reference-link" href="…"> HTML element to the clipboard using document.execCommand("copy") (compatible with both HTTP and HTTPS). Pasting into a rich-text note produces a reference link chip displaying "Note title › 'preview text'".

**Navigation and scroll** — when the link is clicked, goToLinkExt parses the annotation params into viewScope and navigates. The scroll is performed via a postMessage (trilium-scroll-to-annotation) to the PDF.js iframe, which calls scrollToAnnotation. A pendingAnnotationIdRef ref survives spurious viewScope resets and a viewerAreaReadyRef ensures the scroll message is only sent after the viewer's listener is registered.

**ID rotation fallback** — PDF.js assigns temporary editor IDs (e.g. pdfjs_internal_editor_0) to newly created annotations. After the PDF is saved, extractFromSavedData re-reads the file and permanent object-reference IDs (e.g. 12R) replace them. A copied link might therefore contain a stale ID. resolveAnnotation (now in its own module pdfAnnotationResolver.ts, with 10 unit tests) handles this: it first tries exact-ID matching, and if that fails, falls back to matching on page number + the annotationPreview content.

**Right-click context menu** — text annotations support right-click → Copy link / Change colour / Delete. The colour change and delete are routed via trilium-set-annotation-color and trilium-delete-annotation postMessages to the PDF.js annotation storage API. The context menu is attached via jQuery document delegation (not Preact's onContextMenu) to ensure it fires reliably inside Trilium's legacy widget wrapper.

## Feature 2 — Image (area) annotations

https://github.com/user-attachments/assets/ac8da28f-48cd-48a0-9c4e-c4de368fc628

### What it achieves
Users can select a rectangular region of any page in the PDF viewer, capture it as a PNG, and have it appear in the right panel as a thumbnail. Area annotations support notes/comments, colour-coded overlays visible directly on the PDF, a right-click context menu, and a copy button that produces a resizable, linked image when pasted into a note.

### How it was implemented
**Capture tool** — area.ts (new file in packages/pdfjs-viewer) adds a crosshair button to the PDF.js toolbar. On drag-release it reads the page canvas pixels for the selection rectangle, encodes them as a base64 PNG, and sends a pdfjs-viewer-area-capture postMessage to the parent.

**Storage** — images are uploaded via POST /api/notes/:noteId/attachments/upload (the binary-safe multipart route) so they are stored as proper binary PNG attachments with role: "image". Metadata (attachment ID, page, rect, comment, colour) is persisted as a areaAnnotation note label JSON on the PDF note. This means:

* No child notes are created in the note tree.
* Annotations survive across sessions and sync devices.
* Deletion removes both the label and the attachment.

**Right panel** — PdfAreaAnnotations.tsx reads the areaAnnotation labels, shows PNG thumbnails, and on hover displays an enlarged preview centred in the viewport. A right-click context menu offers Copy link / Add note / Change colour / Delete.

**Persistent overlays on the PDF** — area.ts draws semi-transparent coloured rectangles on the page divs for every stored area annotation. A pagerendered event listener redraws overlays after PDF.js's virtual scroller evicts and recreates page divs. Right-clicking an overlay inside the PDF viewer itself sends a pdfjs-viewer-area-right-click postMessage; the parent shows the same context menu.

**Copy as resizable image** — the copy button produces `<figure class="image"><a href="annotationHash"></a></figure>`, which pastes into CKEditor as a resizable linked image (Trilium's ImageResize plugin is registered). Clicking the image in the note navigates to the PDF and scrolls to the area. No new image note is created.

**Scroll to area** — the trilium-scroll-to-area message triggers scrollToArea in area.ts. If the target page is already rendered, it scrolls precisely to the vertical midpoint of the captured rect. If the page is virtualized (not in DOM), it scrolls to an estimated position and sets pendingScrollTarget; the pagerendered listener then performs the precise scroll once the page is actually rendered.

## Shared infrastructure

* `pdfAnnotationColors.ts` — a single source of truth for the 5-colour palette (Blue / Yellow / Green / Red / Purple) shared by text annotations, area annotations, and the PDF overlay renderer. The context menu renders coloured circle icons using CSS rules with higher specificity (0,3,0) than Trilium's theme-level override (0,2,0).
* `textPrompt.ts` — a non-blocking Bootstrap modal that replaces window.prompt() for "Add note" dialogs, matching Trilium's design system.
* `pdfAnnotationResolver.ts` — extracted pure function with 10 unit tests covering exact-match, fallback-match, ellipsis stripping, and edge cases.
* `link.spec.ts` — 7 new tests covering calculateHash serialisation and parseNavigationStateFromUrl round-trip for all three new annotation URL params.

Implements #8635

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ Thumbs.db
 vite.config.*.timestamp*
 vitest.config.*.timestamp*
 test-output
+test-results
 __screenshots__
 .vitest-attachments
 
--- a/apps/client/src/components/note_context.ts
+++ b/apps/client/src/components/note_context.ts
@@ -58,6 +58,14 @@ export interface NoteContextDataMap {
     pdfAnnotations: {
         annotations: PdfAnnotationInfo[];
         scrollToAnnotation(annotationId: string, pageNumber: number): void;
+        setAnnotationColor(annotationId: string, color: string): void;
+        deleteAnnotation(annotationId: string, pageNumber: number): void;
+    };
+    pdfAreaAnnotations: {
+        annotations: PdfAreaAnnotationInfo[];
+        scrollToArea(pageNumber: number, rect: { x: number; y: number; width: number; height: number }): void;
+        deleteArea(attachmentId: string, attributeId: string): void;
+        updateArea(attributeId: string, patch: { comment?: string; color?: string }): void;
     };
     saveState: {
         state: SaveState;
--- a/apps/client/src/services/clipboard_ext.ts
+++ b/apps/client/src/services/clipboard_ext.ts
@@ -35,3 +35,41 @@ export function copyTextWithToast(text: string) {
         toast.showError(t("clipboard.copy_failed"));
     }
 }
+
+/**
+ * Copies both an HTML and a plain-text representation to the clipboard, so pasting into
+ * a rich-text editor (e.g. CKEditor) preserves formatting (links, images) while pasting
+ * into a plain-text field falls back to readable text.
+ *
+ * Uses the `copy` event's `clipboardData.setData()` instead of the classic "select a hidden
+ * DOM node, then execCommand('copy')" trick. The latter relies on each browser's own
+ * selection-to-clipboard HTML serialization, which is inconsistent — in particular Firefox
+ * does not reliably populate the `text/html` clipboard flavor that way, so pasted content
+ * silently degrades to plain text. Writing the flavors directly during the `copy` event is
+ * the documented cross-browser-reliable approach and does not require any selection at all.
+ */
+export function copyRichText(html: string, plainText: string): boolean {
+    function listener(e: ClipboardEvent) {
+        e.clipboardData?.setData("text/plain", plainText);
+        e.clipboardData?.setData("text/html", html);
+        e.preventDefault();
+    }
+
+    try {
+        document.addEventListener("copy", listener);
+        return document.execCommand("copy");
+    } catch (e) {
+        console.warn(e);
+        return false;
+    } finally {
+        document.removeEventListener("copy", listener);
+    }
+}
+
+export function copyRichTextWithToast(html: string, plainText: string) {
+    if (copyRichText(html, plainText)) {
+        toast.showMessage(t("clipboard.copy_success"));
+    } else {
+        toast.showError(t("clipboard.copy_failed"));
+    }
+}
--- a/apps/client/src/services/link.spec.ts
+++ b/apps/client/src/services/link.spec.ts
@@ -173,6 +173,74 @@ describe("calculateHash", () => {
     it("produces only the param string when note path is empty", () => {
         expect(calculateHash({ ntxId: "n1" } as any)).toBe("#?ntxId=n1");
     });
+
+    it("includes annotationId and annotationPage when set", () => {
+        const hash = calculateHash({
+            notePath: "root/abc",
+            viewScope: { annotationId: "12R", annotationPage: 5 }
+        } as any);
+        expect(hash).toBe("#root/abc?annotationId=12R&annotationPage=5");
+    });
+
+    it("includes annotationPreview when set", () => {
+        const hash = calculateHash({
+            notePath: "root/abc",
+            viewScope: { annotationId: "12R", annotationPage: 3, annotationPreview: "hello world" }
+        } as any);
+        expect(hash).toContain("annotationId=12R");
+        expect(hash).toContain("annotationPage=3");
+        expect(hash).toContain("annotationPreview=hello%20world");
+    });
+
+    it("omits annotationPage when falsy", () => {
+        const hash = calculateHash({
+            notePath: "root/abc",
+            viewScope: { annotationId: "12R" }
+        } as any);
+        expect(hash).not.toContain("annotationPage");
+        expect(hash).not.toContain("annotationPreview");
+    });
+});
+
+describe("parseNavigationStateFromUrl — annotation params", () => {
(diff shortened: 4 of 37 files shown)


eliandoran:
Hi,

Thanks for the PR, but without a discussion or presentation on how this annotation system works I will not be able to merge it.

Also, how does this work in comparison to the existing annotations we have at PDF level?

rishacha:
@eliandoran  - This introduces the area highlight - if I want to highlight an area in the PDF  - like an image or a chart. As for the existing features in pdf annotations, it adds links to the existing annotations as well as the new area highlights. I can click on a button, get a pasteable link that can be added to notes. Once you click on the link, it'll take you to the exact page in the PDF where the annotation is present. This is a rough PR. I want this feature badly so I vibe-coded it and I expect it to have issues and require some review cycles.


How can I help accelerate the merge? Do I present in a video call?

rishacha:
@eliandoran - How can I accelerate the merge for this? What do you need from me?

eliandoran:
Hi, @rishacha .

I would have to look at the implementation and see if it matches the philosophy of Trilium.

To accelerate it, please provide some information: how it works, what needs it solves.

Please note that opening a pull request does not guarantee it will get merged.

rishacha:
Hi @eliandoran

I've updated the PR description with implementation details and attached demo videos showing the workflow end-to-end.

The PR addresses two related use cases:

* Area highlights for non-text PDF content such as diagrams, charts, and images.
* Deep links that allow navigation directly to a PDF annotation from elsewhere in Trilium.

I've tried to document both the motivation and the implementation so the feature can be evaluated on its merits and on its fit with Trilium's design philosophy.

At this point, I'd appreciate feedback on any specific concerns regarding the approach, architecture, UX, or maintainability. If there are aspects that don't align with the project's direction, I'm happy to iterate on them.

rishacha:
@eliandoran  - Did you get a chance to examine this PR?

contributor:
@rishacha Did you consider displaying both annotation types in single list? It is likely better UX.

rishacha:
Yes @contributor - that actually makes sense. I'll add it in a single list on the right pane. Thank you for taking the time and providing this feedback.

contributor:
> Yes @contributor - that actually makes sense. I'll add it in a single list on the right pane. Thank you for taking the time and providing this feedback.

You better wait for @eliandoran opinion.

My opinion is:
- adding links to existing annotations is great feature on top of existing annotations. I would merge it if it were separate PR
- area annotations is nice too, but it has too many open ended possibilities: how to store? how to show? how to integrate the button with pdf.js toolbar - this is a brittle part that may silently break with pdf.js releases.


And in general (not related to this PR), if the both annotations lend in single UI (or even just annotation links)  - I think right side bar is too restrictive and too small panel for annotations feature. The real-life annotations list can be large so it may better fit into (readonly) note split or something more dedicated (and not lumped with toc/pages)

adoriandoran:
@rishacha, I have checked your PR. The feature actually seems promising. But for now, there are some issues that need to be addressed:

- Pasting links doesn't seem to work in Firefox. Plain text is pasted instead of a link.
- If you close the current PDF tab, open it again, and then create new area, it will actually create two areas at the exact same location.
- Opening the context menu on an area and selecting another color doesn't always work (after refreshing, the new color is used, so it might be a race condition or something like that).

I'll do thorough testing after you solve these issues.

rishacha:
@eliandoran @adoriandoran - FYI - I'm still working on this - Struggling a bit with `playwright` tests. Will try to post an update ASAP

PR #10308, opened 2026-06-27
feat(attr): add #openAtBottom label to open notes scrolled to the bottom

kleutzinger:
## Summary

Adds a new built-in label `#openAtBottom` that scrolls the viewport to the bottom of a note's content when the note is opened. Useful for append-only notes such as logs or journals.

## Behavior

- On opening a note with `#openAtBottom`, the viewport scrolls to the end and the cursor is placed there
- Works even when the editor has no focus (e.g. navigating via the note tree)
- Works for both text notes (CKEditor5) and code notes (CodeMirror)

## Implementation

The scroll is triggered from two places to cover all open paths:

- **`onContentChange`** — fires on every note switch; scrolls the `.scrolling-container` and moves the CKEditor model selection to `"end"` (or calls `scrollToEnd()` for CodeMirror)
- **`onEditorInitialized`** — covers the initial mount case where content may arrive before the editor is ready

### Files changed

- `packages/commons/src/lib/attribute_names.ts` — type-safe `openAtBottom: boolean` label definition
- `packages/commons/src/lib/builtin_attributes.ts` — registers the label for attribute autocomplete
- `apps/client/src/widgets/type_widgets/text/EditableText.tsx` — `scrollEditorToBottom()` helper sets the CKEditor model selection to `"end"` and scrolls `.scrolling-container`; called from both `onContentChange` and `onEditorInitialized`
- `apps/client/src/widgets/type_widgets/code/Code.tsx` — in `onContentChange`, calls `scrollToEnd()` and scrolls `.scrolling-container` (same pattern as `insertBlankLineAtTop`)
- `packages/codemirror/src/index.ts` — fixes `scrollToEnd()` to include `scrollIntoView: true`; previously it only moved the cursor without scrolling CodeMirror's own view unless `focus()` was called afterward
- `docs/User Guide/.../Attributes/Labels.md` — documents the new label in the predefined labels table

## Test plan

- [ ] Add `#openAtBottom` to a text note with enough content to scroll; navigate away and back — viewport and cursor should be at the bottom
- [ ] Same for a code note
- [ ] Verify that clicking back into an already-open note does not re-scroll to the bottom
- [ ] Verify a note without the label still opens at the top
- [ ] Verify `#openAtBottom` appears in attribute autocomplete

### DIFF
--- a/apps/client/src/widgets/type_widgets/code/Code.tsx
+++ b/apps/client/src/widgets/type_widgets/code/Code.tsx
@@ -112,6 +112,15 @@ export function EditableCode({ note, ntxId, noteContext, debounceUpdate, parentC
             codeEditor.setText(content ?? "");
             codeEditor.setMimeType(note.mime);
             codeEditor.clearHistory();
+            if (note?.hasLabel("openAtBottom")) {
+                requestAnimationFrame(() => {
+                    editorRef.current?.scrollToEnd();
+                    const scrollContainer = containerRef.current?.closest(".scrolling-container");
+                    if (scrollContainer instanceof HTMLElement) {
+                        scrollContainer.scrollTo({ top: scrollContainer.scrollHeight, behavior: "instant" });
+                    }
+                });
+            }
         },
         dataSaved,
         updateInterval
--- a/apps/client/src/widgets/type_widgets/text/EditableText.tsx
+++ b/apps/client/src/widgets/type_widgets/text/EditableText.tsx
@@ -70,6 +70,10 @@ export default function EditableText({ note, parentComponent, ntxId, noteContext
                     viewScope.bookmark = undefined;
                 });
             }
+
+            if (note?.hasLabel("openAtBottom")) {
+                scrollEditorToBottom(watchdogRef.current?.editor as CKTextEditor | undefined, containerRef.current);
+            }
         },
         dataSaved(savedData) {
             // Store back the saved data in order to retrieve it in case the CKEditor crashes.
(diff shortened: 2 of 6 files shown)


maphew (changes requested):
I found three behavior gaps:

1. The new checks use `hasLabel("openAtBottom")`, so `#openAtBottom=false` still enables the behavior. This also prevents a note from disabling an inherited setting. Advise using the existing `isLabelTruthy()` helper.

2. The implementation exists only in `EditableText` and `EditableCode`. `ReadOnlyText` and `ReadOnlyCode` are separate rendering paths and never inspect the label, although the documentation promises the behavior for text and code notes without an editable-only restriction. Advise applying the scroll behavior to both paths, while keeping cursor movement editable-only.

3. Text-note loading queues a bookmark scroll and then independently queues the bottom scroll. When navigation includes `?bookmark=...`, the later bottom callback can override the explicit destination. Advise giving bookmark navigation priority and skipping the bottom scroll for that load.

eliandoran:
Hi,

Thanks for the PR, but I believe this one is better served as a custom widget instead of being part of the core.

Here's a draft of how the widget could look like (untested and probably needs some tweaking):

```js
class OpenAtBottomWidget extends api.NoteContextAwareWidget {
    get parentWidget() { return "center-pane"; }
    doRender() { this.$widget = $("<div>"); return this.$widget; }

    async noteSwitched() {
        if (!this.note?.isLabelTruthy("openAtBottom")) return;
        const editor = await api.getActiveContextTextEditor();
        editor?.model.change((w) =>
            w.setSelection(w.createPositionAt(editor.model.document.getRoot(), "end")));
    }
}
module.exports = new OpenAtBottomWidget();
```

PR #10358, opened 2026-07-02
perf(sync): fill the pull response byte cap across row fetches

eliandoran:
## Summary

Follow-up to #10351. That PR raised the pull response cap to 8 MB, but `getChanged` still fetches a single `LIMIT 1000` batch per request — so on metadata-dense ranges (notes/branches/attributes at ~300 bytes each) a response carries only ~0.3 MB, and the **row limit, not the byte cap, determines the number of round-trips**. This was directly observable in #10351's benchmarks: request count fell sublinearly with cap size, and 105 chunks in the 1 MB baseline returned exactly 1000 rows at ~0.75 MB.

This loops the 1000-row fetch inside `getChanged` until the response byte estimate crosses the cap or the range is exhausted. Cursor semantics are preserved: the client-visible `lastEntityChangeId` only advances over records actually included in the response (plus batches consisting entirely of the client's own changes, as before), so a byte-capped mid-batch stop re-serves the remainder on the next request.

## Benchmarks (initial sync, request counts)

| Database shape | Before | After |
|---|---:|---:|
| **Metadata-heavy fixture** (180k small changes: 60k notes + branches + attributes) | 184 requests | **8 requests (23×)** |
| **Blob-heavy 2 GB document** (from #10351) | 332 requests | **207 requests (1.6×)** |

The blob-heavy case is near its byte-bound floor (~311 responses must carry blob content that already fills the cap), so the win there is modest — the target of this change is metadata-heavy databases (many notes, few attachments), where round-trips over a high-latency link are the dominant initial-sync cost. Both benchmark targets reproduce their source exactly (entity counts and content bytes verified).

## Testing

- New route test: 1100 metadata rows come back in **one** response (two LIMIT-1000 fetches), with correct cursor and outstanding count — fails on the old single-fetch code.
- Existing cursor-semantics tests (own-instance skip, byte-cap boundary, empty-batch cursor advance) all pass unchanged: 47/47 sync specs, full server suite 3786 passed, core typecheck clean.

### DIFF
--- a/apps/server/e2e/sync/sync-instance.ts
+++ b/apps/server/e2e/sync/sync-instance.ts
@@ -1,0 +1,253 @@
+import { type ChildProcess, spawn } from "child_process";
+import { join } from "path";
+
+/**
+ * A real Trilium server instance (the production `dist/main.cjs` bundle, same as the e2e
+ * webServer) spawned against its own data directory, plus the HTTP plumbing needed to drive
+ * the sync protocol end-to-end: document setup, password, ETAPI seeding, sync-from-server and
+ * completion polling.
+ *
+ * Reused by sync e2e specs; extend it rather than hand-rolling fetch calls in tests.
+ */
+export default class SyncInstance {
+    readonly dataDir: string;
+    readonly port: number;
+    readonly baseUrl: string;
+
+    private child?: ChildProcess;
+    private etapiToken?: string;
+
+    constructor(dataDir: string, port: number) {
+        this.dataDir = dataDir;
+        this.port = port;
+        this.baseUrl = `http://127.0.0.1:${port}`;
+    }
+
+    /** Spawns the server and waits until it responds. Safe to call again after {@link stop}. */
+    async start() {
+        const serverDir = join(__dirname, "..", "..");
+
+        this.child = spawn(process.execPath, ["dist/main.cjs"], {
+            cwd: serverDir,
+            env: {
+                ...process.env,
+                TRILIUM_DATA_DIR: this.dataDir,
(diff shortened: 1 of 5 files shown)


eliandoran:
Added a companion commit: a reusable end-to-end test of the sync protocol (`apps/server/e2e/sync/`). Two real production-bundle instances over HTTP: ETAPI-seeded source (sized to cross the 1000-row fetch batches many times + one record above the byte cap, i.e. it directly guards this PR's change), full initial sync, incremental sync in both directions, and an offline DB comparison. Runs in ~25 s in the server e2e project; skipped in Docker mode.

One subtlety it surfaced, for the record: locally-saved text notes are stored as TEXT while synced content is stored as a Buffer, so SQLite `LENGTH(content)` counts characters on one side and bytes on the other — byte comparisons must `CAST(content AS BLOB)`. Same bytes either way; documented in the spec.

eliandoran:
Added `perf(sync): cache content hashes behind an entity_changes fingerprint` — the recurring `/api/sync/check` cost (~470 ms per call on a 2 GB document, paid by every client every ~60 s on both sides) drops to ~27 ms when nothing changed, via a (COUNT, MAX(id)) fingerprint over `entity_changes` that any write path moves. Unit tests pin the raw-SQL-writer invalidation; the sync e2e and full suite are green.

Also prototyped and **dropped** a streamed pull-response serialization: measured peak RSS was *worse* than the single `JSON.stringify` at the 8 MB response size (Node's write-queue overhead exceeds the big-string cost), and CPU was within noise — so it stays out.

PR #10399, opened 2026-07-05
feat(options): add "Wipe database" action to Advanced settings

eliandoran:
## What

Adds a **"Wipe the database"** action to the **Advanced** options page. It permanently deletes every note, attachment, revision, attribute and setting and rebuilds an empty database — a clean reset without reinstalling.

This is especially useful on **standalone/mobile**, where the database lives in the WebView's OPFS with no filesystem access, so there's no easy way to reset a device today.

## How it works

- **Purely at the SQL layer** — drops all tables, then rebuilds an empty document via `createInitialDatabase` (skipping demo content) and reloads becca in-process. This single code path works identically on **server, desktop and standalone/mobile**.
- The route is registered in the **shared core route builder** (`buildSharedApiRoutes`), so it is exposed by both the Express server and the standalone browser router — no per-platform plumbing.
- Guarded by a verbose confirmation token (`?really=…`), mirroring the existing `password/reset` endpoint (guards against accidental calls; the usual auth middleware handles security).

## UX

- Destructive-tinted button in a dedicated **"Wipe database"** section of Advanced options.
- Confirmation modal with a caution admonition, a "make a backup first" hint, and a **5-second countdown** that keeps the confirm button disabled — the user has to pause and read before it becomes actionable.
- On success, the client reloads so it rebuilds its own caches (froca) and lands on the fresh setup/login flow.

## Testing

- `sql_init.spec.ts` — real-DB wipe: proves user content is gone and the DB is left valid/initialized (root note, schema, `initialized` flag).
- `database.spec.ts` — route confirmation guard (rejects wrong token, wipes on correct token). Runs under **both** the node (better-sqlite3) and standalone (WASM) suites via `CoreApiTester`.
- `wipe_database.spec.tsx` — client dialog: countdown gating, posts with the token + reloads on success, no reload on failure.

`pnpm typecheck` clean; the full core suite (2956 tests) passes under the WASM/standalone runtime, confirming the cross-platform path.

## Notes / open questions

- Shown on all builds (like the neighbouring vacuum/integrity/anonymize actions). On a shared multi-user **server** a wipe is inherently dangerous; happy to gate it to standalone+desktop if preferred.
- Sync clusters: a wipe currently clears local data and would re-pull on next sync. If "truly forget everything" semantics are wanted for synced instances, that can be a follow-up.

### DIFF
--- a/apps/client/src/translations/en/translation.json
+++ b/apps/client/src/translations/en/translation.json
@@ -1422,6 +1422,22 @@
     "vacuuming_database": "Vacuuming database...",
     "database_vacuumed": "Database has been vacuumed"
   },
+  "wipe_database": {
+    "title": "Wipe database",
+    "description": "Permanently delete every note and start over with an empty database. Useful for resetting a device without reinstalling.",
+    "wipe_label": "Wipe the database",
+    "wipe_description": "Deletes all notes, attachments, revisions and settings from this device, then rebuilds an empty database. This cannot be undone.",
+    "wipe_button": "Wipe database…",
+    "dialog_title": "Wipe the database?",
+    "warning_heading": "This permanently deletes everything and cannot be undone.",
+    "warning_body": "All notes, attachments, revisions, attributes and settings on this device will be erased and replaced with a fresh, empty database.",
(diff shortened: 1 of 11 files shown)


eliandoran:
Superseded by the "Start over" functionality.

PR #10439, opened 2026-07-09
feat(mobile): route iOS local API requests through a native WKURLSchemeHandler

eliandoran:
## Motivation

On iOS, Capacitor loads the app on the `capacitor://` scheme, where WebKit refuses to register a service worker — so the request routing the SW performs on Android/web (page ↔ in-page SQLite worker) was reimplemented as in-page interceptors, one per request-initiation channel: `fetch`, XHR, ``, and most recently `<style>`/`<link>` (icon pack fonts, custom themes). Each new channel had to be discovered as a user-facing bug; engine-initiated loads such as audio/video sources, iframes, and CSS `@import` were still uncovered.

## Approach

Intercept at the **WKURLSchemeHandler level** instead — below the content layer, where every request on the `capacitor://` scheme arrives regardless of which engine subsystem issued it. Local API paths (`/api/`, `/sync/`, `/search/`, `/bootstrap`) are forwarded into the page, answered by the SQLite worker, and completed natively:

```
WebKit engine → TriliumAssetHandler (native)
  → evaluateJavaScript(window.__triliumNativeRequest(payload))
    → localFetch → SQLite worker
      → postMessage("triliumScheme" script message)
        → native completes the WKURLSchemeTask
```

All four in-page interceptors are removed.

### Native notes

- Capacitor's `loadView()` is `final` and WebKit raises `NSInvalidArgumentException` when re-registering a scheme handler, so the registered `WebViewAssetHandler` instance is re-typed to `TriliumAssetHandler` via `object_setClass` (same isa-swizzle pattern the app already uses for `WKContentView`). The subclass holds no stored properties; state lives in `TriliumSchemeBridge`.
- Requests intercepted before the JS bridge loads are queued and flushed on its `ready` message.
- Cancelled tasks leave the pending map in `stop` (completing a stopped `WKURLSchemeTask` raises).
- POST bodies are drained from `httpBodyStream` when `httpBody` is nil; `Content-Length` is normalized to the delivered bytes.
- Bodies cross the bridge as base64 in both directions.

## Verification

- New `ios-native-bridge.spec.ts` (8 tests: roundtrip, body decode, GET-body guard, >32 KB binary integrity, error paths, prefix-mismatch guard); full standalone suite green (210 files / 2,985 tests).
- Full Xcode simulator build succeeds (Capacitor 8.4.1, iOS 18 SDK).

## Outstanding before marking ready

- [ ] On-device validation: boot, icon pack fonts, images, note saves (body path), media, and a full sync of a large DB
- [ ] Perf measurement: every request now crosses to native with base64 marshalling on the main thread; if large sync/attachments regress, the planned mitigation is a hybrid fast path (fetch/XHR handled in-page again, scheme handler as catch-all for engine-initiated loads)
- [ ] Consider a native watchdog timeout for pending tasks and failing pending tasks on page navigation
- [ ] Update the CLAUDE.md mobile section, which still documents the removed interceptors


maphew (commented):
The native interception path needs a per-navigation lifecycle. `isJsReady` remains true after the first ready message, so requests during a later page reload can be delivered before the new page installs `window.__triliumNativeRequest` and are then lost. Cancellation also removes the pending task but leaves its request in `queuedRequests`, so cancelled work still runs after readiness. Advise resetting bridge state for each navigation and pruning queued requests on cancellation before device validation.

PR #10477, opened 2026-07-12
fix(client): add-link dialog drops the link on a fast double-Enter

### LINKED ISSUE #5669: Add link dialog fails to link if you press enter twice too quickly
### Description

Steps to reproduce:

1. Open "Add link" dialog, search for a note, select with arrow keys
2. Press enter twice, quickly

Expected behaviour:

Same as when pressing enter twice slowly, a link is created.

Actual behaviour:

"Choose note type" dialog for new note is opened.

Javascript event handlers are run single-threaded, so it is in principle possible to fix this issue robustly. The first key event handler needs to (synchronously) set everything up for the second key event handler.

### TriliumNext Version

0.92.7 (this is a long-standing issue)

Other Linux

### What is your setup?

Local + server sync

### Operating System Version

NixOS 25.05 unstable

### Error logs

_No response_

mvanhorn:
## Summary

Fixes the "Add link" dialog dropping the link when Enter is pressed twice quickly. Pressing Enter a second time before the autocomplete dropdown finished closing let a stale "create-note" row get auto-selected, so instead of creating the link the "Choose note type" dialog opened.

## Why

Selecting a suggestion sets the field value and closes the dropdown asynchronously; a second fast Enter arrived while a re-query had re-populated the create-note row and consumed the keypress. The fix records a terminal-selection state the moment a note or external link is selected and, on a plain Enter, stops autocomplete from consuming that stale create-note row while leaving normal form submission intact.

Because the guard would otherwise persist across a later programmatic query (autocomplete.js's `setVal` does not emit an `input` event), it is stored on the element and cleared whenever a fresh query starts -- `clearText`, `setText`, `showRecentNotes`, `showAllCommands`, and `fullTextSearch`. Without that reset, selecting a note and then opening Recent Notes would leave Enter swallowed until the field was manually edited.

## Testing

`pnpm --filter client test src/services/note_autocomplete.spec.ts` -- 62 tests pass. New coverage:

- a stale create-note Enter right after selecting a note is ignored, and create-note selection re-enables once the input is edited;
- each programmatic query helper (Recent Notes, set text, show-all-commands, full-text search) clears the terminal-selection guard so keyboard selection works again;
- the guard persists on a bare Enter (it is what the guard swallows) until a fresh query clears it.

Closes #5669

### DIFF
--- a/apps/client/src/services/note_autocomplete.spec.ts
+++ b/apps/client/src/services/note_autocomplete.spec.ts
@@ -607,6 +607,20 @@ describe("initNoteAutocomplete wiring", () => {
         expect(selected).not.toHaveBeenCalled();
     });
 
+    it("leaves Enter available for form submission after a terminal selection", () => {
+        const $el = makeEl();
+        noteAutocomplete.initNoteAutocomplete($el);
+        ($el as any).trigger("autocomplete:selected", { notePath: "root/n", noteTitle: "N" });
+
+        const autocompleteKeydown = vi.fn((event: JQuery.KeyDownEvent) => event.preventDefault());
+        $el.on("keydown.aa", autocompleteKeydown);
+        const event = $.Event("keydown", { key: "Enter" });
+        $el.trigger(event);
+
+        expect(autocompleteKeydown).not.toHaveBeenCalled();
+        expect(event.isDefaultPrevented()).toBe(false);
+    });
+
     it("composition end re-sets the autocomplete value", () => {
         const $el = makeEl();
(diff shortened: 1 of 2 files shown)


eliandoran:
Hi,

What are the steps to reproduce this issue?

I can't reproduce the steps as described in #5669

eliandoran:
@mvanhorn, thanks for your implementation, I've tried to simplify it a bit in #10481.
Let me know in the next nightly or release if you still have issues with the auto-complete.

Issue #9754, opened 2026-05-15
JavaScript error while OCR processing existing image files


jshprentz:
### Description

Batch OCR processing existing image files stopped after about 45 minutes with a popup JavaScript error message.

<img width="569" height="280" alt="Image" src="https://github.com/user-attachments/assets/ccd788eb-a3a0-4cf7-a7cf-3ddb93e50b65" />

After clicking OK, the popup reappeared again and again.

The Media Options Text Extraction panel showed:

<img width="799" height="338" alt="Image" src="https://github.com/user-attachments/assets/deb79f1f-9045-4f53-b0d2-014574db7f42" />

The repeated error popups made the rest of Trilium unresponsive. I closed Trilium from the Windows taskbar. Batch OCR did not resume when I reopened Trilium.

### TriliumNext Version

v0.103.0

### What operating system are you using?

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 11 Pro Version 25H2 OS build 26200.8457

### Error logs

[trilium-2026-05-14.log](https://github.com/user-attachments/files/27783003/trilium-2026-05-14.log) is attached.

Trilium 0.103.0 starts for the first time at 17:46:43.364, an upgrade from Trilum 0.102.2.

Batch OCR processing starts at 19:38:57.416. The first of many repeated failed OCR attempts was logged at 19:53:52.569. I noticed the popup message and clicked OK after about 10 minutes.
```
19:53:52.569 Starting OCR text extraction for MIME type: image/jpg with language: eng
19:53:52.569 Starting image OCR text extraction (language: eng)...
19:53:52.579 ERROR: Image OCR text extraction failed: Error: Error attempting to read image.
20:03:04.250 ERROR: Failed to process OCR for attachment AA1sFERQlIHq: Error: Error attempting to read image.
20:03:04.250 ERROR: Failed to process OCR for attachment AA1sFERQlIHq: Error: Error attempting to read image.
```

eliandoran:
Interesting that the OCR took so long. Do you have a large amount of images in your database?

I suppose the issue you are having is that there is an unsupported or broken image in your database. We'll implement the necessary fix of ignoring this kind of errors and simply continue processing.

jshprentz:
@eliandoran asked "Do you have a large amount of images in your database?"

I don't know how to see the number of images or the number of notes in my Trilium database. I don't know how to judge whether either is a "large amount." These log lines may provide a clue:
```
[...]

PR #10499, opened 2026-07-13
feat(ocr): extract text from scanned PDFs via per-page OCR

eliandoran:
## Summary

Scanned PDFs produced no searchable text: the PDF processor only pulled the embedded text layer, which such documents lack. This adds an OCR fallback that rasterizes those pages through PDFium and reads them with the existing Tesseract pipeline.

The scope grew beyond scans. Any PDF whose text cannot be selected is now readable — including documents whose text was converted to vector outlines, which no text extractor can reach. A hotel voucher of that shape carries 2559 `constructPath` operations and not one text-showing operator; both pdf.js and PDFium report zero characters in it. It now yields **3284 characters at 0.93 confidence** across both pages, including the guest name, hotel, booking reference and room type.

## Reading the page instead of the images it paints

The obvious approach — pull out each page's embedded images and OCR those — only works when the page *is* one scanned image. On that voucher the single embedded image is the 180×91 logo, so OCR returned `"SMART TOURS"` and nothing else, and stored it as the document's text.

Rendering the page covers everything image extraction did, plus several cases it could not:

- text drawn as vector outlines
- a scan split across image strips, which extraction OCRs separately and joins in operator order rather than reading order
- a page whose `/Rotate` the images themselves know nothing about
- stamps and annotations drawn over a scan

It also bounds the work: a page becomes a fixed number of pixels regardless of the resolution it was scanned at. The image path had no such bound — a 600 dpi scan would have allocated a 140 MB RGBA buffer.

It costs nothing. End to end (PDF bytes → a PNG Tesseract can read) the two measured 175–240 ms and 186–256 ms on the sample scan; PNG encoding dominates both. Recognition is unchanged — 94 confidence and 718 characters from the embedded image, 93 and 716 from the render, same text. So image extraction is removed outright rather than kept as a fast path, and with it the channel conversion, the minimum-image-dimension heuristic and the per-image loop.

## Deciding which pages need OCR

A page was treated as scanned only when its embedded text fell under 16 characters. Scanners rarely leave a page that bare — a header, a watermark or a partial text layer clears any bound low enough to be safe, and the body of the scan is never read. The voucher's text layer is the single line `ACME Corp — Invoice 2024-11-03`: 30 characters, and the page was skipped.

The bound now scales with page area. At 0.0005 characters per square point it comes to about 240 characters on Letter and A4 — far below a sparse but genuine text page, far above a header. Geometry comes from the viewport pdf.js already parsed; a document that will not report it is classified against a Letter-sized default rather than skipped.

Detection stays per page, so mixed documents are handled correctly, and a text page misjudged as a scan is lossless — the render yields no text and the page keeps what it had.

## Rendering choices, and why

**Colour, not grayscale.** PDFium's own grayscale conversion flattens coloured text into its background — on the demo document it swallows the entire "Organize your thoughts" panel, 60 characters that colour rendering keeps and that no increase in scale recovers. Rendering BGRA and letting Tesseract do its own grey conversion recovers them at no measurable cost: 1027 ms vs 1068 ms end to end on the sample scan, identical PNG size, identical recognition.

**Scale 2** (~144 dpi) is the knee of the quality curve: confidence 82 at scale 1, 93 at 1.5, 94 at 2, then flat while the pixels to encode and read double again. It matches OCR-ing the embedded image at its native resolution.

**The channel swap happens in place.** BGRA already carries four bytes per pixel, so producing the RGBA bitmap Jimp encodes needs no second buffer — the page belongs to the call that rasterized it and is dropped as soon as the PNG exists. Saves an 8 MB allocation and a full-page copy per page, byte-identical output. Peak transient memory per page is lower than the grayscale path it replaced.

## Why PDFium, and what it costs

PDFium ships as WebAssembly, which sidesteps what a native rasterizer would have cost. `@napi-rs/canvas` is 30 MB per platform — 20 MB of Skia statically linked against libpng, libjpeg-turbo, libavif, harfbuzz and an SVG renderer, plus 11 MB of ICU data the desktop build already carries a copy of — and it would need an esbuild external plus a per-platform binary in `copyNodeModules`.

The wasm is 4 MB, identical on every target, and bundles normally: `main.mjs` carries no reference to it, and it loads from a lazy chunk on the first page that needs rasterizing.

PDFium's loader resolves the wasm against `import.meta.url`, which in the split ESM bundle is a hash-named file under `chunks/` rather than anywhere the wasm was copied. The bytes are therefore read from `RESOURCE_DIR` and handed to `init()` — the same two paths `core_assets.ts` uses for `schema.sql`: copied beside the server's assets by the build, resolved through `node_modules` when running from source.

## Shared Tesseract recognizer

The long-lived Tesseract worker and the confidence filtering are lifted out of `ImageProcessor` into a shared `TesseractRecognizer`, so image notes and PDF pages reuse one worker, its loaded language data and one code path. Calls are serialized through a promise queue, so a manual reprocess arriving mid-batch cannot terminate a worker another call is mid-recognition with, and only one CPU-heavy job runs at a time.

Image behaviour is unchanged, including the `#9754` Electron `errorHandler` and the language-switch/terminate logic.

## Batch selection

`getBlobsNeedingOCR()` continues to select only `NULL` text representations. Treating a stored `''` ("processed, found no text") as "needs OCR" would leave blank, vector-only or encrypted PDFs with no terminal state, re-rasterizing them on every batch run. Scanned PDFs stored before this change can be re-run individually via *Process OCR*, which forces reprocessing.

## Testing

- `pdf_processor.integration.spec.ts` runs the pipeline **unmocked** — real unpdf extraction, real PDFium rendering, a real Tesseract worker — against a committed two-page image-only PDF, asserting phrases from both pages so multi-page OCR is exercised end to end.
- `tesseract_recognizer.spec.ts` covers worker lifecycle, the serialization queue and confidence filtering against the nested `blocks → paragraphs → lines` shape tesseract.js actually returns.
- `pdf_processor.spec.ts` covers the embedded-text fast path, per-page detection including the header-only scan and the area scaling, unreadable geometry, the BGRA→RGBA conversion, render failure, and the page cap.
- The production wasm path was verified by running the integration spec with `TRILIUM_RESOURCE_DIR` pointed at `dist/`, not just the source-tree fallback.
- Full OCR suite green; `pnpm typecheck` clean.

## Documentation

The User Guide OCR page now describes PDF support accurately — text PDFs, scans and other PDFs whose text cannot be selected, mixed documents, and the 50-page OCR cap — with a note that scanned-PDF OCR arrives in v0.106.0 and that pre-existing files need an individual *Process OCR*.

## Follow-ups (not in this PR)

- `MAX_OCR_PAGES` is a constant (50); could become a user option.
- A blank page is still rendered and recognized before being found empty; a cheap uniform-colour check on the bitmap could skip it.
- PDFium could serve the standalone build too, but the wasm would then be fetched by the browser and needs a content-hashed URL — `init({ wasmBinary })` makes that straightforward, unlike the unhashed filename `@sqlite.org/sqlite-wasm` requires.


maphew (changes requested):
A fixed 16-character embedded-text threshold can skip OCR for scanned pages that contain only a header, watermark, or partial text layer. Advise detecting meaningful page coverage or combining OCR when a substantial page image exists.

Advise also propagating OCR-required page failures instead of converting every exception to empty text and then persisting a partial successful result.

PR #10526, opened 2026-07-16
Update Preact Components for Scripts

BeatLink:
# Expose reusable Preact widgets & new form controls to scripts

## Summary

Promotes Trilium's internal table and calendar view primitives into the shared reusable-component folder, adds a family of new form-control components (ported from the `libformcontrols` addon), and exposes all of them on the public frontend Script API (`trilium:preact`) so custom widgets and Render Notes can use them. Includes unit tests and User Guide showcase updates.

### New reusable components (`apps/client/src/widgets/react/`)

- **`Table`** — generic [Tabulator](https://tabulator.info/) data grid, decoupled from the note/collection model (re-exports the existing wrapper).
- **`Calendar`** — generic [FullCalendar](https://fullcalendar.io/) wrapper, decoupled from the note/collection model (re-exports the existing wrapper).
- **`FormNumber` / `FormTime` / `FormDatetime`** — thin wrappers over `FormTextBox` that lock in `type="number"` / `"time"` / `"datetime-local"`, so the browser's native pickers are used while inheriting `FormTextBox`'s controlled-input handling, min/max clamping, and `onChange(value, validity)` / `onBlur` / `inputRef` API.
- **`FormToggleButton`** — a boolean toggle rendered as a pill button, styled with the theme's `btn` / `btn-primary` + `active` classes (matching `Button` / `ThemeModeSelector`), *not* Bootstrap's unstyled `btn-check`. Uses `aria-pressed` for accessibility.
- **`ColorPicker`** — combines preset color swatches with the browser's native `<input type="color">`, using the theme `Dropdown` for its popover. Value is a CSS color string (`onChange("")` clears).

### Public Script API exposure

Updated `apps/client/src/services/frontend_script_api_preact.ts` and the ambient type surface in `packages/commons/src/lib/script_api_preact.ts`:

- All of the above, plus **`ExternallyControlledCollapsible`** (the controlled variant of `Collapsible`), are now available via `import { … } from "trilium:preact"`.
- Both the runtime `preactAPI` and the ambient type surface are updated in lockstep; the existing drift-guard passes.

### Minimal base-source changes

Added `export` to three existing interfaces so the new wrappers/re-exports can reference them — no behavioral changes:

- `TableProps` (`collections/table/tabulator.tsx`)
- `CalendarProps` (`collections/calendar/calendar.tsx`)
- `FormTextBoxProps` (`react/FormTextBox.tsx`)

### Tests

15 new tests, using happy-dom + raw Preact `render()` (zero new deps):

- `FormInputWrappers.spec.tsx` — `FormNumber` / `FormTime` / `FormDatetime`
- `FormToggleButton.spec.tsx`
- `ColorPicker.spec.tsx`
- `collection_reexports.spec.ts` — `Table` / `Calendar` re-export identity
- `test/render.ts` — shared render helper

### Docs

- `CLAUDE.md`: documented the new components; corrected a stale `CollapsibleSection` reference to `Collapsible`.
- User Guide **Widget showcase**: demonstrates the new form controls and `Table` (`Calendar` noted but not demoed, as it requires an explicit FullCalendar `plugins` array).

## Notes / limitations

- `Table` and `Calendar` are exposed with **loosened ambient types** (their Tabulator/FullCalendar option surfaces are passed through as index signatures), consistent with the Script API's documented typing philosophy for library-heavy props.
- `Calendar` is a thin wrapper — consumers must supply a `plugins` array to render a view.

## Verification

- `pnpm typecheck` — clean (drift guard green)
- `pnpm exec eslint` on changed files — clean
- New unit tests — 15/15 passing

## Files changed

21 files, +681 / −40.


BeatLink:
@gemini-code-assist please rereview

apps/client/src/widgets/react/Calendar.tsx
```diff
@@ -0,0 +1,9 @@
+/**
+ * Public re-export of the generic FullCalendar-based calendar wrapper as a shared Preact widget.
+ *
+ * The implementation lives at `../collections/calendar/calendar` for historical reasons; this module
+ * exposes it from the reusable `widgets/react/` component folder so features outside the note
+ * collection view can consume it without reaching into `collections/calendar`. It is decoupled from
+ * Trilium's note model — it deals purely in FullCalendar options and a `calendarRef`.
+ */
+export { default, type CalendarProps } from "../collections/calendar/calendar";
```
eliandoran: There's no benefit of having a wrapper class, simply export it in `frontend_script_api_preact.ts` instead.

apps/client/src/widgets/react/collection_reexports.spec.ts
eliandoran, line 1: Redundant test, consider rewriting the assertions and relocating to the Preact frontend.

apps/client/src/widgets/react/FormDatetime.tsx
eliandoran, line 1: What's the purpose of this wrapper class? Why not simply use `<FormTextBox>` with `type=datetime-local` directly in the script instead of the API?

We don't expose additional widgets to the front end API, only widgets that are actually used by Trilium.

apps/client/src/widgets/react/FormNumber.tsx
eliandoran, line 1: What's the purpose of this wrapper class? Why not simply use `<FormTextBox>` with `type=number` directly in the script instead of the API?

We don't expose additional widgets to the front end API, only widgets that are actually used by Trilium.

apps/client/src/widgets/react/FormTime.tsx
eliandoran, line 1: What's the purpose of this wrapper class? Why not simply use `<FormTime>` with `type=time` directly in the script instead of the API?

We don't expose additional widgets to the front end API, only widgets that are actually used by Trilium.

apps/client/src/widgets/react/FormToggleButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * A checkbox rendered as a pill button (pressed = on), consistent with Trilium's theme.
+ *
+ * Uses the same `btn`/`btn-primary`/`btn-secondary` + `active` classes as {@link Button} and the existing
+ * in-repo toggle buttons (e.g. `ThemeModeSelector`) rather than Bootstrap's `btn-check`, which is not
+ * styled in Trilium's theme. Accessibility is conveyed via `aria-pressed`.
+ */
+export default function FormToggleButton({ label, icon, currentValue, onChange, kind = "primary", size, disabled, title, className }: FormToggleButtonProps) {
```
eliandoran: Same here, no extra widgets.

apps/client/src/widgets/react/Table.tsx
eliandoran, line 1: Wrapper class, integrate directly in `frontend_script_api_preact`.

eliandoran (changes requested):

BeatLink:
@eliandoran Made the changes you requested.

apps/client/src/widgets/react/ColorPicker.tsx
eliandoran, line 1: Hmm... I think I missed this one during my review,

What's the purpose of this color picker?

BeatLink: lets widgets pick colors. Mainly i use it for the color of a note since there is no promoted attribute picker for it

eliandoran: @BeatLink , makes sense, but wouldn't it make more sense to keep it as a [reusable Preact component](https://docs.triliumnotes.org/user-guide/scripts/frontend-basics/preac/component-libraries) instead of shipping it with the app (considering it's not used by the app)?

BeatLink: Fair enough. I guess i could restructure it to use the same color picker from the note context menu. Is that a preact component?

eliandoran: @BeatLink , yes, it's a Preact component. If you think it's useful and not too coupled to note color selection, you can give it a try.

BeatLink: Made the changes. I had to generalize the picker a bit.

eliandoran (approved):

PR #10529, opened 2026-07-17
feat(llm): add GitHub Copilot provider via Copilot CLI ACP mode

eliandoran:
Adds a `copilot-agent` LLM provider that lets users with a **GitHub Copilot subscription** use Trilium's built-in chat without an API key — mirroring the existing Claude Agent design, but driving the user's installed **GitHub Copilot CLI in ACP mode** (`copilot --acp`) as a subprocess. Authentication is owned entirely by the CLI (`copilot login`, or credentials shared with another editor integration on the machine).

## What's here

- **`acp_client.ts`** — dependency-free Agent Client Protocol client (newline-delimited JSON-RPC over stdio): framing, request/response correlation, agent→client requests, subprocess lifecycle.
- **`copilot_agent.ts`** — the provider. Implements `chatChunks()`: maps chat notes to ACP sessions (resume when the transcript matches, reseed via history replay when it diverged), streams `session/update` notifications into `LlmStreamChunk`s, selects models via `session/set_model`, and exposes the CLI's live model catalog with premium-request multipliers mapped onto `costMultiplier`.
- **`copilot_binary.ts`** — bring-your-own-binary resolver (`TRILIUM_COPILOT_PATH` → `copilot` on PATH), Windows `.cmd`-shim aware.
- **`copilot_mcp_endpoint.ts`** — private loopback HTTP MCP endpoint (random port + 128-bit secret path) exposing note tools to the agent. ACP takes MCP servers by URL, not in-process, and this is deliberately independent of the user-facing `mcpEnabled` toggle.
- **`transcript.ts`** — hash/replay helpers extracted from `claude_agent.ts` and shared between both agent providers.
- Registration server-side (`llm/index.ts`) and client-side (`AddProviderModal` card, icon, translations).

## Security

Note tools are pre-approved via `--allow-tool=trilium` and built-in file/shell/network tools are denied via `--deny-tool`. The permission callback is **fail-closed**: it denies anything that reaches it, so the worst case is a note tool failing to run — never a shell command on the server host. Verified live against CLI 1.0.71 (`--deny-tool=shell` blocks shell before any permission request reaches the callback).

## Validation

- Full suite **4031 passing** (80+ new unit tests across `acp_client`, `copilot_agent`, `copilot_binary`), typecheck clean, lint clean.
- Live end-to-end against the real CLI: initialize → session/new → set_model → streaming reply; real permission-request payload confirmed rejected.

## Known constraints (draft — needs decision before merge)

1. **Note tools require an org policy on Copilot Business/Enterprise.** The CLI skips third-party MCP servers unless the org's "MCP servers in Copilot" policy is enabled (off by default). Chat works regardless; the note-tools path is gated by this and is currently covered by **unit tests only** — it could not be live-tested on a Business account.
2. The permission model is fail-closed by design (see above) because the CLI presents tool calls with opaque IDs and human-friendly titles, making name-based allow-listing unreliable.


eliandoran:
@greptile-apps

PR #10553, opened 2026-07-18
feat(llm): add semantic note search via local Ollama embeddings

Kureii:
### Summary
Add a `semantic_search_notes` tool that finds notes **by meaning** rather than exact keywords, using embeddings from a local Ollama instance. This complements the keyword-based `search_notes` tool and is the retrieval building block for RAG-style workflows with local models (together with the Knowledge Base mode from #9343).

### Changes
- **`embeddings.ts`**: client for Ollama's `/api/embed` endpoint with request batching; embeddings are computed lazily and cached **in memory** keyed by `noteId + blobId`, so a note is only re-embedded when its content changes and nothing is persisted to the database
- **`semantic_search_tool.ts`**: async AI SDK tool registered directly in the provider tool set (like the web search tools — registry tools must stay synchronous for MCP); searches the whole tree or a subtree (`ancestorNoteId`), candidates capped at the 300 most recently modified notes
- Only enabled when an **Ollama provider** is configured; returns a descriptive error when the embedding model is missing (`ollama pull nomic-embed-text`)
- New `llmEmbeddingModel` option (default `nomic-embed-text`) with a settings row under AI / LLM
- `search_query:` / `search_document:` task prefixes are applied automatically for nomic models

### Notes
- Stacked on #9338 (needs the Ollama provider).
- Verified end-to-end against a local Ollama instance: query embedding + cosine ranking returns the semantically relevant note first with a clear margin.

### Possible follow-ups
- Persistent embedding index (DB table) for large vaults
- Embedding providers other than Ollama (OpenAI, Google)


maphew (changes requested):
Semantic search automatically posts note titles and content to the first configured Ollama URL, while Ollama configuration accepts any HTTP(S) host. Advise requiring explicit consent or restricting this path to loopback unless remote note disclosure is enabled.

Advise also avoiding or reporting the silent newest-300 candidate cutoff. An older relevant note is otherwise never embedded or ranked, so the tool can return a plausible but incomplete answer without disclosing the truncation.

PR #10554, opened 2026-07-18
perf(llm): keep agent subprocesses alive across chat turns

eliandoran:
Both agent providers spawned a CLI subprocess, ran a handshake, and threw it all away on **every** chat turn. This keeps the process warm between turns, cutting seconds of pure overhead off each message — none of it model latency.

Measured on Windows against the real CLIs (times to first token):

| provider | before | after | saved |
|---|---|---|---|
| Copilot | 6912 ms | 3078 ms | **3.8 s (55%)** |
| Claude  | 2265 ms |  866 ms | **1.4 s (62%)** |

Two commits, one per provider — they solve the same problem but the mechanisms have nothing in common.

## Copilot: one shared client, many sessions

ACP is built for this — one connection hosts many sessions addressed by `sessionId`, and prompting is a call on a session that already exists. The client is now shared process-wide and reaped after 5 minutes idle. The protocol trace tells the story:

```
cold  [initialize, session/new, session/set_model, session/prompt]
warm  [session/set_model, session/prompt]
```

The `session/new` alone was ~2.4 s: it refreshes auth and fetches the model list.

Consequences worth reviewing:

- **Cancellation is now in-band.** The client is shared, so disposing it on abort would kill every other chat's turn.
- **Session mappings record the process generation.** A session lives inside its process, so after a reap or crash the mapping is stale and the turn falls back to `session/load`. Prompting a live session directly is what makes the warm path cheap — `session/load` on an already-loaded session is rejected by the CLI (`"already loaded"`).
- **Flipping note access starts a new session**: MCP servers are fixed at `session/new`.
- `AcpClient` gained an `onExit` hook so a dead subprocess evicts itself.

## Claude: streaming input mode

`query({ prompt: "text" })` binds the subprocess to the returned iterator, so it dies when iteration ends. Passing an `AsyncIterable` instead keeps one subprocess per chat. That is also the only mode in which the `Query` control methods work — which makes a mid-chat model switch a **~0 ms `setModel()`** instead of a cold start.

Iteration is deliberately manual (`query.next()` in a loop, stopping at the turn's `result`) rather than `for await … break`: breaking out of a for-await calls `iterator.return()`, which closes the query and kills the process this exists to keep warm.

The design follows [zed's claude-code-acp adapter](https://github.com/zed-industries/claude-agent-acp), which solves the same problem — minus its persistent-consumer and turn-deferred machinery, which exists to interleave queued and autonomous turns. A Trilium chat is strictly one turn at a time.

Decisions worth reviewing:

- Options fixed at `query()` construction (system prompt, note tools, thinking) are **fingerprinted**; a change retires the session.
- **An aborted turn closes the session** rather than pooling it: it stops mid-stream, so the state is unknown and the SDK can wedge mid-`next()` (`interrupt()` is not guaranteed to yield — [claude-agent-acp#680](https://github.com/zed-industries/claude-agent-acp/issues/680)). The next turn pays a cold start.
- `Pushable.push` after `end()` is a **no-op, not a hang** — enqueueing onto a dead stream would otherwise wait on a promise nobody settles ([claude-agent-acp#338](https://github.com/zed-industries/claude-agent-acp/issues/338)).
- One subprocess **per chat** here (vs. one shared for Copilot), so the idle reaper and the warm-session cap bound real memory.

## Testing

359 server LLM tests pass. New unit coverage: process reuse across turns and chats, a shared spawn for concurrent turns, per-session update routing, crash recovery, config-change reseeds, cancellation, and the idle reap.

Two **opt-in** live integration tests drive the real CLIs and assert the latency drop — skipped by default since they spend subscription budget:

```
TRILIUM_COPILOT_LIVE_TEST=1 pnpm --filter server test copilot_agent.live
TRILIUM_CLAUDE_LIVE_TEST=1  pnpm --filter server test claude_agent.live
```

## Notes for the reviewer

- Targets `worktree-copilot-agent-provider`, not `main` — the Copilot provider isn't in `main` yet, so this stacks on that branch.
- Draft, since it can't land before its base does.
- The behaviour change that touches the most existing tests is the resume path: within one live process, a resume is now a bare `session/prompt` rather than `session/load`. Those assertions were updated, not deleted.

### DIFF
--- a/apps/server/src/services/llm/providers/acp_client.ts
+++ b/apps/server/src/services/llm/providers/acp_client.ts
@@ -51,6 +51,13 @@ export interface AcpClientOptions {
      * When no handler is set, requests are answered with "method not found".
      */
     onAgentRequest?: (method: string, params: unknown) => Promise<unknown> | unknown;
+    /**
+     * Called once when the subprocess dies on its own (crash, kill, agent
+     * exit) — never for a deliberate {@link AcpClient.dispose}. A pooled
+     * client uses this to evict itself so the next turn starts a fresh agent
+     * instead of handing out a corpse.
+     */
+    onExit?: (error: Error) => void;
 }
 
 export class AcpClient {
@@ -80,12 +87,12 @@ export class AcpClient {
         // left to report.
         proc.stdin.on("error", () => {});
 
-        proc.on("error", err => this.failAll(new Error(`Failed to start the ACP agent: ${err.message}`)));
+        proc.on("error", err => this.die(new Error(`Failed to start the ACP agent: ${err.message}`)));
         proc.on("exit", (code, sig) => {
             // A deliberate dispose() kills the subprocess — that exit is expected
             // and must not surface as an error for in-flight (cancelled) requests.
             if (!this.disposed) {
-                this.failAll(new Error(`The ACP agent exited unexpectedly (${sig ?? `code ${code}`}).`));
+                this.die(new Error(`The ACP agent exited unexpectedly (${sig ?? `code ${code}`}).`));
             }
         });
     }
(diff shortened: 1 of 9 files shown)


maphew (commented):
Advise gating Claude-session reuse on successful turn completion. An exception from `query.next()` can leave `held.closed` and `signal.aborted` false, so `finally` releases an indeterminate session for the next turn. Also advise scanning past busy entries when enforcing the warm-session cap; the current eviction stops at the first busy entry even when later idle sessions can be closed.

Issue #6991, opened 2025-09-15
Search: Add Search Ranking Customizability


werererer:
## Motivation
As knowledge bases grow in size, users become increasingly reliant on search to find ideas when they are relevant. Search success depends on both speed and accuracy, i.e., search ranking. Current work focuses on further increasing speed [^2]. However, while ranking issues are being addressed, the current state is still not satisfactory for consistently finding the correct note in large-scale knowledge graphs.

## Proposed solutions
Let the user configure the search ranking themselves: I think it would be optimal to let the user decide the weights themself under some settings. Reason being that there is no neutral search, and every decision for ranking in search from us is a decision onto the cognition of the user [^1]. We assume that some users want to exercise their free will onto their future cognition.

Since only "power users" will start seeking out such setting it could be text based, reducing the programming effort, while increasing the users freedom for those that do care. 
### Implementation Suggestion
Instead of hardcoding search relevance, introduce a *weighted scoring system*  

Each field could be given a weight:
E.g.:  
```
- Title: weight = 1.0  
- Labels: weight = 1.0 (or configurable)  
- Content: weight = 0.5  
```

### Proposed Algorithms to implement
Lets not reinvent the wheel and use established methods:

Either implement exactly one of the following algorithms or become algorithm agnostic:

**Preferably**: 
- BM25 [^6] [^7][^11]
BM25 is used in
Used in: Elasticsearch [^15], ArangoDB [^19]

**Alternatively**:
- TF-IDF [^8][^9][^12]
Used in: Neo4j plugins [^18], ArangoDB [^19]

**Less preferably**
Machine Learning Approaches like
- Learning-to-Rank [^4][^5][^13]
Used in: Google Search [^17], Solr [^16]
- Or Neural Embedding Models (Word2Vec [^9], InferSent [^10], BERT [^14])
*Why Less preferably*: We want to search predictability in our *own* notes, not (yet) multiuser knowledge graphs with varying output of quality like the internet itself

#### Positive Side-Effects of this approach
Implementing this solution allows simpler future implementation and experimentation of things like semantic search like Word2Vec, improving recommendation of  "similar notes" section in trilium.

## Wrapping it up
Allowing customizability for rank in search results improves scalability of trilium while ensuring users getting the information they need, when they need it.

I have an interest in implementing this, however, if I do it, it will take some time to find enough time for it, as the focus is currently on bug fixes first.

[^1]: https://www.pnas.org/doi/10.1073/pnas.1419828112
[...]

PR #10633, opened 2026-07-23
Search overhaul: matching accuracy, ranking, results UI, jump to match

### LINKED ISSUE #3098: Focus on the searched word
### Describe feature

hi.
I have to scroll the word I'm looking for in the page, is there a solution?
sample word : "Bash starts in sh compatiblity mode when"
Is it possible to immediately focus on the words in the page?
If there is more than one searched word, I think it should be the list where we focus on those words.

Animation

### Additional Information

_No response_

### LINKED ISSUE #5332: (Feature request) Improve search result highlighting
### Describe feature

This is more like a feature enhancement rather than a bug in my eyes. i guess it just wasn´t thought of when creating the highlight script.

If you search with advanced search strings like:

note.content = 'somestring' or regex

your results aren´t highlighted. That makes searching for notes way harder. Just highlighting the appropriate search result like normal keyword search would help a lot:

see:
image

vs:

image

if someone experienced is going to work on this im will help gladly. just reach out to me.

### Additional Information

_No response_

### LINKED ISSUE #5667: Better search with search excerpts?
### Discussed in https://github.com/orgs/TriliumNext/discussions/1712

<div type='discussions-op-text'>

<sup>Originally posted by **TheNomad11** April 15, 2025</sup>
TriliumNext is best! There is one issue that frustrates me and that is search! When searching for a word, we just get a list with the notes. Then we have to go through all of them and have a look. Mostly I do not find what I am looking for. 
Maybe it is already possible and I haven't found it out yet, but search excerpts like in Google Search would be extremly helpful. What do you think? </div>

### LINKED ISSUE #6225: Settings for displaying search results
### Describe feature

Currently, the behavior is as follows:
- if the number of results exceeds the page result number, the total number of results is shown alongside the page links. Page links have `...`, hiding an arbitrary number of pages.
- the number of links per page is not parametrized.

Request:
- Display the number of results even when there is only one page
- Allow users to decide the number of results per page.

### Additional Information

As for the ellipsized pages, I don't know what can be done: I faced a situation where I had a rough idea of where my target result would be, but it was hidden very far from the start in the ellipsized pages...

### LINKED ISSUE #9422: Either docs or code issues with quick search exact/fuzzy matching
### Description

What docs states about `=`:

> Quick Search now supports the exact match operator (=) at the beginning of your search query. This allows you to search for notes where the title or content exactly matches your search term, rather than just containing it.

But that's not what actually happens. `=` seems to enable exact (vs fuzzy)  matching and has nothing to do with keywords being the full text or just part of it. Indeed, I see no way of making sense of the following distinction if that were the case:

> =Project Plan - Finds notes with title exactly "Project Plan" or content exactly "Project Plan"
> ='hello world' - Use quotes for multi-word exact matches

Indeed, `='A B'` matches the exact sequence when it's contained in the title or content, while `=A B` match either exact A or exact B, again as part of the target text.

Up to this point I believe the problem is about how the feature was documented and not about how it was implemented. There is a confusion about what "exact" means: either vs fuzzy (as it's implemented) or vs partial (as it's documented).

But, moreover, fuzzy matching (that is, without `=`) only seems to affect title matching. For example, `chese` will match `cheese` in a title but not in contents. This may be an actual bug or another ill-documented featured.

### TriliumNext Version

0.102.2

macOS

### What is your setup?

Local (no sync)

### Operating System Version

macOS Tahoe

### Error logs

_No response_

### LINKED ISSUE #9426: Fuzzy search not working
### Description

```
note.title ~= Books
=> Unrecognized expression books
```

```
#title ~= Books 
=> Misplaced or incomplete expression "="
```

```
#title ~* Books 
=> Unrecognized expression books
```

### TriliumNext Version

0.102.2

macOS

### What is your setup?

Local (no sync)

### Operating System Version

macOS Tahoe

### Error logs

_No response_

### LINKED ISSUE #10616: Search Issues
I have a very large note database, and search is one of the most important features for me in Trilium. However, the more I use it, the more often I encounter situations where search does not work as expected: it takes a long time to find the required information, and sometimes it is impossible to find it at all.

Below are the problems and inconveniences that, in my opinion, require attention:

### Difficulty finding whole words

As far as I understand, an exact word search can be performed using the `=` prefix. However, if the searched word is inside brackets, it is not found. At the same time, if the same word appears nearby without brackets, the word inside brackets is highlighted as a match.

<details><summary>Video demonstration - brackets</summary>
<p>
Image
</p>
</details>

Although fuzzy search can be used as an alternative, it also has several issues described below.

### Fuzzy search result ranking

Searching for `"sync"` also finds `async`, and searching for `"program"` finds `programmer`. Notes containing `async` or `programmer` may appear higher in the results than notes containing the exact match.

Another example:

When searching for `you and me`, a note containing scattered words (`menu`, `and`, `you`) may appear higher in the results than a note containing the exact phrase.
<details><summary>Video demonstration - you and me, sync</summary>
<p>
Image
</p>
</details>
I understand that without quotation marks the search works by matching individual words. However, in my opinion, results that are closer to the original query (especially exact word or phrase matches) should have a higher ranking in the search results.

### Inconsistent fuzzy search behavior

Some examples:

- `sync` finds `Send`
- `ceck` finds `Tech`

However:

- `combinef` does not find `combined`
- `inspiratiom` does not find `inspiration`
<details><summary>Video demonstration - Fuzzy search</summary>
<p>
Image
</p>
</details>

### Search by link previews

The search does not index the titles of inline links or link previews. At the same time, it also does not find the URLs contained inside such links.

<details><summary>Video demonstration - Link</summary>
Image
</details>

### Problems with highlighting search results

- Matches found using regular expressions are not highlighted.
- Exact matches containing diacritical characters are also not highlighted

<details><summary>Video demonstration - ktorý</summary>
Image
</details>

#### Collapsible blocks

- If the found text is located inside a collapsible block, the block does not automatically expand.
- The same issue occurs when searching within a page using `Ctrl+F`: the program shows the number of matches inside collapsed blocks, but the blocks themselves remain collapsed.

<details><summary>Video demonstration - Collapsible blocks</summary>
<p>
Image
</p>
</details>

#### Code blocks

In the search results preview, code blocks are displayed without line breaks and without horizontal scrolling. If the matched text is outside the visible area, it is highlighted, but there is no way to see it.

#### Navigation to the found match
One of the most frustrating thing about using search is that if a note is large, after opening a search result we have to manually scroll through the page for a long time to find the relevant location. Additionally, when a page contains many highlighted matches, it becomes difficult to identify which fragment corresponds to the selected search result.

perfectra1n:
This overhauls search matching, ranking, the results UI, and match navigation. It started from the complaint list in #10616 and picks up the related open issues along the way.

Closes #10616, closes #9426, closes #9422, closes #5667, closes #5332, closes #6225, closes #3098. Related: #6991 (this lands a heuristic ranking upgrade rather than BM25; all weights live in one table, which is where configurable weights would plug in later if we want them).

## Matching fixes

* Exact word/phrase matching (the leading `=` prefix) is now punctuation aware. `=sync` finds a note containing `(sync)`, `sync,` or `"sync"`. Previously content was split on whitespace only, so any punctuation next to the word broke the match. Edge punctuation is stripped symmetrically from the query and the content, and underscores, `+` and inner apostrophes are kept, so `c++`, `_private` and `d'Artagnan` still work.
* Fuzzy matching now scales edit distance with token length, like Elasticsearch's AUTO fuzziness: 1 or 2 characters means exact only, 3 to 5 characters allow 1 edit, 6 and more allow 2. This stops `sync` from matching `Send` and `ceck` from matching `Tech` (both were 2 edits on a 4 character word), while `combinef` still finds `combined`.
* Fuzzy matching now also applies to note body content, via the existing progressive search phase 2. Before, fuzzy only ran against titles and attributes, which is why a typo in a word that lived in the body could not be found at all.
* `~=` and `~*` now work on properties and labels (`note.title ~= boks`, `#author ~= tolkein`). This was a lexer bug: `~` was always treated as a relation prefix, so `~=` never reached the parser even though the comparator supported it (#9426). `~*` also keeps fragment matching, so `~* progr` finds "programming".
* Reference links and link previews are now searchable. Internal reference links store stale or empty anchor text, so the target note's title is resolved at match time (injected resolver, the preprocessor stays browser safe). The `data-url`/`data-title`/`data-description`/`data-site-name` attributes of link embeds and link mentions are indexed too.

## Ranking

Content matches previously contributed nothing to the score; only title and path did. That is why a note whose title merely contained `async` could outrank a note whose body had the exact word `sync`, and why scattered words beat an exact phrase. Content matches are now classified into tiers (exact phrase, all words in proximity with an in-order bonus, exact word, prefix, substring, fuzzy) and scored accordingly. The weights are chosen so an exact title match always dominates, and that invariant is enforced by a test.

## Behavior change reviewers should look at: attribute `=` is strict again

`#capital=Vienna` used to match a label value of "Vienna Austria" because the `=` comparator drifted into word matching at some point. That contradicts the documented semantics (#9422). This branch restores strict full-value equality (case and diacritic insensitive) for attribute and property `=`/`!=`, and moves the word-match behavior to an internal operator used only by the leading `=` fulltext prefix, whose user-visible behavior is unchanged.

Who this affects: ETAPI searches, saved searches (and therefore bulk search-and-execute), the script APIs, share search, and `getNotesWithLabel`. All of them run with `fuzzyAttributeSearch: false` and silently inherited the word-match drift. An ETAPI regression test now pins the strict behavior. Quick search and autocomplete are unaffected (they run with relaxed attribute matching, which is now documented and covered by a test as well).

## Results UI

* The full search page now shows snippet cards: highlighted title, note path, a match-centered excerpt, and badges for matched attributes. Excerpts are generated server side by the same code quick search already used, and are fetched lazily per page through a new `POST /api/search-note/:noteId/result-details` endpoint (max 100 ids per request, ids are validated against the actual result set). Other view types for search notes are untouched; the cards are just the "list" view.
* The result count is always visible, and the page size is configurable through a new synced option `searchResultsPageSize` (#6225).
* Highlighting fixes: matches with diacritics are highlighted (`ktory` highlights `ktorý`), and regex searches (`note.content %= '...'`) highlight their real matches instead of nothing (#5332). Server snippet positions are computed on a guaranteed length-preserving normalization, which also let us drop the `normalize-strings` dependency.
* The jump-to-note dialog (Ctrl+L) now shows content snippets. Inline note pickers are unchanged; the snippet line is opt-in per consumer.

## Jump to match (#3098)

Opening a result from the search page, quick search, or the jump-to-note dialog scrolls to the first match and pre-fills the find bar, so Enter/F3 cycles through matches. The terms travel in a new `viewScope.searchTerms` that round-trips through link hashes, so ctrl+click into a new tab works too. Collapsed `<details>` blocks (CKEditor collapsibles) are expanded automatically when a find match or a bookmark target sits inside one; previously Ctrl+F would count those matches but could not show them.

## Docs

Search.md and Quick search.md were rewritten around concrete examples, and every example in them (32 in total, including the pre-existing ones) is validated by a new `docs_examples.spec.ts` that runs the exact query strings against the real engine. Each test names the doc section it validates. This already paid off: it caught two pre-existing doc bugs (the negation example used `note.ancestor.title`, which is not a recognized specifier, and quick search's description of `=` was simply wrong).

Note for maintainers: these two files were edited directly as Markdown rather than through `pnpm edit-docs:edit-docs`, so a later edit-docs round trip may reformat them. The examples and the spec are kept in lockstep, so a reformat should not change any example text.

## Testing

* trilium-core search area: 33 spec files, 456 tests, including new specs for the tokenizer, AUTO distances, match tiers, the preprocessor, the result-details endpoint, and the docs examples.
* ETAPI: new label-equality regression test. Routes, special_notes, share routes, benchmark and profiling suites all green.
* Client: 182 spec files, 2307 tests, including the new card view, pagination precedence, the highlighting hook (diacritics, CJK, regex, invalid regex), the searchTerms hash round-trip (commas, percent signs, unicode), and the collapsible expansion helper.
* Repo typecheck clean. No non-null assertions were introduced anywhere in the branch.


apps/client/src/services/search_jump.ts
```diff
@@ -0,0 +1,28 @@
...
+export function consumeSearchTerms(noteContext: NoteContext | undefined | null, ntxId: string | null | undefined): void {
+    const viewScope = noteContext?.viewScope;
+    const searchTerms = viewScope?.searchTerms;
+    if (!viewScope || !searchTerms?.length) {
+        return;
+    }
+
+    viewScope.searchTerms = undefined;
+    requestAnimationFrame(() => appContext.triggerCommand("findInText", { ntxId, searchTerms }));
+}
```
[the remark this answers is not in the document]
perfectra1n: Fixed in 8d6168b1ef. The rAF callback now checks that the context's viewScope is still the same object it consumed from; navigation replaces the viewScope, so a stale callback aborts. A fresh navigation from search results carries its own terms and runs its own consumption, so nothing is lost. Covered by a new test that swaps the viewScope before the frame fires.

apps/client/src/widgets/collections/Pagination.tsx
```diff
@@ -122,25 +122,38 @@ function createSegment(start: number, length: number, currentPage: number, setPa
...
     const startIdx = (page - 1) * normalizedPageSize;
     const endIdx = startIdx + normalizedPageSize;
     const pageCount = Math.ceil(noteIds.length / normalizedPageSize);
 
+    // If the page size grows (or the result set shrinks) enough that the current page no longer
+    // exists, step back onto the last valid page so the slice below can't silently yield an empty
+    // page. A no-op in steady state, so existing list/grid callers are unaffected.
+    useEffect(() => {
+        if (pageCount > 0 && page > pageCount) {
+            setPage(pageCount);
```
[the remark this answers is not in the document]
perfectra1n: Fixed in 8d6168b1ef. Worth noting the exact scenario described is close to unreachable (the out-of-range slice is always empty, so the stale request is getNotes([]) and resolves before the clamped request even starts), but the load effect had no staleness guard at all, so rapid page flips could genuinely resolve out of order. The effect now cancels superseded loads on cleanup and skips loading entirely while the clamp is settling, which also removes a one-frame blank flash. Covered by a test that resolves an older page load after a newer one.

maphew:
Independent behavioral check of this branch (1bfbf79) against main (916318b): two fresh dev servers, identical 45-note adversarial corpus seeded via ETAPI (pinned noteIds), 55 search queries run against both, results diffed by exact membership and order. Full tables and scripts are in my test log; highlights:

- 42 of 55 queries byte-identical (diacritics, CJK, Cyrillic, emoji, `C++`/`.NET`, regex/`*=*`/`=*`, orderBy/limit, fastSearch, ancestor scoping, malformed input, empty-query 400). No regressions observed.
- AUTO fuzziness fixes real false positives: on main, `cafe` matches ".NET Core Guide" (Core, 2 edits on 4 chars) and `#author ~= tolkein` matches author=Herbert; both gone on this branch.
- `note.title ~= 'plan'` and `~* 'progr'` return 0 results on main, work here - confirms the 9426 lexer fix end to end.
- Strict attribute `=` behaves exactly as described: `#capital=Vienna` vs value "Vienna Austria" matches on main, not here; `#capital!=Vienna` is now the consistent complement (empty on main, matches here). `#capital='Vienna Austria'` matches on both.
- `=sync` finds body text `(sync)` here, not on main.
- Ranking swaps on `apple`-family queries all trace to content now scoring (promoted notes carry the term/phrase in the body). One cosmetic oddity: an unclosed-quote query (`"apple`) returns the same 6 notes as `apple` but in a different order than the branch's own plain-word order.
- Saved-search note and quick-search resolve with identical membership on both, order per new ranking.

Unit specs: the 6 core search specs pass on both runtimes on both branches; this branch grows them (search 40->42, note_content_fulltext 2->7, lex 25->29, parse 16->18, search_result 13->19, match_quality 12 new).

Client-side UI (result cards, jump to match) not covered - API-level testing only.

Staenly:
Are there any chances that this will be added to the new version?

apps/client/src/services/note_autocomplete.ts
```diff
@@ -163,29 +171,43 @@ async function autocompleteSource(term: string, cb: (rows: Suggestion[]) => void
     cb(results);
 }
 
+// autocomplete.js's own "val" setter is silent by default (it goes through Typeahead.setVal,
+// which skips _checkInputValue when the widget isn't activated) — it never dispatches a native
+// "input" event. Consumers that track the live query text (e.g. NoteAutocomplete.tsx's
+// onTextChange, bound to "input") would otherwise go stale whenever the text is changed
+// programmatically rather than typed. Every function below that sets the value on the user's
+// behalf re-triggers "input" itself so that channel stays generically accurate for any caller,
+// not just one dialog.
```
eliandoran: What's the purpose of this comment?

apps/client/src/services/note_autocomplete.ts
```diff
@@ -364,6 +386,12 @@ function initNoteAutocomplete($el: JQuery<HTMLElement>, options?: Options) {
                             html += `<span class="search-result-attributes">${suggestion.highlightedAttributeSnippet}</span>`;
                         }
 
+                        // Add content snippet below, only for consumers that opted in (e.g. the
+                        // jump-to-note dialog) — link dialogs/relation editors/`@`-mentions stay compact.
```
eliandoran: Redundant comment.

apps/client/src/stylesheets/style.css
```diff
@@ -1057,6 +1057,24 @@ table.promoted-attributes-in-tooltip th {
...
+    display: block;
+    margin-top: 4px;
+    font-size: 0.85em;
+    color: var(--muted-text-color);
+    line-height: 1.3;
+    word-wrap: break-word;
+}
+
+/* Highlight the server-injected <b> tags with the same vars the snippet cards
+   (SearchResultsList.css / ListOrGridView.css) and quick-search dropdown use. */
```
eliandoran: Redundant comment.

apps/client/src/widgets/dialogs/jump_to_note.tsx
```diff
@@ -154,3 +163,20 @@ export default function JumpToNoteDialogComponent() {
...
+ * and the "input"-triggering functions in note_autocomplete.ts).
+ *
+ * The autocomplete response has no per-suggestion token list, so we pass the trimmed raw query
+ * as a single term rather than re-deriving a token list client-side — simple, and good enough
+ * for jump-to-match highlighting.
+ */
+export function deriveSearchViewScope(isCommandMode: boolean, typedText: string | undefined): ViewScope {
+    const searchString = !isCommandMode ? typedText?.trim() : "";
+    return searchString ? { searchTerms: [searchString] } : {};
+}
```
eliandoran: This should be inline and commentless.

apps/client/src/widgets/react/hooks.tsx
```diff
@@ -864,6 +865,28 @@ export function useNoteBlob(note: FNote | null | undefined, componentId?: string
...
+ * this widget's stale, about-to-unmount content) would clear the terms before the real content is
+ * ready, so we skip it.
+ */
+export function useSearchTermsConsumer(note: FNote | null | undefined, noteContext: NoteContext | undefined, ntxId: string | null | undefined) {
+    useTriliumEvent("noteSwitched", ({ noteContext: switchedContext }) => {
+        if (switchedContext.ntxId !== ntxId) return;
+        if (switchedContext.note?.noteId !== note?.noteId) return;
+        consumeSearchTerms(noteContext, ntxId);
+    });
+}
```
eliandoran, line 888: Check if this is used in multiple places, otherwise it doesn't warrant a dedicated hook.

apps/client/src/widgets/type_widgets/code/Code.tsx
```diff
@@ -115,11 +121,18 @@ export function EditableCode({ note, ntxId, noteContext, debounceUpdate, parentC
             codeEditor.setText(content ?? "");
             codeEditor.setMimeType(note.mime);
             codeEditor.clearHistory();
+
+            // Jump to the first search match (and pre-fill the find bar) when navigated from search
+            // results. Runs once the content has loaded into the editor; the same-note re-click case
+            // is covered by useSearchTermsConsumer below.
```
eliandoran: Trim.

apps/client/src/widgets/type_widgets/text/EditableText.tsx
```diff
@@ -65,14 +67,16 @@ export default function EditableText({ note, parentComponent, ntxId, noteContext
             contentRef.current = newContent;
             watchdogRef.current?.editor?.setData(newContent);
 
+            // Jump to the first search match (and pre-fill the find bar) when navigated from search
+            // results. Runs once the content has loaded into the editor; the same-note re-click case
+            // is covered by useSearchTermsConsumer below.
```
eliandoran: Trim or even better integrate as part of the documentation for `consumeSearchTerms`.

apps/client/src/widgets/find.ts
```diff
@@ -392,3 +448,13 @@ export default class FindWidget extends NoteContextAwareWidget {
...
+
+/**
+ * Splits the search terms carried in on note open into the seed (first token — a quoted phrase
+ * arrives as a single token) that drives the find counter and Enter/F3 cycling, and the remaining
+ * tokens that get a muted highlight-all pass (HTML handler only). Extracted and kept pure so the
+ * seed/extras split is unit-testable without instantiating the jQuery widget.
+ */
+export function deriveSeededFind(searchTerms: string[]): { seed: string; extraTokens: string[] } {
+    return { seed: searchTerms[0] ?? "", extraTokens: searchTerms.slice(1) };
+}
```
eliandoran: Inline and remove comments.

apps/client/src/widgets/quick_search.ts
```diff
@@ -292,7 +304,12 @@ export default class QuickSearchWidget extends BasicWidget {
             for (const result of resultsToDisplay) {
                 if (!result.notePath) continue;
 
-                const $item = $(`<a class="dropdown-item" tabindex="0" href="#${result.notePath}">`);
+                // Carry the highlighted search tokens into the opened note (jump-to-match) via the
+                // view scope. Set the href with .attr() rather than interpolating it into the HTML
+                // string so query separators aren't parsed as HTML entities. calculateHash omits
+                // searchTerms when the token list is empty, so this degrades to a plain `#notePath`.
```
eliandoran: Trim.

apps/client/src/widgets/quick_search.ts
```diff
@@ -363,6 +380,15 @@ export default class QuickSearchWidget extends BasicWidget {
...
     }
 
+    /**
+     * View scope attached to each result link so the opened note jumps to the first match and
+     * pre-fills the find bar. Undefined when the last search highlighted no tokens, so links stay
+     * plain `#notePath`.
+     */
+    private resultViewScope(): ViewScope | undefined {
+        return this.lastHighlightedTokens.length ? { searchTerms: this.lastHighlightedTokens } : undefined;
+    }
```
eliandoran: Inline.

docs/User Guide/User Guide/Basic Concepts and Features/Navigation/Search.md
eliandoran, line 1: Clean up the em dashes. If they are being used, it means the phrase itself is too complex and needs to be trimmed down to standard/engineering English.

packages/trilium-core/src/services/search/expressions/note_content_fulltext_preprocessor.ts
eliandoran, line 1: Comments need trimming down or removed.

packages/trilium-core/src/services/search/services/search.ts
eliandoran, line 1: Inline the one-liners.

packages/trilium-core/src/services/search/utils/text_utils.ts
eliandoran, line 1: Trim comments.

eliandoran (dismissed):

eliandoran:
@Staenly ,

It's merged. Check it out on the latest nightly and tell me how it goes.

Staenly:
@eliandoran thank you for merging! I have tested it and here are my notes:
- When clicking on any found note, a tooltip pops up saying: "Note <id> is not a search note".
- When clicking on a found note, it opens in the same tab. Because of this, if we need to check several notes, it becomes difficult to navigate and keep track of what has already been viewed. Plus, when we return to the search results (Alt+←), the search process starts all over again.
- If we perform a search for a new word while on the 2nd page of search results, the new search also opens on the 2nd page (it doesn't reset to page 1).
- If more than one word is entered in the search, navigating to a found note highlights/auto-searches for only the first word on the page.
- With a large number of found notes, switching the view to a table causes the application to freeze.
- Lines broken using Shift+Enter are displayed in the search results as one long word without spaces. 
<details><summary>Details</summary>
sync
sec
</details> 


#### Problems with diacritical signs:
- When searching for `ktorý`, a note with the title `ktory` appears higher in the search results than the exact match `ktorý`.
- When searching for `ktorý`, it finds the note containing `ktorý`, but when navigating to the note, auto-search/highlighting on the page does not trigger.
Let me know if anything is unclear and I should make a video demonstration

PR #10647, opened 2026-07-24
feat(core): virtual note providers; make in-app help a virtual subtree

### LINKED ISSUE #5685: Search within Help User Guide
### Describe feature

There doesn't appear to be a method to search the in-app User Guide. Even 'search in note' function doesn't work. The consequence is that one needs to resort to searching Github docs pages instead.

I expect this is a consequence of wanting to be keep the user notes and the help guide separated, which I agree is a good idea, but it is a relatively serious UX problem, especially for new users.

### Additional Information

_No response_

eliandoran:
Closes https://github.com/TriliumNext/Trilium/issues/5685.

## Summary

Introduces a generic **virtual note provider** mechanism in `trilium-core` and converts the in-app help (User Guide) into its first consumer.

Virtual notes exist only in becca (and, through it, in the client's froca) — they are **never persisted to the database, never synced, and never produce entity changes**. They are meant for content whose structure is owned by the application rather than the user; the interface deliberately anticipates future uses (computed hierarchies, read-only mounts of external data) via optional per-note content (`getContent()`) and refresh-by-reload semantics.

### The mechanism

- `services/virtual_notes.ts` — provider registry: a `VirtualNoteProvider` declares an ID namespace (e.g. `_help`), an anchor note (e.g. `_hidden`), a `getSubtree()` definition, and optional `getContent(noteId)`.
- `becca_loader` injects the declared subtrees as `isVirtual`-flagged `BNote`/`BBranch`/`BAttribute` entities on **every becca load** — so the tree API, froca, title/attribute search, note paths, backlinks, and the LLM help tools all work with zero client changes.
- Ground rules enforced in the entity layer:
  - virtual entities throw on `save()` / `setContent()` / `markAsDeleted()`;
  - persisted branches/attributes cannot attach to virtual notes (no cloning, moving, or annotating — `tree.validateParentChild` also rejects these early with a friendly message);
  - relations *pointing to* virtual notes keep resolving, with a post-injection backfill for backlinks (persisted relations load before virtual notes exist).
- Injection is defensive: a broken provider is logged and skipped without taking down the becca load; providers whose anchor is missing (first load during initial DB creation) wait for the post-`dbReady` reload; a persisted note occupying a virtual ID wins, loudly.

### Help conversion

- The ~280-note help subtree definition is removed from `hidden_subtree.ts` (along with `cleanUpHelp` and the `_help` special cases); `initInAppHelp()` now registers the `_help` virtual provider instead. Provider registration was moved before `initSql` in `initializeCore` since dbReady resolution can trigger the first becca load.
- **Migration 240** hard-deletes the previously persisted `_help` rows — notes, branches, attributes, revisions, attachments — *including their `entity_changes`*, so the cleanup does not churn through sync; every instance in a cluster runs the same migration and converges.
- **SYNC_VERSION 39 → 40**: un-migrated peers would still push `_help` entity changes, so they must upgrade before syncing.
- Locale changes now trigger a becca reload so the translated User Guide title is rebuilt.

### Benefits

- ~1,600–2,200 rows (plus their `entity_changes`) removed from every database and from sync traffic.
- Help updates shipped in a release no longer produce sync churn or startup diffing of 280 notes.
- `checkHiddenSubtree()` gets simpler; help structure can no longer drift.

### Behavioral trade-offs

- Help tree expansion state no longer persists across reloads (virtual branches can't store `isExpanded` changes).
- Existing user clones of help notes, and user attributes on them, are dropped by the migration by design — the help structure is application-owned. Relations pointing to help notes are preserved and keep resolving.
- Renaming/annotating help pages is now cleanly rejected instead of possible-but-reverted.

## Tests

- `services/virtual_notes.spec.ts` — registry unit tests (namespace validation, overlap rejection, replace semantics, content fallback).
- `becca/virtual_notes_injection.spec.ts` — integration: injection into becca, zero DB rows, read-only guards, clone/annotate rejection, provider content, relation backfill across reloads, missing-anchor and out-of-namespace providers.
- `services/in_app_help.spec.ts` rewritten for the provider registration/injection behavior.
- `apps/server/src/services/hidden_subtree.spec.ts` updated: help placement/rename-enforcement tests became virtual-immutability tests, plus an assertion that migration 240 leaves no persisted `_help` rows (the test fixture DB is at dbVersion 239 *with* persisted help rows, so every server-test fork exercises the migration for real).

> **Draft note:** the test suite could not be executed in the authoring sandbox (its egress proxy blocks the `@electron/node-gyp` codeload tarball, so `pnpm install` cannot complete there). All files pass syntax checks; relying on CI here to run the suite — will drive any failures to green before marking ready for review.

https://claude.ai/code/session_01QC2zes6Nr4TdQNtyhTZNAy

---
_Generated by [Claude Code](https://claude.ai/code/session_01QC2zes6Nr4TdQNtyhTZNAy)_


eliandoran:
@greptile-apps

maphew (commented):
No confirmed implementation defect. The main question is fit: issue 5685 asks for accessible User Guide search, while this adds a generic entity subsystem, migration, sync-version boundary, and a 794-file help conversion. Advise confirming that architecture and the intentional loss of help clones, annotations, and expansion state before resolving the large base conflict.

PR #10787, opened 2026-07-31
feat(client): a Connections tab in the right sidebar (note map, paths, backlinks, similar notes)

### LINKED ISSUE #10760: Note map as another tabPane of the new right panel
### Describe feature

Now that the attribute list gets its own tab as part of the redesigned right pane, it would make sense to put the note map as a dedicated tab as well, for direct access. 
I personally find the "split note" way of displaying note maps too cumbersome, in the current versions.

### Additional Information

_No response_

eliandoran:
Closes #10760

## Summary

Everything that says how a note is tied to the rest of the tree now lives in one place: a **Connections** tab in the right pane, holding a **note map**, **note paths**, **backlinks** and **similar notes** as four cards.

Before this, those four were scattered — similar notes and note paths were status-bar panels that slid up over the content, backlinks were a floating-button popup, and the note map was only reachable as a whole-tab view. Each was a different shape of thing, none of them survived navigating to another note, and none could be looked at side by side. The right pane already had the machinery for exactly this (tabs, collapsible cards, peek-vs-dock), so this branch moves them into it rather than growing a fifth pattern.

The status bar keeps its badges — they are good at *counting* — but a press now opens the sidebar on the relevant card instead of showing a second copy of the list. Attributes stay in a status-bar panel of their own; that one is an editor, not a listing.

## The tab

- **Note map** on top — the one card held to a height of its own, so what it leaves over is what the lists below are measured for; it reads as a heading over them rather than a block wedged underneath.
- **Note paths**, then **backlinks**, then **similar notes**. Cloning moved into the note-paths card header (it is what one *does* about a note's paths), and backlinks/similar notes got proper empty states instead of a blank card.
- Entry points aim at a card, not just at the tab: `selectRightPaneTab` now takes `expandWidgetId`, so a status-bar badge opens the pane, brings the connections tab forward, and expands the one card it is about even if the user had it collapsed. The card flashes so the press has somewhere to land.
- A badge peeks the pane rather than docking it — the status bar is no place to reflow the content from — and pressing it again puts the peek away. `reduceTabSelection` is the pure reducer for that, unit-tested.

## Making the note map readable at card size

The existing note map was drawn for a full tab. Most of the diff is getting it to work in a ~300px column:

- **Nodes are note badges**: the note's own icon inside its dot, coloured by type or by its `color` label.
- **Labels** are drawn smaller than node names, truncated by available width, shadowed so they stay legible over links, and faded out below a quarter of the zoom.
- **Arrowheads** rounded, sized and positioned properly.
- **Hover focus**: hovering a note lights up its own relations and fades the rest, with a transition in and out.
- **Single-note map**: a note with nothing linked to it, or an archived note, used to render an empty graph — both fixed (the archived case in `note_map.ts`: a subtree walk skipped the map root itself, which took its relations with it, since a link needs both ends present).
- Several rendering bugs gone: blank map after navigating to another note, wrong initial zoom then snapping, the wheel being ignored after a note switch, and the map not rebuilding when the theme changes (every colour is baked in at build time, so it has to).
- The **map-type switcher** (link/tree) moved into the card header — extracted to `MapTypeSwitcher.tsx` so the pane can offer the choice without pulling `force-graph` into the boot bundle. The card is rendered eagerly and only the graph inside it is lazy, so the card stays a direct child of the tab body and keeps the styling the pane gives its cards.
- An **expand** button opens the map in the quick-edit popup, one rung up from the card without rearranging the window.

## Per-card help

Each card carries a `?` that answers in place rather than sending the reader off to a help page over the note they were reading — these are short-answer questions ("why does this note have two paths?"). The full page stays one press away at the foot of the popup, opened as a quick-edit popup rather than a split (a split would rearrange the layout, and while the pane is peeked it would open underneath the overlay).

The pointers live in one registry, `services/in_app_help.ts` (`bySidebarSection`), keyed by widget id. The attributes pane picked up the same `?`, and the help icon was unified across the app to the circled one.

## Backend

- **`services/backlink_excerpts.ts`** — extracted out of the `note_map` route, which had it inline. It now takes HTML rather than a `BNote`, which is what let `llmChat` notes join in: their assistant prose is markdown, so it goes through the shared renderer first and wiki-link anchors are retitled from note ID to note title. Backlinks into a chat note now show a real preview instead of nothing.
- **Two long-standing bugs in similar-notes scoring** (`similarity.ts`): `words[idx] - 2` (a string minus a number → `NaN`) silently discarded every word ending in `es` instead of stemming it, and a candidate with no resolvable note path (note hoisting) aborted the whole scan with a bare `return` instead of skipping that one candidate. Both were flagged in-code as suspicious; both now have tests.
- `NoteMapNote` gained the note's icon, so the client can draw it without a second round trip.

## Tests

- `pnpm --filter client test src/widgets/note_map src/widgets/sidebar …` — 80 passing, including new specs for the map utils, the tab grouping, `reduceTabSelection`, and the expand-request lifecycle.
- `pnpm --filter server test similarity note_map backlink_excerpts` — 50 passing, including the two similarity regressions above and the archived-note map case.
- `pnpm typecheck` clean.

## Open points

- **Docs are not updated yet.** The User Guide still describes similar notes / note paths / backlinks where they used to be. The `?` popups link to the existing help pages, whose text hasn't been revised for the new location either.
- The branch is behind `main` and will want a rebase before merge.
- Worth a look from reviewers: whether the status-bar badges should peek or dock the pane, and whether attributes belong in the sidebar too rather than staying a status-bar panel.

## Manual checks I'd like a second pair of eyes on

- [ ] Note map legibility in the card across the bundled themes (light, dark, and the system-following ones — the map rebuilds on theme change now)
- [ ] Badge → sidebar round trip when the pane is already docked on another tab
- [ ] Peek dismissal: pressing a badge twice should put the peek away, not close-then-reopen


eliandoran:
@greptile-apps

contributor:
@eliandoran please add an option to keep using status bar flyouts.

For example, I only want to quick look at note paths. In status bar flyout I could get exactly that - note paths. Clean and simple.

But now, it opens very complex and information dense right side bar. 
When right side bar was already opened, it may be very disruptive - I was focused and working with TOC, but suddenly it is gone and I'm switched to another tab to see Note Paths (buried in note maps graph and other info not useful in __currernt__ flow/context)

I'm glad that note attributes at status bar survived this change. It is one of the best addition of "new" layout. And the best implementation of status bar panel that does not hide when you change a note.

eliandoran:
@contributor , fair, what about the other ones (note paths, note map, backlinks & similar notes)?

contributor:
@eliandoran maybe all of these can have two places to live:
- in sidebar (as in this PR) - useful for note exploration to remind what is that or out of curiosity
- when invoked from status bar, they can be displayed as overlay/floating "single section" of right panel (full height, without extra sections)

PR #10797, opened 2026-08-01
feat(attributes): select, copy and paste attributes in the sidebar panel

eliandoran:
Adds selection and a clipboard to the sidebar attributes panel, and reworks how the panel presents its runs of attributes on both a desktop and a phone.

## Selecting and the clipboard

Rows can be picked out — by a checkbox in the kind icon's slot, by a modifier, by a range with shift, by pressing anywhere on a row once picking has started, or by holding a row down on a phone. What can be done to them is said at the foot of the list, in a bar that holds there while the list is scrolled.

Copying writes the attributes as **the text the attributes editor already spells them out in** (`#label=value ~relation=#root/noteId #tag(inheritable)`), so a copy round-trips through that editor, through a note, or out of Trilium entirely. Pasting folds them into the note's own: a name it already carries is replaced, unless its definition says the field holds a set, in which case the pasted value joins what is there.

`getPreprocessedData` moved out of `AttributeEditor` into a new `attribute_clipboard` service, which both surfaces now share.

## The panel's runs

The four cards became one card of folding runs on a desktop — four card headers were a great deal of a hand-wide pane to spend on saying which run is which — gathered by a line drawn down from each heading's chevron, with the count worn as a badge beside the run's name. Collapsed runs are remembered under the keys the cards used, so a pane that had a section put away has it put away still.

A phone keeps the cards it always had. The tree is a pointer's affordance — hairlines, hover, a fold — and the sheet is the width of the screen, so there is no room to save by merging them; each run is a card, with its count in the header and its own way in beside it.

## Mobile fixes found along the way

- The selection bar was drawn **underneath** the rows. Every menu is lifted to `z-index: 3000` on a phone; the rows borrow a menu's dressing, and the rule that un-fixes their position leaves that behind. It does not lie idle as it would on an ordinary static box — the list is a flex item, and a flex item takes its z-index whether positioned or not.
- The bar was also invisible: it composites the card's own wash over the app surface, which lifts it off a right-pane card that paints no ground and hides it against a phone's card that paints precisely that.
- The kind icon's badge was stripped while rows were picked out — a rule clearing the slot for a checkbox a phone never puts there.
- Pasting had no way in at all on a phone, worst of all on a note carrying nothing: no row to hold down means no selection, no bar, and a list with no height to press beside. It is now in the first card's header.

## Testing

- `AttributeList.spec.tsx` — 44 tests
- `AttributeList.mobile.spec.tsx` — 5 tests, new (the device is read once at module load, so the two layouts cannot be told apart in one file)
- `attribute_clipboard.spec.ts` — 17 tests
- Full client suite: 236 files / 2909 tests passing

Behaviour was also checked against a running instance on both a desktop viewport and a phone one; four defects the happy-dom suite could not see were found that way — the stacking above, the surface composition, a CSS cascade specificity bug, and checkbox activation semantics.


maphew (commented):
The reload path prunes removed selected IDs but can leave `selectionAnchor.current` pointing to an attribute no longer present in `selectableIds`. The next Shift-click cannot resolve the anchor and silently skips range selection. Advise clearing or replacing the anchor in `pruneSelection()` and covering attribute removal followed by Shift-selection.

PR #10806, opened 2026-08-01
refactor(ckeditor5): consolidate the ckeditor5-* packages into one

eliandoran:
Folds all seven `packages/ckeditor5-*` packages into `packages/ckeditor5`, leaving one CKEditor package in the monorepo.

## Why

None of the seven were doing what a package boundary is for:

- **No external consumers.** Every one was imported only by `@triliumnext/ckeditor5`.
- **Never published.** No `publishConfig`, no publish step in CI.
- **No bundling benefit.** Everything was statically imported into `plugins.ts`; the one real lazy-load (KaTeX/MathLive) lives *inside* the math plugin and is unaffected.
- **No upstream to track.** The "forks" were hand-ported long ago and have diverged through several CKEditor API rename waves. None can be rebased on upstream.
- **Boundaries already leaked.** `admonition_toolbar.ts` and `admonition_type_dropdown.ts` lived in the aggregate while the plugin they drive lived in a package.

Meanwhile they cost real things: five carried a `ckeditor5-package-generator` scaffold — sample pages, their own ESLint/Stylelint/vitest configs, phantom devDependencies — that no root script or CI job ever invoked. Root ESLint global-ignores `packages/*`, so none of it ran.

Most importantly, the split was **hiding untested code**. `footnotes` (~1300 lines) and `admonition` (~670) had vitest configs declaring a 100% coverage gate and *no tests at all*; `math` and `mermaid` declared the same gate but CI ran them without `--coverage`, so it never fired. The aggregate explicitly excluded `**/ckeditor5-*/**` from its own report.

## What moved

Each plugin now lives in `packages/ckeditor5/src/plugins/<name>/`, renamed to the repo's snake_case convention, with tests co-located as `*.spec.ts` and licence + a provenance `README.md` beside the code:

| Package | Notes |
|---|---|
| `keyboard-marker` | GPL-3.0, combines with this repo's AGPL-3.0 — the split bought no licence isolation |
| `admonition` | Derived from CKEditor's block-quote; arrived with zero tests |
| `collapsible` | Trilium-original |
| `utils` | Trilium-internal |
| `footnotes` | ISC, via ThomasAitken from Forum Magnum; arrived with zero tests |
| `mermaid` | CKSource fork |
| `math` | ISC, isaul32 fork; brings `mathlive` + `@ckeditor/ckeditor5-icons` onto the aggregate |

Attribution is preserved: every vendored plugin keeps its original licence file, and the two files genuinely derived from CKEditor's block-quote keep the CKSource copyright plus the GPL-2.0 §2(a) notice-of-change that was previously missing.

## Bugs found on the way

Writing the missing tests turned up real defects:

- **Footnote renumbering.** `_removeFootnote()` used `` `${ index ?? 0 + i + 1 }` ``, which parses as `index ?? (0 + i + 1)`. Since `index` is non-null there, *every* trailing footnote was assigned the same number.
- **AutoMath had never worked.** It listened for `inputTransformation` on the `Clipboard` facade rather than `ClipboardPipeline`, so pasting `\[x^2\]` silently did nothing. Its test suite had been `describe.skip`ped behind a TODO wondering whether the test or the implementation was broken. Fixed, then **removed** — with `Ctrl+M`, the toolbar button and the `$$` autoformat already covering equation entry, and no `$$` support on paste, the feature was too narrow to keep.
- **A `mathVirtualKeyboard` re-declaration** in the math input view collided with MathLive's own types once co-located, and had been cascading into a dozen unrelated typecheck errors across `apps/client` and `apps/build-docs`.

Dead code found and dropped: an `editor.plugins.has("Autoformat")` check inside a plugin that `requires` Autoformat; a `listItem` schema check that could not fire since CKEditor 5 v41 models list entries as attributed blocks; unused query helpers and constants; and several impossible null checks. Guards that are genuinely unreachable but worth keeping are marked `/* v8 ignore next -- defensive: … */`, matching the convention already used elsewhere in the repo.

## Behaviour changes

Small but worth knowing:

- Footnote renumbering is now correct after deleting an emptied footnote.
- Pasting a bare `\[x^2\]` still does nothing (AutoMath removed), which matches the behaviour everyone has actually been getting.
- `blockquote.css` was trimmed to the single declaration that overrides upstream (2em padding vs 1.5em); the other nine were duplicating rules `ckeditor5.css` already ships.

## Tests

Roughly 1400 tests in the aggregate, up from ~850, with the previously untested plugins brought up to the aggregate's gate.

## Known issue before merge

⚠️ **The aggregate's 100% coverage gate does not pass yet.** Lines and functions are at 100%, statements 99.91%, branches 98.91%. The residue is branch permutations in the MathLive sync logic (`main_form_view.ts`, `math_input_view.ts`, `math_ui.ts`), which need a set of small combination tests. CI will fail on this until they're written — hence draft.

Also worth a manual look before merge: the math dialog and the footnote flows, since both are exercised here only through the browser-mode suite rather than the running app.


eliandoran:
@greptile-apps

eliandoran:
@greptile-apps

PR #10818, opened 2026-08-02
Localize the text editor through editor.t() instead of translation keys

### LINKED ISSUE #8508: Editor - Hardcoded Translations / No Weblate Options
### Describe feature

The following elements seems to be hardcoded and there are no translation options available at Weblate.

Image

Image

Image

Image

Image

Image

Image

Image

Image

Image

Image

### Additional Information

_No response_

eliandoran:
Closes #8508

## What this changes

The rich text editor used to localize through a host bridge: a plugin called `t("some.i18next.key")`, the client resolved that key against `translation.json`, and a `config.translate` entry carried the function into the editor. Two consequences fell out of that. Weblate showed translators raw key paths instead of the English sentence they were translating, and an editor built without a host — a test, a standalone instance — rendered the key itself at the user.

Plugins now use CKEditor's own translation function, passing the **English text as the message id**:

```ts
const t = editor.t;
t("Insert a table.");
t("Insert footnote %0", index);
```

The English entry lives under `text-editor.ck` in `apps/client/src/translations/en/translation.json`, keyed by the slug of that text (`insert-a-table`). The host builds a CKEditor dictionary from that section and appends it to the editor's `translations` config, so the message id resolves to the locale's string; with no dictionary at all, the id *is* correct English. Call site plus English entry is the whole procedure — there is no message list to maintain, because the English catalog is the registry.

`apps/client/src/services/i18n.spec.ts` enforces it in both directions: it scans the editor package's source for translation calls, so a string with no English entry fails, and so does an entry no plugin asks for.

## Notable pieces

- **Renaming an upstream string.** Trilium calls CKEditor's bookmarks "anchors". Those pairs live in `MESSAGE_OVERRIDES` (`packages/ckeditor5/src/messages.ts`) and are applied after the core dictionary, so the rename now holds in every locale rather than English only — which the old `window.CKEDITOR_TRANSLATIONS` override could not do.
- **Strings built before an editor exists** (the slash-command definitions) go through `translateMessage()`, same lookup minus the editor.
- **Keystrokes inside a message** come from `renderShortcut()`, which hands the editor ready-made `<kbd>` markup. Key names stay in the app-wide `keyboard_shortcut_keys` catalog that the command palette and the help dialog already share, so they are not duplicated into the editor's own messages. This replaces the `config.translate` bridge.
- **Toolbar group labels** ("Text formatting", "Insert", "Alignment") are translated where the toolbar config is built. CKEditor takes a nested dropdown's `label` verbatim — see `ToolbarView#_createNestedToolbarDropdown` — so it never reaches a translation function on its own, and those three stayed English in every locale.
- **Existing translations were re-keyed, not lost.** Strings already translated under the old key paths were migrated to their new derived keys across all locales that had them.
- Romanian wording pass on the editor strings, and the removal of a file-upload progress bar that nothing rendered.

## Testing

The client suite passes locally (241 files, 2933 tests), including the new coverage for `messages.ts`, `i18n.ts`, `shortcut.ts` and the toolbar config. The `packages/ckeditor5` browser-mode suite runs in CI.


eliandoran:
@greptile-apps

PR #10825, opened 2026-08-02
fix(build): preserve native SQLite artifacts for Docker architectures

iansherr:
## Why this PR exists

Plugin testing in isolated Docker environments revealed that a server artifact built on one host could omit the native `better-sqlite3` binaries required by another supported Docker architecture.

The existing build behavior was appropriate when artifacts were primarily used on the platform where they were produced. Cross-platform Docker testing introduced the requirement that the same server artifact remain deployable across supported Linux architectures and libc variants.

## What changed

- Preserves Linux x64 and ARM64 glibc binaries in the server artifact.
- Preserves Linux x64 and ARM64 musl binaries for Alpine images.
- Keeps the existing host-target trimming behavior for other application builds.
- Adds explicit build-artifact verification for the four Docker-native SQLite binaries.
- Removes the build-check dependency on the general server test setup.

## Validation

- `pnpm --filter=server test-build` — 4 build checks passed
- `pnpm --filter=server verify-build-artifacts` — 4 native artifacts verified
- `pnpm --filter=server docker-build-debian`
- `pnpm --filter=server docker-build-alpine`
- Debian container health check passed
- Alpine container health check passed

## Scope

This PR is limited to server artifact composition and verification. Plugin-manager behavior and developer-harness changes are submitted separately.


maphew:
Independently rebuilt this branch (head 0f03031c2f) and tested a combination not covered by CI or the PR description: arm64 + musl (Alpine), via `docker buildx build --platform linux/arm64 -f Dockerfile.alpine`.

Context: `Check Docker build` in dev.yml runs on `ubuntu-latest`, so it only covers x64 for both Dockerfile and Dockerfile.alpine. The `linux-arm64` Playwright job uses a native arm64 runner but runs the server directly with node, not through Docker. `main-docker.yml`, which does build/push the multi-arch images, only triggers on push to main, tags, or manual dispatch, never on PRs. So arm64 in a Docker container (either libc) had no pre-merge coverage before this.

Built and ran arm64 Alpine under QEMU (no native arm64 hardware here):

```
docker buildx build --platform linux/arm64 -f Dockerfile.alpine \
  -t triliumnext-alpine-arm64 --load .
```

Build succeeded on the first try. Then:

```
docker run --rm --platform linux/arm64 triliumnext-alpine-arm64 \
  node -e "require('better-sqlite3')(':memory:')"
```

loads and opens fine (`linuxmusl-arm64.node`, better-sqlite3 13.0.2). Ran the actual container (not just node -e) and let the built-in HEALTHCHECK run to completion. It reached `healthy` and `GET /api/health-check` returned 200.

Not tested: arm64 + glibc (Dockerfile/Debian) in Docker specifically. Also no real arm64 hardware here, only QEMU emulation, so timing or syscall-dependent bugs, like the musl pwritev2 issue this Dockerfile already works around, could still differ on real silicon.

Full commands and log available on request.

eliandoran:
Hi,

Thanks for the PR, but SQLite artifacts are **intentionally** trimmed. We publish separate images for each of the platform which reduces the native artifacts by 4.4 MB for a piece of code that never gets executed.

If this is used by an internal tooling on your side, you should adapt it to rebuild with the proper architecture information.

PR #10826, opened 2026-08-02
dev: make plugin development and test harnesses reproducible

iansherr:
## Why this PR exists

Plugin development required isolated Trilium instances with configurable data directories, ports, and script sources. The existing script-deployer workflow used fixed local paths and did not initialize the complete server core before deployment testing.

Those defaults were reasonable for the original single-instance development workflow. Plugin development exposed the need for a more configurable and faithful local test harness without changing production defaults.

## What changed

- Adds environment overrides for the development data directory, port, and script source directory.
- Initializes the complete server core before database and script-deployment setup.
- Keeps backend scripting explicitly enabled for this trusted local-only harness.
- Adds the missing `@stylistic/eslint-plugin` dev dependency required by the repository's existing ESLint configurations, allowing clean installs to run lint.
- Adds the minimal Split.js TypeScript configuration required by the repository typecheck.
- Gives intentionally expensive cold-import and image-processing tests explicit per-test timeouts instead of changing the global timeout.
- Handles asynchronous script-watch updates without leaving untracked promises.

## Validation

- `pnpm typecheck`
- Script-deployer tests: 22 passed
- Split.js tests: 12 passed
- Focused plugin-manager tests: 9 passed
- Scoped ESLint checks for changed client/server/script-deployer files
- Isolated script-deployer smoke test with a fresh database and port 39842
- `git diff --check`

## Manual verification

- Set `TRILIUM_DEV_DATA_DIR`, `TRILIUM_DEV_PORT`, and `TRILIUM_SCRIPT_SOURCE_DIR`.
- Start the script-deployer harness.
- Confirm a new database is created in the configured directory.
- Confirm the configured port is used and scripts are deployed.
- Stop the process and verify that the normal Trilium data directory was not touched.

## Scope

This PR is limited to development and test-harness reliability. Production plugin behavior and Docker artifact composition are submitted separately.


maphew (commented):
`TRILIUM_DEV_PORT` is converted with `Number()` and propagated without validation. Advise rejecting non-integer and out-of-range values at startup so configuration errors fail at the harness boundary instead of later in server initialization. The PR also combines harness initialization, timeout changes, and Split.js configuration; advise considering separate changes if maintainers do not need them as one unit.

eliandoran:
Thank you for the PR, but most if it has become stale as a result of us changing the tests around. The one about the `script-deployer` is real and we'll fix it in a follow up commit on main.

PR #10837, opened 2026-08-03
docs: add Zenith Hosting as a third-party cloud hosting option

odpay:
hey, i work on Zenith Hosting, a startup doing one-click hosting for open source apps. we added Trilium Notes to our marketplace, so this PR adds a small "Deploy with Zenith" badge and a three step entry under Third-party cloud hosting (links to https://zenith.hosting/host/trilium-notes).

one file, docs only, additions only. a couple of notes so nothing is a surprise:

- i put the section above PikaPods, happy to move it below if you'd rather keep them in the order they arrived.
- i only touched the markdown under `docs/`, since `apps/server/src/assets/doc_notes` is marked `exportOnly` in `edit-docs-config.yaml`. say the word if you want the in-app help re-exported too.
- glad to reword it or drop the badge image if you'd prefer the page stay plain text.

we share a cut of revenue with the projects we host, so a share of every Trilium subscription goes back to Trilium. happy to explain if you're curious: hello@zenith.hosting

### DIFF
--- a/docs/User Guide/User Guide/Installation & Setup/Server Installation/Third-party cloud hosting.md
+++ b/docs/User Guide/User Guide/Installation & Setup/Server Installation/Third-party cloud hosting.md
@@ -8,6 +8,16 @@ As an alternative to [hosting your own Trilium instance](1.%20Installing%20the%2
 
 ## Cloud instance providers
 
+### Zenith Hosting
+
+[![Deploy with Zenith](https://cdn.zenith.hosting/buttons/deploy-with-zenith.svg)](https://zenith.hosting/host/trilium-notes)
+
+1.  Go to [zenith.hosting/host/trilium-notes](https://zenith.hosting/host/trilium-notes) and sign up.
+2.  Click deploy and pick a name for your instance.
+3.  Trilium comes up at your own subdomain, ready to sync with the desktop and mobile clients.
+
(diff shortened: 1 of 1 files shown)


odpay:
good catch from the bot, fixed. dropped the badge image and left a plain text link.

rendering the page would have pulled the image from our cdn, which leaks reader ip and page-view timing to us for anyone just reading the docs. that is not a trade a docs page should make. the pikapods entry above has no badge either, so this matches the page style better anyway.

happy to reword the section or close this if you would rather not list a second provider.

eliandoran:
Hi there,

Thanks for the offer but I would say it's a bit rushed to add it now.

Trilium works based on trust and we would need to build that trust first.

- Revenue sharing was mentioned but there was no prior discussion to that or an official agreement.
- Not sure about your update policy, with PikaPods we generally know when updates land.
- We don't know your support policy, we had another third-party service that we had to remove from our lists of recommendations because they didn't respond to support tickets.

odpay:
fair enough, and closing it was the right call.

answers anyway, so they're not left hanging:

revenue share: 30% of what a trilium instance earns us, paid quarterly. if you want the actual terms, email hello@zenith.hosting.

updates: we pin, currently v0.103.0, and roll existing instances onto the new pin when we move it. Automatic updates are being developed actively.

support: goes to support@zenith.hosting, not your tracker.

we'll come back when there's a track record to point at.

PR #10841, opened 2026-08-03
Link previews, the attachment listing, and the imports that feed them (closes #10839, closes #8195)

### LINKED ISSUE #8195: Promoted attributes aren't aligned when using custom or legacy themes
### Description

I've noticed that when using custom / legacy themes, the promoted attributes are shown one per line.
I actually never complained because I thought that it was my theme that was badly designed, but this has been going on for at least the 10 last releases.
On these holidays, I tried to fix my theme, and came to the conclusion that it cannot be fixed.

Here's a screenshot when using a trilium legacy theme

Image

Here's a screenshot when using a trilium "non legacy "theme:

Image

I've disabled any #appCss notes I foundto make sure there is no interaction.
I ended loading a fresh trilium instance where I can replicate the issue.

As a side question, how do I get to create a custom theme which wuold be based on non legacy code ?

Thanks.

### TriliumNext Version

0.100.1

Windows

### What is your setup?

Local + server sync

### Operating System Version

Win 10 22H2

### Error logs

_No response_

### LINKED ISSUE #10593: Make favicons in inline links optional
### Describe feature

Favicons take up a lot of bytes inside the note's blob because the favicon itself is embedded in the note as a blob itself.

I like the inline links, but the favicon blows up my note size. Maybe the favicons could be optional, or disable-able. Inline links are still useful without the favicon.

Also maybe the favicon could be loaded/cached from the web whenever it's displayed (rather than the bytes being in the note). If there's no internet available and no cache, then fine don't show the favicon.

### Additional Information

_No response_

### LINKED ISSUE #10839: Toasts not rendering in legacy themes
### Description

In earlier versions, notifications would pop up at the top after login or page refresh, such as Bing Theme prompts and startup information (see attached screenshot).
After updating to 0.104.1, these popups disappear when using any theme other than the official built-in default theme. Backend script execution is already allowed.This instance is deployed via Docker on CentOS 7. Feedback from other users indicates this issue may not be limited to CentOS 7 host systems.
Image

### TriliumNext Version

0.104.1

Other Linux

### What is your setup?

Server access only

### Operating System Version

Centos 7

### Error logs

_No response_

eliandoran:
Closes #10593

Fifty-six commits that started with link previews and ended up touching everything a preview leans on: the pictures it stores, the list those pictures show up in, and the importers that bring them in from elsewhere.

## Link previews

A preview no longer carries its pictures inside the note's HTML. Both the cover image and the favicon are downloaded server-side and stored as attachments of the note, which is what keeps a single card from pushing a note past the 32KB `autoReadonlySizeText` threshold and flipping it read-only. One favicon is kept per site and one cover per page, so a note linking one site twenty times stores one icon.

Nothing a preview draws is ever fetched from the site it describes: an `` needs no click, so a remote URL there would tell a third party that the reader opened the note. Only an inline image or an attachment of this instance is allowed to reach an `src`, on the shared page as much as in the app, and a YouTube embed shows the stored thumbnail until the reader asks for the player.

The icon itself took several passes to get right:

- Icons are read at the size a 3x display needs, and an `.ico` offering six sizes is trimmed to the one that will be drawn — the case that left triliumnotes.org, whose icon is 114KB, with no icon at all.
- Every `rel` naming an icon is now collected and tried in turn, ending at the conventional `/favicon.ico`. A declared icon is a promise, not a delivery, and taking only the first one cost the preview its icon whenever that one 404'd.
- Among candidates, the form that weighs least wins: scalable, then compressed, then the icon file. An `.ico` holds uncompressed bitmaps, so python.org's icon is 9.7KB one way and 3.3KB the other.
- An icon drawn for one background is corrected for ours, measured from its own pixels and applied as a class so that switching theme fixes the icons without measuring anything again.
- A preview whose icon could not be had draws nothing where it would have gone. The title says more than a placeholder did.

A page that publishes no description of its own now gets one from its opening sentence, and a Twitter card's description is read where there is no OpenGraph one.

## Attachments

The listing is a grid of cards rather than a stack of blocks. Every attachment carries an icon, names its role the way a reader would, and hides the media type that repeated it. The reader's own attachments are separated from the ones the app made for itself, and the app's fold away; a scheduled erasure is a badge on the card rather than a banner across it.

## Imports

Anytype and Notion bookmark cards, and link previews arriving by ZIP or single-file import, have their pictures reconstructed as attachments rather than left inlined or dropped — the same shape a preview created in Trilium has.

## Images

ICO is handled as a picture rather than as an opaque file, so one can be dropped into a note. What counts as a picture, and what a format is called, are each decided in one place. An image URL's address is vetted before anything is downloaded from it.

## Fixes carried along

- Toasts are visible outside the Next theme again, and stacked ones are spaced (closes #10839)
- Promoted attributes lay their cells out in a row under every theme (closes #8195)
- A reference link to a non-note href shows `[missing note]` instead of nothing
- A shared web view only frames an absolute http(s) source, kept inside the `src` attribute
- `SECURITY.md` states that a shared note publishes its own content
- `CLAUDE.md` explains which backdrop-blur path a dropdown menu is on

## Notes for review

- 120 files, +5997/-897.
- Server, commons and core specs cover the new behaviour; the favicon contrast measurement is unit-tested against synthetic bitmaps and was checked against real icons (Wikipedia, GitHub, python.org, MDN) in a browser.
- Existing notes keep the preview metadata they were created with — nothing re-fetches it — so improvements to icon resolution apply to links pasted from here on.


eliandoran:
@greptile-apps

PR #10856, opened 2026-08-04
fix: stabilize Mermaid rendering and detect unlabeled diagrams (#9722)

### LINKED ISSUE #9722: Multiple inline Mermaid diagrams don't render without making a change
### Description

I just updated to the latest version and i wanted to check the latest changes in the inline mermaid diagram

When I open a note that has **more than one** inline diagram only one renders while the other doesn't, and I have to refresh the code (edit something) so it renders

Image

### Request
With the new update, the mermaid notes now support TreeView and Wardley Maps, and the note stays on the last chosen option when it opens again
I think the inline diagram should be the same, I have mentioned that in this issue #7931 

### TriliumNext Version

0.103.0

Ubuntu

### What is your setup?

Local (no sync)

### Operating System Version

Tuxedo OS 24.04

### Error logs

_No response_

wsjwu58-cmd:
## Summary
Fixes #9722 (Inline Mermaid Diagram Doesn't Render).
image
image

- Fix blank Mermaid previews (and “disappearing” content in preview mode) when a note has multiple diagrams: render generation is now per-preview-element, and `mermaid.render()` calls are serialized.
- Detect Mermaid diagram bodies in unlabeled / `language-text-x-trilium-auto` code fences so import and existing notes promote them to Mermaid widgets instead of staying as “auto-detect” source blocks.
- Minor preview CSS: allow horizontal scroll for wide diagrams; clear preview when source is emptied.

## Test plan
- [x] Open a text note with several Mermaid widgets; confirm every preview shows an SVG (not blank).
- [x] Switch a widget to Preview mode; confirm the diagram remains visible (source is only hidden via CSS).
- [x] Import markdown with unlabeled fences whose body starts with `sequenceDiagram` / `flowchart` / `graph` / etc.; confirm they become Mermaid widgets (`language-mermaid`), not auto code blocks.
- [x] Open an existing note that already stores `language-text-x-trilium-auto` Mermaid source; confirm it upcasts to Mermaid widgets on load.
- [x] Confirm ordinary unlabeled/auto code (shell, plain text) is not promoted.
- [x] Run commons unit tests for `mermaid_detect` / markdown renderer; run ckeditor5 mermaid editing/rendering specs where browser tests are available.


eliandoran:
Hi @wsjwu58-cmd ,

Thanks for the PR.

The fix for #9722 is good and I plan to integrate it.

The other part of the PR, automatically detecting code blocks that look like Mermaid is not something I would want merged in, though:

* It's not part of the fix, it's a completely separate feature so it deserves its own PR.
* From what I know, this has not been requested before.
* Heuristics can produce false positives, and I feel that the value vs risk is not high enough.
* Most sources that render Mermaid use the proper language tag. I haven't encountered any case where the mermaid is inline.

I will take the point of extracting only the fix.

eliandoran (approved):

PR #10923, opened 2026-08-07
feat(geomap): stand the buildings up once the view is leaned over

eliandoran:
The `3D` button changes the angle the map is seen from and nothing else, which leaves it a flat map drawn at a slant. This gives it something to show.

## The finding

The VersaTiles tiles the geo map already downloads **carry building heights**. This is not in the [Shortbread schema](https://shortbread-tiles.org/schema/1.0/) the styles are built against — that documents a single `dummy` property — but VersaTiles ships a [`building_heights` extension](https://docs.versatiles.org/compendium/specification_shortbread_extensions.html) which is live in production. Decoding a real z14 tile, the `buildings` layer keys are:

```
['dummy', 'height', 'min_height', 'hide_3d']
```

`hide_3d` marks footprints whose upper storeys are separate features, which must be left down or they are drawn as a solid block standing in the space those features describe. The schema was designed for this.

Coverage is total, because the tiles fall back to `building:levels × 3.66` and then to 5 m:

| | buildings | with height | median | max |
|---|---|---|---|---|
| Manhattan | 2325 | 100% | 55 m | 541 m (1 WTC) |
| Tokyo Shinjuku | 7580 | 100% | 5 m | 243 m |
| Bucharest | 2857 | 100% | 5 m | 128 m |
| rural Iowa | 42 | 100% | 5 m | 5 m |

**No new request, no new dependency, no API key.**

## What this does

One `fill-extrusion` layer, added only while the view is leaned over and removed when it is laid flat again. Standing the buildings up on a map looked straight down at buys nothing — a roof seen from above is the footprint the style already draws — and costs the doubled draw of the flat fill this hides, so a map nobody tilts pays for none of it.

The layer goes **under the style's own labels**, since a place name buried behind a tower reads as a missing label.

## Screenshots

Lower Manhattan, `versatiles-colorful`, before and after:

| today | with this |
|---|---|
| a tilted flat map | the skyline, at the heights OSM records |

Verified by rendering the real style JSON headlessly with the exact layer spec in this PR — light and dark themes, zero MapLibre errors. Bucharest and Tokyo render correctly too.

## Notes for review

- **The heights are experimental.** VersaTiles' docs reserve the right to rename or drop them without bumping the schema version, hence the `coalesce` fallbacks — the layer degrades to a townful of low blocks rather than to a flat grey plain.
- **Only the built-in vector styles.** The raster layer and a style named by URL carry no footprints; those maps are left exactly as they are.
- A style switch is handled: `keepAdditions` carries the layer over but appends it above everything in the incoming style, so it is installed again on each `style.load` and repainted for the map that arrived.
- `useMapPitch` moves from `MapToolbar.tsx` to `map.tsx`, the toolbar no longer being the only thing that needs to know which way the view is facing.

## What I deliberately left out

- **A `sky` layer.** I tried it, and it is unreachable: MapLibre's default `maxPitch` is **60**, and at 60 the horizon is off-screen at every zoom, so the sky never draws. Raising `maxPitch` makes it visible but smears the labels badly at extreme pitch. Not worth it.
- **Terrain.** Feasible keyless (VersaTiles serves a terrarium DEM at `tiles/elevation`, verified against known summits), but it interacts badly with extruded buildings — MapLibre samples one elevation per polygon centroid, so buildings on slopes float — and on MapLibre 5.24 it carries a terrain render-to-texture cache that runs at a 0.2% hit rate (fixed in 6.0) plus a `setTerrain` GPU leak (fixed in 6.1). It wants the v6 upgrade attached, so it belongs in its own PR.

### DIFF
--- a/apps/client/src/widgets/collections/geomap/Buildings.spec.tsx
+++ b/apps/client/src/widgets/collections/geomap/Buildings.spec.tsx
@@ -215,6 +215,22 @@ describe("geo map Buildings", () => {
         expect(extrusion(map)?.paint?.["fill-extrusion-color"]).not.toBe(overLightMap);
     });
 
+    it("takes them down on a style that arrives with nothing to stand up", () => {
+        // Switching to the raster map while the view is leaned over. `keepAdditions` now leaves the
+        // layer behind rather than handing it on without its source (see map.spec), but this does
+        // not lean on that: whatever a style load brings, the layer is left standing only where
+        // there is something under it to draw from.
+        const map = fakeMap();
+        renderBuildings(map);
+        act(() => map.tiltTo(45));
+        const carriedOver = extrusion(map);
+        if (!carriedOver) throw new Error("nothing to carry over");
+
+        act(() => map.loadStyle([ ...rasterStyleLayers(), carriedOver ]));
+
+        expect(extrusion(map)).toBeUndefined();
+    });
+
     it("stops listening to a map it is taken off, giving the flat buildings back", () => {
         const map = fakeMap();
         const container = renderBuildings(map);
--- a/apps/client/src/widgets/collections/geomap/Buildings.spec.tsx
+++ b/apps/client/src/widgets/collections/geomap/Buildings.spec.tsx
@@ -1,0 +1,229 @@
+/**
+ * The buildings of the map underneath, stood up while the view is leaned over.
+ *
+ * What is checked here is mostly what the layer must *not* do. It draws from a source only the
+ * built-in vector styles have, so a map on the raster layer has to be left alone rather than asked
+ * for footprints it has none of. It replaces the style's own flat drawing of the same buildings, so
+ * that one has to go down while it stands and come back when it does not. And it reads a height
+ * that is an experimental extension of the tile schema rather than part of it, so a build of the
+ * tiles without it must leave a town of low blocks rather than a flat grey plain.
+ */
+import type { LayerSpecification, Map as MapLibreGLMap } from "maplibre-gl";
+import { render } from "preact";
+import { act } from "preact/test-utils";
+import { describe, expect, it } from "vitest";
+
+import { renderInto } from "../../../test/render";
+import Buildings, { BUILDINGS_LAYER } from "./Buildings";
+import { ParentMap } from "./map";
+
+const SHORTBREAD_SOURCE = "versatiles-shortbread";
+
+/** The built-in vector styles, in the shape this component reads them: footprints, then labels. */
+function vectorStyleLayers(): LayerSpecification[] {
+    return [
+        { id: "land", type: "fill", source: SHORTBREAD_SOURCE, "source-layer": "land" },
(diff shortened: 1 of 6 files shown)


eliandoran:
Thanks — issue 1 was real, and reproducing it showed it is worse than described. Fixed in b711113697.

**The actual symptom.** I replicated the sequence against real MapLibre (load `versatiles-colorful` → tilt → install extrusion → `setStyle(rasterStyle, {transformStyle: keepAdditions(...)})`):

```
style.load #1
extrusion installed: true
--- switching to raster ---
ERROR: layers[1]: source "versatiles-shortbread" not found
setStyle returned normally
after switch — style.loads: 1
buildings-3d still present: true
shortbread source present: true
layers now: background,water-ocean,land-glacier,…,buildings-3d,poi-amenity,…
```

MapLibre validates the whole style before applying any of it, so the switch is **refused outright**. `style.load` never fires a second time and the map stays on the vector style — selecting "OpenStreetMap" while tilted appears to do nothing.

**Why the suggested fix isn't sufficient.** Removing the layer at the top of `install()` only helps if `install()` runs, and it doesn't: it is driven by `style.load`, which never fires because the switch was rejected. The effect doesn't re-run either — its deps are `[map, tilted, isDarkTheme]`, and none of them change when switching between two light styles.

**The fix.** The hole is in `keepAdditions`, which was written for `Markers` and `GpxTrack` — additions that bring their own source across with them. It never considered a layer drawing from the *style's own* source, which cannot be handed on because the source is the map that has just gone. Such a layer is now left behind with it. Same sequence after the change:

```
style.load #2
buildings-3d still present: false
shortbread source present: false
--- maplibre errors ---
[]
```

The raster map renders correctly, still tilted. I took the `install()` change too, as defence in depth rather than as the fix — it makes `install` self-consistent regardless of how it is reached. Both are covered by tests (`map.spec.ts` for the carry-over rule, `Buildings.spec.tsx` for the style load).

---

Issue 2 I'm declining, as it doesn't hold up. `Buildings.tsx`, `Buildings.spec.tsx` and `map.tsx` are all **ignored by the format config** — `eslint.format.config.mjs` doesn't match them, so `@stylistic/max-len` never applies:

```
Buildings.tsx    0:0  warning  File ignored because no matching configuration was supplied
Buildings.spec.tsx  0:0  warning  File ignored because no matching configuration was supplied
map.tsx          0:0  warning  File ignored because no matching configuration was supplied
```

`map.spec.ts` *is* covered and already carries 51 `max-len` errors on `main`. I did add four new ones there, which I've fixed. Counting stylistic errors across the whole geomap directory, with my changes and on `main`:

| rule | with changes | on main |
|---|---|---|
| max-len | 51 | 51 |
| array-bracket-spacing | 4 | 4 |
| comma-dangle | 3 | 3 |
| object-curly-spacing | 3 | 3 |

Identical. The prose lines at 101–107 characters match the surrounding files, which run to the same width throughout.

PR #10927, opened 2026-08-07
feat(share): allow slashes in shareAlias

serialuart:
Two changes are made to accomodate this:

- The /share route is adapted to allow for share aliases with slashes in them (previously if a slash was found, it would be treated as a different, nonexistent route).
- The share theme gains a new variable, "relativeBasePath", which contains the relative path to the base /share endpoint from the current note (e.g. for /share/a this will be "./", for /share/a/b/c this will be "../../../", etc.). (We keep using relative paths for assets, note downloads, etc. rather than hardcoding the /share prefix to avoid breaking non-standard setups that serve the shared notes under a custom URL (I do this with a rule in my reverse proxy, for example - my share is at notes.(domain)/, rather than notes.(domain)/share).)

This can be used for creating "subpaths" in shares, e.g. you can have a note with a shareAlias of /a, /a/b, /a/b/c, ... (or skip in-between paths entirely, or use nesting unrelated to how the notes are nested - this change allows for using slashes anywhere in the share alias).

Note that this only applies to shared notes. Exports are unaffected and behave the exact same way as they did before this change - they have their own folder structure, and all slashes in share aliases will simply be deleted from the final note HTML filenames.

---

Not sure if this isn't too much of a hack. I guess having full support for auto-generated subpaths based on node roots would achieve the exact same goal, at least for my use case (something like what exports already do, but for shares); even then, it would probably be a toggleable option (with a label like `#shareAutoPath` or something more obvious?), and would require similar base changes as what I've done here.

### DIFF
--- a/apps/server/src/share/content_renderer.ts
+++ b/apps/server/src/share/content_renderer.ts
@@ -98,6 +98,7 @@ export function renderNoteForExport(note: BNote, parentBranch: BBranch, basePath
     return renderNoteContentInternal(note, {
         subRoot,
         rootNoteId: parentBranch.noteId,
+        relativeBasePath: './',
         cssToLoad: [
             `${basePath}assets/styles.css`,
             `${basePath}assets/scripts.css`,
@@ -120,6 +121,12 @@ export function renderNoteForExport(note: BNote, parentBranch: BBranch, basePath
 export function renderNoteContent(note: SNote) {
     const subRoot = getSharedSubTreeRoot(note);
 
+    // Get the relative URL to the share root. This is required for links to
+    // other notes, assets, etc.
+    const shareAlias = note.getLabelValue("shareAlias") ?? note.noteId;
+    const shareAliasSlashCount = (shareAlias.match(/\//g)||[]).length
+    const relativeBasePath = shareAliasSlashCount > 0 ? "../".repeat(shareAliasSlashCount) : "./";
+
     const ancestors: string[] = [];
     let notePointer = note;
     while (notePointer.parents[0]?.noteId !== subRoot.note?.noteId) {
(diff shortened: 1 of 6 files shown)


eliandoran:
Hi, @serialuart .

What's your use case?

I agree that it looks a bit like a workaround. I've wondered whether we should have path-based shares instead of IDs, not sure why the original decision to use the note ID was taken.

The [static HTML export](https://docs.triliumnotes.org/user-guide/advanced-usage/sharing/static-html) kind of does this as we needed it for the website, but it's for static hosting only, not for sharing.

serialuart:
> What's your use case?

I started using Trilium as a public-facing personal wiki recently (inspired by some folks doing what's called [digital gardening](https://github.com/MaggieAppleton/digital-gardeners)); I liked the idea of having the notes be automatically updated live as I write them without extra work on my part (manual exporting and uploading to the server), so I ended up using the Sharing feature for this.

This mostly gets me there, but I was interested in having human-readable hierarchical URLs (mostly just because human-readable URLs look nice, though I also figured it would be nice for anyone who wants to link back to my site).

I guess automated exports would achieve the same result, with the extra benefit of the paths already working like I want them to :) (Admittedly I haven't looked into exports much before deciding to make this PR, in hindsight it probably would've been the more optimal option...)

> I've wondered whether we should have path-based shares instead of IDs, not sure why the original decision to use the note ID was taken.

That would be nice. I suppose it would be a better approach than this PR :)

I'll close it for now, maybe I'll look into making share paths more export-like instead (or just end up doing export automation...)

eliandoran:
@serialuart , sounds good.
I'll take note of the request, maybe we can handle it some time in the future.

PR #10924, opened 2026-08-07
Claude/trilium turkish support 3zp9xi

erkdgn:

PR #10939, opened 2026-08-08
feat(i18n): complete Turkish translation and add Turkish as a display language

erkdgn:
Completes the Turkish translation catalogues to 100% and registers Turkish as a selectable display language.

Supersedes #10924, which was closed because I accidentally deleted my fork. The commits are identical; the review history and my detailed answer to @eliandoran's question about translation quality are over there.

### What's in it

**Translations** — client `translation.json` (3378 keys) and server `server.json` (435 keys), both at 100% coverage.

**Locale registration** — following the checklist in `docs/Developer Guide/…/Adding a new locale.md`:

| File | Change |
|---|---|
| `packages/commons/src/lib/i18n.ts` | Locale entry + `"tur"` added to the `tesseractCode` union |
| `packages/commons/src/lib/dayjs.ts` | `dayjs/locale/tr` |
| `…/collections/calendar/index.tsx` | `fullcalendar/locales/tr` |
| `…/type_widgets/canvas/i18n.ts` | Excalidraw `tr-TR` |
| `packages/ckeditor5/src/i18n.ts` | `ckeditor5/translations/tr` |
| `…/type_widgets/spreadsheet/locales.ts` | `null` — Univer ships no Turkish bundle, so it falls back to English (same as `de`/`it`/`cs`) |

PDF.js needed no entry; `viewer/locale/tr` already exists.

The locale table in `packages/ckeditor5/src/i18n.spec.ts` isn't derived from `DISPLAYABLE_LOCALE_IDS` — it's pinned by hand — so the `[ "tr", "tr" ]` row is added there too. Without it the new mapping would compile but go untested.

### Quality verification

I ran an audit comparing the English source against the Turkish catalogues:

| Check | client | server |
|---|---|---|
| Coverage | 3378/3378 | 435/435 |
| Orphan keys (no English counterpart) | 0 | 0 |
| Missing keys | 0 | 0 |
| **Placeholder mismatches** (`{{var}}`, `$t()`, `<tag>`) | **0** | **0** |

The entries that stay identical to English are proper nouns (`Ctrl`, `Bing`, `LaTeX`, `MIME` and so on), i.e. intentional.

Terminology was aligned as well: "script" → "betik", "revision" → "revizyon", "attribute" → "özellik". The `script` inside `<code>` tags is left in English since it's a literal attribute name. Two strings that had drifted from their current English source (`attribute_detail.promoted_title` and `multiplicity_title`) were rewritten.

**Tests:** client 3694/3694, server 5012 passed (41 skipped), ckeditor5 1466/1466, commons 666/666. Also verified in the dev environment: Turkish appears in the language picker, and selecting it switches the interface to Turkish.

### Note

The first draft of the translations was produced with Claude; I then put them through the audit above and read a broad sample by hand. A review by another Turkish speaker would still be welcome.


erkdgn:
On the Weblate note: this was discussed in #10924 and @greptile-apps agreed there ([comment](https://github.com/TriliumNext/Trilium/pull/10924#issuecomment-5220514773)):

> Your Weblate/workflow analysis is correct — the workflow gates on `weblate:*` branches so the default branch is never force-pushed, and a merge here will be imported on Weblate's next pull as you describe.

In short: `.github/workflows/i18n.yml` only triggers on `weblate:*` branches, so Weblate never force-pushes the default branch, and merging a PR that edits a locale file is imported on Weblate's next pull rather than discarded. `.claude/skills/translating-locales/SKILL.md` in this repo spells this out as well.

eliandoran (changes requested):
Can you check the following?

1. clone collides with duplicate and copy — this is the one that actually matters. Clone is rendered three ways: kopyala (13×), çoğalt (5×), klon (1×). But çoğalt is already this file's word for duplicate, and kopyala is copy. Concretely, the tree context menu ends up with:
Copy / clone → "Kopyala / çoğalt"
Duplicate → "Çoğalt" ← different operation, same word
And sidebar_help.note_paths_cloning becomes self-contradictory:
"Bir yer eklemeye kopyalama denir: hiçbir şey kopyalanmaz…"
("Adding a place is called copying: nothing is copied…")
The English only works because clone ≠ copy. Fix: klonla/klon throughout for clone, keep çoğalt for duplicate, kopyala for copy.
2. delete_notes.delete_clones_description_other — "{{count}} kopyaları". Turkish takes no plural suffix after a numeral; should be "{{count}} kopyayı". The _one form is correct.
3. about.channel.nightly: "Nightly" → "Her gece" (= "every night"). It's a release-channel name; leave it as Nightly.
4. hidden-subtree.multi-factor-authentication-title (server.json): "İADD" — not a standard Turkish abbreviation, and the client says iki adımlı kimlik doğrulama in full. This is a settings-page title.
5. space_usage.status_content — "(of which {{attachmentsSize}} are note attachments)" → "({{attachmentsSize}} ek olarak)", which reads as "(additionally X)". Meaning lost.

eliandoran (approved):

Issue #5451, opened 2025-02-01
Request to add polylines or polygons on the geomap note type


kalmigs:
### Describe feature

I think there are many use, but for me is to map out owned lots and add notes about it.

PR #10951, opened 2026-08-08
Drawing on the geomap

eliandoran:
A proof-of-concept for drawing on the geo map, exploring [Terra Draw](https://github.com/JamesLMilner/terra-draw) as the drawing library (addresses the shape half of #5451). Four tools: **path, polygon, rectangle, circle**.

## What it does

Four drawing buttons on the geo map's edit toolbar each arm a Terra Draw session (with an instruction toast; Escape cancels). Finishing a shape creates an ordinary child **text note** whose geometry lives in a label — the `#geolocation` format, pluralized:

```
#geoShape=line:48.858093,2.294694 48.860294,2.338629 …     ← path
#geoShape=polygon:48.85,2.29 48.86,2.35 48.9,2.3           ← polygon (and rectangle)
#geoShape=circle:48.85,2.29 500                            ← circle: centre + reach in metres
```

A rectangle is just a faster hand for a polygon and persists as one; a circle stores the two numbers the drawing *meant* rather than the 64-corner ring Terra Draw approximates it with — the ring is regenerated at render time with great-circle arithmetic. Polygon rings leave the closing point unwritten (it's the first one again).

The detail pane opens on the new note with the title selected for typing over, exactly as a placed marker is offered for naming. The note's `#color` label colors the shape — a line as its stroke, an area as a translucent wash under the same stroke. The content stays free for prose: the note *is* the shape, the way a marker note is its pin and a GPX note is its track.

## Design decisions being spiked

- **Terra Draw is input-only.** One session component (`DrawShape.tsx`, parametrized by tool) owns the map only while drawing is armed; the moment a shape finishes, it becomes a note and the session's ephemeral rendering is discarded. Persisted shapes are drawn by one renderer (`ShapeLayer.tsx`) following `GpxTrack`'s source-and-layer idioms (waits on `MapStyleLoaded`, re-adds on `style.load`, guarded teardown). Nothing persisted depends on the library, and adding a tool is a mode mapping, a toolbar entry, and instruction text.
- **One note per shape, geometry in an attribute.** Shapes ride the froca attribute payload like markers do, so the map draws them with zero per-shape content fetches. The compact encoding (`shapes.ts`, pure + unit-tested) avoids quoting issues in the attribute editor and keeps the label human-readable.
- **Drawing is a placement mode** in the geo view's existing state machine, so exclusivity with marker placement, Escape handling, and the instruction toast all reuse the current machinery.
- `geoShape` is registered in commons' `builtin_attributes.ts` (typed `useNoteLabel` support) with help text in the attribute-detail popup.

Terra Draw (`1.32.3`) and its MapLibre adapter (`1.4.1`) are new dependencies of the client, bundled only into the lazy-loaded geomap chunk — the startup bundle is untouched (verified against the build output and Codecov's bundle report: ~96 kB, all in the geomap chunk).

## Deliberately out of scope for the spike

- Hiding the `#geoShape` label from the attribute editor — needs the `isAutoLink`-style exemption on **both** client rendering and the server's `updateNoteAttributes` delete loop; without the server half a hidden label would be wiped by any ribbon edit, so it's all-or-nothing and left for a follow-up.
- Hit lines / context menus for shapes, the title written along a shape, editing an existing shape (Terra Draw's select mode), and freehand modes (excluded by design — unbounded point counts don't belong in a label).
- Known cosmetic gap: between finishing a shape and froca delivering the new note, the shape can blink off briefly.

## Testing

- Specs: `shapes.spec.ts` (encoding round-trips, rejection cases, circle ring generation checked against haversine distance), `DrawShape.spec.tsx` (session lifecycle and feature→shape conversion per tool, against a mocked Terra Draw), `ShapeLayer.spec.tsx` (line/area/circle rendering, style-wait, style-switch re-add, teardown), `EditToolbar.spec.tsx` (all tool buttons, armed state, read-only refusal), `api.spec.ts` (shape-note creation).
- All 180 geomap widget tests pass; typecheck and lint clean on touched files; production client build succeeds.

https://claude.ai/code/session_013Nsn3vffwq6kUJHMHm9F8L


maphew (commented):
No technical findings. The input-only Terra Draw boundary, note-backed geometry, parser validation, style-reload handling, and focused tests form a coherent prototype. The omitted editing, hit targets, freehand mode, hidden-label handling, and brief post-finish blink are stated scope limits. Advise rebasing and rerunning the focused geomap and build checks before the product decision.

Issue #10570, opened 2026-07-19
Collapsible blocks improvements


epicnerd1990:
### Describe feature

I use the nightly releases and just updated to 104. There are so many great changes with this build! Task state checkboxes is a dream, and the addition of collapsible blocks has been an amazing upgrade to the cleanliness of my notes. As soon as I saw the blocks feature in the nightly, I implemented it many of my notes to avoid long, scrolling sections where sub-notes would have been too much. After using this feature lots and reviewing the final release implementation, I wanted to present a few things I noticed that aren't each worth their own issue request but might be worth mentioning.

1. For some reason, if you try to select a paragraph of text in a multi-paragraph section of content, the mouse wont allow you to select more then a single paragraph at a time. This interrupts the workflow when you have to select each paragraph for deletion and it feels quite unituitive. I'm using Arch linux and the desktop electron version, but this also happens on the web portal provided by a synced docker server instance too. I can take a video screenshot if others cant reproduce.

2. Is there a way to easily maintain the open state of collapsible blocks? Currently in 104, if you navigate to another note and come back, the state may stay short term within the frontend but wont save anywhere, so if you load the note in another tab or come back later, the state is reset. In the same way that the new collapsible bullets use the `data-trilium-collapsed="true"` tag on `<li>`, could the same engine check for a similar tag applied to `<details>`?
This is handy where you might have important and non-important details showing. For example, a bunch of my notes have a stack of 5 blocks, but the bottom 2 are rarely used. I'd keep those closed and the others open, and it provides a clean, matching separation of sections on the note.

3. Partially alongside the above, collapsible blocks are always closed AND collapsible bullets are always open when the note is printed or exported as PDF. Bullets being open make sense, as closing them in the UI is probably more for reducing visual clutter. I'd like to review the idea that collapsible blocks being open as well, or perhaps an attribute added like `PrintListVisibility` or `PrintAllSections` that a user can set to ensure all blocks are open or closed?

Thanks for your time!

eliandoran:
Interestingly enough, (1) is actually a bug in Chromium.

adoriandoran:
[...]

Issue #10284, opened 2026-06-25
Spoiler Block Support in Text Notes


rami-shalhoub:
### Describe feature

As Trilium text notes heavily uses HTML, it should be easy to implement a spoiler block 
it has the following HTML syntax
```
<details>
<summary>Spoiler Alert</summary>
This text is hidden until clicked.
</details> 
```   
and it looks like this:
<details>
<summary>Spoiler Alert</summary>
This text is hidden until clicked.
</details>   

### Temporarily Work Around  
the only way I was able to implement this was by importing a markdown file that already has the spoiler tags, or by using the "import markdown from clipboard`

eliandoran:
Upcoming version will already support it.

Issue #8229, opened 2025-12-31
[Feature Request] Bulleted lists support collapsing and expanding


studyduck:
### Describe feature

If bulleted lists supported collapsing and expanding like Roam Research and Logseq, this would be highly appealing to users who prefer outline-based note-taking software.

drisch314:
@studyduck I'm currently developing a widget that enables this! (as well as collapsible section headings)

I'm still working on it and taking user feedback, but I think the collapsing bullet points are pretty stable. Here's a link if you want to try it out: https://github.com/drisch314/trilium-collapsible

And if you want to give me some notes, feel free to leave a comment in the [discussion](https://github.com/orgs/TriliumNext/discussions/8201).

Issue #4969, opened 2024-05-09
Note editors other than CKEditor


meichthys:
zadam/trilium#4405 (VIM)
zadam/trilium#2259 (VIM)
Discussion, with some alternatives presented: [TriliumNext/discussions/24](https://github.com/orgs/TriliumNext/discussions/24)

eliandoran:
@Alumniminium , is the bounty still up? :)

CyberFoxar:
@eliandoran,Even is Alumniminium is not up, I'm willing to contribute another $100 for that ! I'd really love another, better editor. Never done such payed bounty thing, but we'll figure it out x)

eliandoran:
@CyberFoxar , CKEditor is going to stick since we feel it's mature enough and a pretty good WYSIWYG editor.
However that doesn't mean we're not open to a different paradigm for people who prefer Markdown-based editors.

rom1dep:
Linking previous discussions on the subject: https://github.com/TriliumNext/Notes/discussions/24

CyberFoxar:
> [@CyberFoxar](https://github.com/CyberFoxar) , CKEditor is going to stick since we feel it's mature enough and a pretty good WYSIWYG editor. However that doesn't mean we're not open to a different paradigm for people who prefer Markdown-based editors.

Agree ! CKEditor is a very good editor, but I think we're seeing its limitations, outlined in the discussion linked above. I won't rehash the arguments for it, but I think having an alternative editor (that's not CodeMirror) for markdown and/or a way to mod CKEditor to get markdown in/out quicker (and without having to import/export) would be a good think to explore. 

Hence why I'm willing to put forth a bounty !

Alumniminium:
> [@Alumniminium](https://github.com/Alumniminium) , is the bounty still up? :)

bet

eliandoran:
@CyberFoxar , @Alumniminium ,

If you'd like to add a bounty on it, here it is: https://oss.issuehunt.io/r/TriliumNext/Trilium/issues/4969

Irvinwop:
@Alumniminium @CyberFoxar Are the independent $200 and $100 pledges still available following the IssueHunt refunds? If so, what would count as acceptance, and would the amount be shared with any existing contributor?

@eliandoran Is a Markdown source editor with live preview, available alongside CKEditor, still a direction you would consider? Before implementation, I would seek agreement on the design, delivery phases and use of AI assistance as required by the contribution guide.

No implementation has been started.

[...]

PR #10977, opened 2026-08-10
Spike: stable block IDs for per-block deep links

eliandoran:
Draft spike, not for merge — opened to make the shape reviewable. Comes out of an analysis of six years of "block-based system" requests.

> **Update:** the first push had a hang bug (infinite post-fixer loop), found by running it. Fixed in `e771f5b`; details in [Status](#status--please-read-before-reviewing) below.

## Why: the demand, and what it actually is

There is no single block-system request. There's a cluster of ~17 issues from **~26 distinct people** (11 issue authors, 15 more backing them in comments) plus ~36 👍, splitting into four asks:

| Ask | Issues |
|---|---|
| Block references / transclusion (Roam, Obsidian `[[ ]]`, Logseq) | #1233, #2726, #4533 |
| Deep links to a paragraph/block (OneNote-style) | #1021 (19 👍), #9336 |
| Blocks as manipulable objects (reorder, collapse, annotate) | #319, #940, #3112, #8229, #10570, #10284 |
| Replace/augment the editor to get blocks | #4969 (9 👍), TriliumNext/Notes#24 |

Counter-signal worth stating: discussion #24 polled "Should TriliumNext aim to replace CKEditor?" — ~97 votes, roughly 7% in favour. The demand is for block *capabilities*, not a new editor.

Much of cluster 3 has already shipped (`move_block_updown`, `collapsible`, anchors via the relabelled `Bookmark` plugin + `copy_anchor_link`). What's missing is the data-model piece: blocks have no stable identity, so there are no block-level backlinks, no transclusion, no search-result-to-block positioning.

## What this spike does

Adds an optional `blockId` model attribute on `$block`, serialized as `data-block-id`, plus a `copyBlockLink` command that mints one on demand and copies `#root/<noteId>?blockId=<id>`.

Three decisions, each of which removes a problem rather than solving it:

**`data-block-id`, not `id`.** `sanitizer.ts` allows `id` only on `a`, `h2`, `li` — a `<p id>` is stripped on save. It allows `data-*` on every element, so this round-trips with **no sanitizer change**, and stays out of the document-global `id` namespace.

**Assigned lazily.** An ID is minted only when someone copies a link to that block. This is what keeps exported HTML clean, and it's the asymmetry that makes this viable where CKEditor's list item IDs weren't: list IDs are *structural* (every item needs one, hence the export clutter that got them turned off here), whereas a block ID is *referential* — nothing breaks if a block hasn't got one.

**Skipped on the clipboard pipeline**, exactly as list item IDs already are (see the comment in `cuttonote.ts`). Copied HTML carries no IDs, so a paste can never duplicate an ID something links to. That leaves block *split* as the only duplication vector, handled by a post-fixer where **document order decides**: the first occurrence keeps the ID, so splitting a linked paragraph leaves existing links pointing at the top half.

Deliberately **not** routed through `GeneralHtmlSupport` — `htmlSupport.allow` is driven by the user-configurable `allowedHtmlTags` option, so a restrictive setting would silently drop IDs on save.

## Files

- `packages/ckeditor5/src/plugins/block_id.ts` — schema, conversion, post-fixer, command, button
- `packages/ckeditor5/src/plugins/block_id.spec.ts` — conversion round-trip, clipboard-pipeline omission, split/duplicate handling, lazy minting
- `link.ts` — `blockId` view scope + a shared `takeScrollTargetSelector()` now used by both text views (this also drops an existing non-null assertion in `EditableText.tsx`)
- `plugins.ts`, `toolbar.ts`, `en/translation.json` — registration, toolbar placement, label

## Status — please read before reviewing

**Nothing here has been executed.** `pnpm install` fails on an egress-policy `403` for `codeload.github.com` (the `electron/node-gyp` tarball pinned in the lockfile), which leaves no pnpm virtual store. The Chrome-for-Testing hosts webdriverio would fetch a browser from are `403` as well, and the local Chromium/chromedriver pair is version-mismatched (141 vs 147) — so this environment cannot run this package's browser suite by any route. Symbols and converter shape were verified against existing in-repo usage (`table_indent.ts` is the closest analogue), which catches wrong imports but not logic errors.

That gap already cost one bug, and it is the main thing to weigh when reviewing:

> **`e771f5b` — infinite loop in the post-fixer.** The default tree walker visits each element *twice*, as `elementStart` and again as `elementEnd`, handing back the same element both times. The dedupe walk therefore saw every ID collide with itself, reissued it, and returned `true` — which re-runs the post-fixer over a document that still looks equally broken. The editor hung as soon as any block carried an ID, which copying a block link is exactly what creates. Fixed by walking with `ignoreElementEnd` and applying reissues after the walk rather than mutating mid-iteration. The existing specs would have caught it had they been runnable; a regression test now pins it explicitly.

**Please treat the remaining tests as unvalidated** and run the suite locally before trusting any of this.

## Open decisions

1. **Revision side effect.** Lazy minting means "copy link" edits note content — a gesture users read as read-only can produce a revision. The alternative is stamping every block up front, which reintroduces the export clutter.
2. **Toolbar placement.** `copyBlockLink` was added to *both* the classic toolbar and the block toolbar, because `toolbar.spec.ts` asserts the two carry identical item sets and there is no block-only escape hatch (only `FIXED_ONLY_ITEMS`). The block toolbar is the natural home; the classic-toolbar entry exists to satisfy that invariant and may not be what you want.
3. **Markdown export drops block IDs** (`export/markdown.ts` keeps only `kbd`/`sup`/`sub`). HTML-only concept — fine, but should be a stated limitation.
4. **Post-fixer cost** — gated on insert changes, but still a full walk of the roots when one fires. Fine for a spike; worth profiling on large notes.

## What this unlocks (not in this PR)

Block IDs are only worth what's built on them: transclusion (#1233, #2726 — `backlink_excerpts.ts` already parses note HTML and emits `ck-content` fragments, so extracting one block is the same shape of function), block-level backlinks, and block-precise search results (#319). Not block *cloning* (#4533) — that needs write-back into a foreign note's content blob and collides with note-granular `EntityChange` sync.

### DIFF
--- a/apps/client/src/services/link.ts
+++ b/apps/client/src/services/link.ts
@@ -70,6 +70,8 @@ export interface ViewScope {
     tocCollapsedHeadings?:  Set<string>;
     /** When set, scrolls to a bookmark anchor within the note after navigation. */
     bookmark?: string;
+    /** When set, scrolls to the block carrying this `data-block-id` after navigation. */
+    blockId?: string;
 }
 
 /**
@@ -92,7 +94,32 @@ const NOTE_PATH_PATTERN = /^[_a-z0-9]{4,}(\/[_a-z0-9]{4,})*$/i;
 const MAX_SPLIT_PANES_IN_HASH = 8;
 
 /** Hash parameters that belong to a pane's view scope rather than to the window as a whole. */
-const VIEW_SCOPE_PARAMS = ["viewMode", "attachmentId", "bookmark"];
+const VIEW_SCOPE_PARAMS = ["viewMode", "attachmentId", "bookmark", "blockId"];
+
+/**
+ * The CSS selector for the in-note element a view scope wants scrolled into view — a named anchor
+ * (`?bookmark=`) or a block (`?blockId=`) — or `null` when it asks for neither.
+ *
+ * Consuming: the target is cleared as it is read, so a later re-render of the same note does not
+ * yank the reader back to it. Callers should therefore only call this once they are ready to
+ * scroll. Shared by the read-only and editable text views, which locate the element differently
+ * (rendered content vs. the editor's DOM root) but agree on what to look for.
+ */
+export function takeScrollTargetSelector(viewScope: ViewScope | undefined): string | null {
+    if (viewScope?.bookmark) {
+        const selector = `[id="${CSS.escape(viewScope.bookmark)}"]`;
+        viewScope.bookmark = undefined;
+        return selector;
+    }
+
+    if (viewScope?.blockId) {
+        const selector = `[data-block-id="${CSS.escape(viewScope.blockId)}"]`;
+        viewScope.blockId = undefined;
+        return selector;
+    }
+
+    return null;
+}
 
 interface CreateLinkOptions {
     title?: string;
--- a/apps/client/src/translations/en/translation.json
+++ b/apps/client/src/translations/en/translation.json
@@ -3957,6 +3957,7 @@
       "copy-anchor-reference-link": "Copy anchor reference link",
       "copy-formatting": "Copy formatting",
       "copy-image-to-clipboard": "Copy image to clipboard",
+      "copy-link-to-this-block": "Copy link to this block",
       "copy-to-clipboard": "Copy to clipboard",
       "copy-url": "Copy URL",
       "create-a-bulleted-list": "Create a bulleted list",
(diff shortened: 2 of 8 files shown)


maphew (commented):
The block-ID model is useful, but navigation can lose its target. `takeScrollTargetSelector()` clears the target before EditableText knows that CKEditor has a DOM root, so content arriving before editor initialization can consume the link without scrolling. Advise clearing only after a target is found or retrying after editor readiness.

ReadOnlyText also reruns the scroll effect only when `blob` changes, so navigating to another block in the same open note does not retrigger it. Advise making the view-scope target part of the effect trigger. The revision side effect, toolbar placement, Markdown omission, and profiling need are already declared spike decisions.

PR #10998, opened 2026-08-12
docs: add Hostim to third-party cloud hosting

pavel1337:
Adds Hostim as a second entry under *Cloud instance providers*.

I maintain [Hostim](https://hostim.dev), a small EU container hosting platform, and we added a one-click Trilium template — so this is me adding our own entry. Happy to drop it or reword it however you prefer.

Same shape as the PikaPods entry above it: the official `triliumnext/trilium` image, `TRILIUM_DATA_DIR` on a persistent volume, HTTPS terminated in front, no forks or patches. Nothing to add to the disclaimer box either — no free credits, no revenue share — so I left it untouched. The intro line already says "two services", which now matches the list again.

Both the User Guide markdown and the paired in-app help HTML are updated, per `docs/Developer Guide/Developer Guide/Documentation.md`. No files added or moved and no images, so the meta file is untouched.


eliandoran:
Hi,

Thanks for the offer, but for now we are not looking to offer visibility to other third-party providers. In addition, this seems to be a more generic Docker solution, rather than a dedicated Trilium provider.

PR #11042, opened 2026-08-16
fix(etapi): support base64-encoded binary content when creating attachments

cozyboy:
## Problem

`POST /etapi/attachments` stores the `content` field verbatim as a string. Since JSON cannot carry raw binary data, binary payloads sent base64-encoded (the natural choice for API clients) were silently stored as the base64 *text*, producing corrupted attachments — e.g. images that fail to decode when served back with their `image/*` MIME type.

The `encoding` convention already exists elsewhere in the codebase:
- `AttachmentRow.encoding?: "base64"` documents "will be decoded from base64 to binary before storage"
- the ZIP importer decodes `encoding: "base64"` (see `packages/trilium-core/src/services/import/zip.ts`)
- the web API note-update path decodes it too (see `updateNoteData` in `packages/trilium-core/src/services/notes.ts`)

But the ETAPI attachment route rejected the property outright (`PROPERTY_NOT_ALLOWED`), so clients could not use the convention at all.

## Fix

- Accept an optional `encoding` property (`"base64"` only) on `POST /etapi/attachments`
- Decode `content` from base64 to bytes before storage (via `binary_utils.decodeBase64`)
- Reject unsupported encodings with `400 INVALID_ENCODING`
- Document `content` / `encoding` in `etapi.openapi.yaml`
- Add spec tests: base64 binary round-trip and unsupported-encoding rejection

## Testing

`pnpm vitest run --config apps/server/vite.config.mts spec/etapi/attachment-content.spec.ts` → 5/5 passed (2 existing + 2 new + protected-attachment cases).

## Usage

```bash
curl -X POST "$SERVER/etapi/attachments" -H "Authorization: $TOKEN" -H "Content-Type: application/json" -d '{
  "ownerId": "NOTE_ID",
  "role": "image",
  "mime": "image/png",
  "title": "screenshot.png",
  "content": "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg==",
  "encoding": "base64"
}'
```

### DIFF
--- a/apps/server/spec/etapi/attachment-content.spec.ts
+++ b/apps/server/spec/etapi/attachment-content.spec.ts
@@ -53,6 +53,46 @@ describe("etapi/attachment-content", () => {
         expect(response.text).toStrictEqual(text);
     });
 
+    it("creates an attachment with base64-encoded binary content", async () => {
+        const binary = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 0x00, 0x01, 0x02, 0x03]);
+
+        const createResponse = await supertest(app)
+            .post(`/etapi/attachments`)
+            .auth(USER, token, { "type": "basic"})
+            .send({
+                "ownerId": createdNoteId,
+                "role": "image",
+                "mime": "image/png",
+                "title": "binary attachment",
+                "content": binary.toString("base64"),
+                "encoding": "base64"
+            })
+            .expect(201);
+        const attachmentId = createResponse.body.attachmentId;
+        expect(attachmentId).toBeTruthy();
+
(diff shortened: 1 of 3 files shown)


cozyboy:
Thanks for the review! Both suggestions are now addressed in the updated commit:

- **P1**: `content` is validated against the RFC 4648 base64 grammar before decoding; malformed input returns `400 INVALID_BASE64_CONTENT` instead of silently persisting truncated bytes. Added a spec test covering this.
- **P2**: replaced the postfix non-null assertion with the already-narrowed `params.ownerId`.

Updated spec suite (`attachment-content.spec.ts`) passes 6/6 locally.

PR #11073, opened 2026-08-18
CardDAV address book server as a backend script (prototype)

eliandoran:
## What this is

A NextCloud-style address book that CardDAV clients (DAVx⁵, iOS/macOS Contacts, Thunderbird) can sync against, implemented **entirely as a Trilium backend script** (`apps/script-deployer/scripts/carddav.ts`, a `#customRequestHandler` note) — deliberately, to probe the limits of the scripting system. The gaps that surfaced are recorded in the developer guide: *Interoperability / "CardDAV prototype and scripting gaps"*.

## Data model — one note per contact

- An address book is any note with `#carddavAddressBook`. On first use the script provisions a "Contacts" `book` collection (table view) with inheritable promoted attribute definitions.
- A contact is a child note: `FN` ↔ note title, `NOTE` ↔ note content, and `EMAIL`/`TEL`/`ORG`/`TITLE`/`BDAY`/`URL`/`ADR`/`CATEGORIES`/`N`/`UID` map to labels — so `#email *=* "@acme.com"` search works and promoted email/phone fields render mailto/tel buttons.
- Unmapped vCard properties round-trip via `#vcardExtra` (except `PHOTO`).

## Protocol subset

`OPTIONS`, `PROPFIND` (depth 0/1), `REPORT` (`addressbook-query`, `addressbook-multiget`), `GET`, `PUT` with `If-Match`/`If-None-Match`, `DELETE`; principal/home-set discovery, per-contact ETags, collection CTag; HTTP Basic auth against a `#carddavPassword` label on the script note. No `sync-collection` REPORT (clients fall back to CTag polling) and no `/.well-known/carddav` (scripts cannot register routes outside `/custom`), so clients need the explicit base URL `http://<host>/custom/carddav/`.

## Enablers fixed along the way

- **`apps/server/src/app.ts`**: `express.text()` now also accepts `text/*`, `application/xml`, `application/*+xml` — bodies of those types never reached (synchronous) custom handlers; `req.body` stayed `undefined`.
- **`packages/trilium-core/src/services/script_context.ts`**: allow `require("zlib")` — the existing sample scripts already import it and failed at runtime under the module allowlist.
- **script-deployer**: deploy `#customRequestHandler` from front matter; enable backend scripting in the dev instance; fix the fresh-database startup path (`cls` was used before `initializeCore()` ran — the server now starts first, like the setup wizard).
- `scripts/probe.ts`: minimal custom-handler echo script used to verify method routing and body parsing.

## Testing

- Verified end-to-end against the script-deployer dev instance with curl driving the request shapes DAVx⁵ uses: discovery chain, address-book auto-provisioning, contact create/read/update/delete (`201`/`204`/`404`/`412`), ETag change on update, stale `If-Match` rejection, multiget with a missing href, unknown-property round-trip, and the Trilium-side note carrying the expected title/labels/promoted definitions.
- `pnpm typecheck` clean; script-deployer (22), custom-route (6), script-context (server + standalone) and all 153 ETAPI specs pass; changed files pass the format config.

## Draft status / open questions

- Real-device DAVx⁵/iOS testing not yet done (curl-level flows only).
- A phone-side edit of a contact's note text replaces rich Trilium content with plain paragraphs (`NOTE` is plain text) — accept, or treat note content as Trilium-only?
- The plaintext `#carddavPassword` label is a stopgap; the gaps report proposes `api.crypto` / ETAPI-token validation for scripts as the proper fix, plus further harness improvements (route-table dispatch for custom handlers, script bundling/testability in script-deployer, opt-in async handlers).

https://claude.ai/code/session_019KM3Lo92WTx84W1S6pqSTb

---
_Generated by [Claude Code](https://claude.ai/code/session_019KM3Lo92WTx84W1S6pqSTb)_

### DIFF
--- a/apps/script-deployer/scripts/carddav.ts
+++ b/apps/script-deployer/scripts/carddav.ts
@@ -1,0 +1,989 @@
+/**
+ * @trilium-script
+ *
+ * id: carddav
+ * type: backend
+ * title: CardDAV Address Book Server
+ * customRequestHandler: carddav(/.*)?
+ */
+
+/*
+ * A CardDAV server implemented entirely as a Trilium backend script, exposing address book
+ * notes to CardDAV clients (DAVx⁵, iOS/macOS Contacts, Thunderbird) at /custom/carddav/.
+ *
+ * Data model — "one note per contact":
+ * - An address book is any note carrying #carddavAddressBook. One named "Contacts" is created
+ *   under the root on first use, as a `book` collection with table view and inheritable
+ *   promoted attribute definitions, so contacts edit nicely in the Trilium UI too.
+ * - A contact is a child note of an address book. vCard properties map to labels (#email,
+ *   #phone, #firstName, ...), FN maps to the note title and NOTE to the note content, so
+ *   contacts are searchable (`#email *=* "@acme.com"`) and cloneable like any other note.
+ *
+ * Authentication: HTTP Basic. The password is the value of #carddavPassword on this script
+ * note; the username is ignored. Requests are refused until that label is set.
+ *
+ * Protocol subset: OPTIONS, PROPFIND (depth 0/1), REPORT (addressbook-query and
+ * addressbook-multiget), GET, PUT, DELETE — the set DAVx⁵ and Apple clients need for
+ * two-way sync. No sync-collection REPORT (clients fall back to CTag polling) and no
+ * /.well-known/carddav (scripts cannot register routes outside /custom, so the client must
+ * be configured with the base URL directly).
+ */
+
+const BASE_PATH = "/custom/carddav";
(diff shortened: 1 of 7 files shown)


maphew (commented):
Two replacement and concurrency paths need tightening beyond the declared prototype limits. DELETE ignores `If-Match`, so a stale client can remove a newer contact. Advise applying the same ETag precondition used by PUT and returning 412 on mismatch. A replacement PUT without NOTE also leaves the old note content in place and emits it again later. Advise clearing content when NOTE is absent.

The added guide path also fails the documentation build because its parent cannot be resolved. The lack of device testing, explicit custom URL, missing sync-collection support, and plaintext-password stopgap are already stated prototype constraints.

PR #11101, opened 2026-08-20
fix(board): stop a deleted column resurrecting through the persisted mirror

### LINKED ISSUE #11100: Issue After Update to Version 0.105.0 – Unable to Delete Kanban Board Value
### Description

Hi Team,

After updating to version 0.105.0, I'm tried create new column and test to delete values from the Kanban Board but failed. I’ve tried multiple times but the issue persists.

1. TCD
2. TCD Review

Could you please help check what might be causing this and advise on the fix?

Image

Thank you.

### TriliumNext Version

0.105.0

Ubuntu

### What is your setup?

Server access only

### Operating System Version

Ubuntu

### Error logs

_No response_

yzxcj797:
Fixes #11100.

## What happens on 0.105.0

On a server-hosted instance, deleting a kanban column (right-click → *Delete column*, confirmed) silently fails — the column comes back empty every time. Local desktop usage mostly wins the race, which is why this surfaced as server-access-only.

## The race

`removeColumn` deletes the notes' status labels, then `storeColumns` saves the view config (column dropped) and fires the definition write that drops it from `selectOptions`. The config save re-renders the board, and the refresh it triggers resolves columns from the `statusDefinition` memo — which still holds the pre-write options, because `attributes.setLabel` only awaits the HTTP round-trip while the memo is invalidated by the websocket `entitiesReloaded` that lands later. On a server connection that window reliably loses.

The stale options resurrect the column in `resolveBoardColumns` (the definition leads), `getBoardData` then writes it back into `board.json`, and the render-time definition sync sees an already-up-to-date definition. Now the column lives in the persisted mirror with neither the definition nor any note naming it — and a persisted column never leaves on its own, so the delete is permanently undone.

## The fix

A board that **owns** its definition is the one case where the persisted list is only a mirror of that same resolution, written by the board itself. For those boards, `resolveBoardColumns` now filters the persisted columns against the definition options and the discovered note values, so the delete sticks as soon as either settles, and the mirror self-heals on the next refresh.

Boards without a definition of their own keep the old behaviour exactly: their persisted list is the arrangement itself (pre-definition columns, values held by off-board notes), not a mirror of anything, and `isOwned === false` (inherited/template definitions) also keeps it.

## Verification

- `vitest run src/widgets/collections/board/` — 47/47 (6 files), including new cases: a mirror drops a column nothing else names, keeps one still named by the definition or the notes, non-mirror boards keep arranged columns, and `getBoardData` heals a resurrected `board.json` in one pass.
- Repo typecheck is clean for these files (the local `apps/web-clipper` project lacks the generated `.wxt` assets here; those errors reproduce on an unmodified tree and are environment-only).

### DIFF
--- a/apps/client/src/widgets/collections/board/api.spec.ts
+++ b/apps/client/src/widgets/collections/board/api.spec.ts
@@ -8,7 +8,12 @@ import server from "../../../services/server";
 import { buildNote } from "../../../test/easy-froca";
 import { BoardViewData } from ".";
 import BoardApi from "./api";
-import { BOARD_TEMPLATE_ID, getStatusDefinition } from "./columns";
+import {
+    BOARD_TEMPLATE_ID,
+    getStatusDefinition,
+    isRecentlyRemovedColumn,
+    unnoteColumnRemoved
+} from "./columns";
 
 vi.mock("../../../services/bulk_action", () => ({
     executeBulkActions: vi.fn(async () => {})
@@ -38,7 +43,7 @@ function createApi(
         () => {},
         getStatusDefinition(board, statusAttribute)
     );
-    return { api, saved };
+    return { api, saved, board };
 }
 
 describe("BoardApi column mutations", () => {
@@ -65,6 +70,35 @@ describe("BoardApi column mutations", () => {
             [ "To Do", "Shipped", "In Progress" ],
             [ "Shipped", "In Progress" ]
         ]);
+    });
+
+    it("clears the removal mark once the definition write settles, keeps it on failure", async () => {
+        const viewConfig: BoardViewData = { columns: [ { value: "Kept" }, { value: "Gone" } ] };
+        const { api, board } = createApi(viewConfig, [ "Kept", "Gone" ]);
+        let settle: ((ok: boolean) => void) | null = null;
+        vi.spyOn(api as never as { syncColumnsToDefinition: () => Promise<void> }, "syncColumnsToDefinition")
+            .mockImplementation(() => new Promise<void>((resolve, reject) => {
+                settle = (ok: boolean) => (ok ? resolve() : reject(new Error("write failed")));
+            }));
+
+        await api.removeColumn("Gone");
+        expect(isRecentlyRemovedColumn(board.noteId, "Gone")).toBe(true);
+
+        // The definition write failing keeps the marker: the delete has not
+        // stuck, and the marker still wins the mirror ambiguity until its TTL.
+        settle?.(false);
+        await Promise.resolve();
+        await Promise.resolve();
+        expect(isRecentlyRemovedColumn(board.noteId, "Gone")).toBe(true);
+
+        await api.removeColumn("Gone");
+        settle?.(true);
+        await Promise.resolve();
+        await Promise.resolve();
+        // Settled successfully: any later appearance of the value is a
+        // recreation, so the marker must not veto it.
+        expect(isRecentlyRemovedColumn(board.noteId, "Gone")).toBe(false);
+        unnoteColumnRemoved(board.noteId, "Gone");
     });
 
     it("does not save anything for a duplicate column", async () => {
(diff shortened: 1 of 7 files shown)


yzxcj797:
Both findings addressed in bac4dc5ae.

**P1 (in-flight columns filtered out) — valid and reworked.** The blanket filter could indeed drop a brand-new empty column (absent from the in-flight definition and from the notes) and lose it permanently if the definition write failed. The filter now drops only values the board's own column UI removed: `removeColumn` and the old name of `renameColumn` mark the value per-board (in-memory, unmarked when the same name is added or renamed back), and `getBoardData` filters the persisted mirror by those marks. Unmarked columns always survive; the removed one still cannot survive convergence — a new test pins both directions, including the transient stale-definition refresh (it documents that the resurrection shows once via the definition leg and the write-back, then dies).

**P2 — comments condensed to the enduring invariant and the overlong added lines wrapped** (pre-existing long lines in these files left untouched).

yzxcj797:
Valid — a value recreated from another surface during the session stayed marked, and the filter would then eat its persisted entry. Fixed in 010add1e: a **discovered note value** (a note carrying the status again — recreation intent expressed outside this board's own UI, reachable from a synced client, another split, or a script) now clears the mark at the refresh that observes it. Deliberately only that leg: clearing on the stale-definition leg would unravel the #11100 fix itself, and a value present only in `board.json` with no definition and no notes is exactly the delete-vs-lost-write ambiguity the mark exists to resolve in favor of the delete. New test pins the recreation path; board suites 48/48.

yzxcj797:
The remaining residual was the recreation path that never carries a note (attribute panel / synced client writing the definition directly). Rather than grow the disambiguation heuristics, the mark now **expires after 60 s** (9c4cd854): the delete race it exists for resolves in seconds — the definition write's entity change arrives or its failure surfaces — so inside the window the delete still wins the persisted-only ambiguity, and past it a recreation reads back instead of being vetoed for the whole session. Expired entries are dropped, so the map stays bounded. Suite 49/49 including the expiry boundary cases.

yzxcj797:
The tension between this and the previous finding is real, and the precise resolution is now in e18c3b30: the marker clears **as soon as the definition write settles successfully** — no 60 s window remains for a landed write, any later appearance is a genuine recreation. A failed or never-settling write keeps the marker until the TTL, and past it the column reads back — which for a failed delete is the honest state (it did not stick; that was also the pre-#11100 behaviour for failed writes), not a resurrection bug. `storeColumns` resolves to whether the write landed so the caller reacts without re-handling the rejection; the new spec case drives both settle outcomes (failure keeps the mark, success clears it) — board suites 50/50.

yzxcj797:
Valid and subtle — the settle-based clear had exactly that hole: the racy mirror write-back could outlive the definition write, and with the marker gone the next refresh accepted the stale mirror and synced the value back into the definition. Reworked in 7d1ffadc: the marker now clears **only at full convergence** — a refresh where the value appears in *no* source: not in the definition options that refresh read (stale definition keeps it armed), not in the raw persisted list pre-filter (racy write-back keeps it armed), not on any note. That is the only state from which the column cannot return except by genuine recreation; the write-settle hook is gone entirely and the TTL remains the backstop for writes that never settle. New spec drives the three-phase sequence (stale definition + clean mirror → armed; polluted mirror → armed; converged → cleared); board suites 50/50.

cagelight:
Hope to see this reviewed and merged soon. It's currently a severe bug that makes kanban boards painfully cluttered with stale data. I just started using Trilium not too long ago and it's a bit of a glaring flaw in an otherwise great piece of software.

maphew (changes requested):
The removal marker closes the race only inside the renderer that handled the delete. `removedColumnsByNote` is module-local, so another browser client or desktop renderer can receive the attachment update before the definition update without a marker. It can resolve the stale definition, write the removed column back to `board.json`, and then sync that column into the definition again. Advise carrying removal intent across clients, or ordering and modeling the writes so no renderer can treat the intermediate state as a new column. A two-renderer ordering test should cover this sequence.

eliandoran:
Thanks for the effort, we'll handle this bug in #11246.

PR #11102, opened 2026-08-20
feat(tab-row): switch tabs with the mouse wheel when the row fits

### LINKED ISSUE #11095: An ability to switch tabs using a mouse wheel scroll
### Describe feature

Would be pretty neat to have an option to switch open tabs by hovering cursor over tabs row and just scrolling the mouse wheel, similar to Firefox' **toolkit.tabbox.switchByScrolling** option in **about:config**.

### Additional Information

_No response_

yzxcj797:
Closes #11095.

## What it does

Scrolling the wheel over the tab row now activates the adjacent tab (next on scroll-down, previous on scroll-up, wrapping around), the way Firefox's `toolkit.tabbox.switchByScrolling` does — **but only while the row has nothing to scroll horizontally**. With more tabs than fit, the wheel keeps its established job of scrolling the row, so nothing a user already relies on changes.

## Design

- The decision lives in a pure `tabWheelAction(event, container)` helper (`widget_utils.ts`), next to the horizontal-scroll wheel helper it coordinates with.
- The switch handler binds **before** the horizontal scroller: with no overflow it owns the wheel (and stops the scroller's no-op scroll); with overflow it returns `null` and the scroller is untouched.
- Modifier-held wheels (Ctrl = zoom, Alt/Shift = their own shortcuts) and momentum ticks with a sub-unit delta switch nothing.
- Both wheel axes count, so a tilted/trackpad horizontal wheel switches tabs too.
- Activation goes through the existing `activateNextTab` / `activatePreviousTab` commands, so wrapping, sub-contexts and pinned-tab behavior stay consistent with the keyboard shortcuts.

## Verification

- `widget_utils.spec.ts` (new, 5 cases): switching both directions on both axes when the row fits, null while the row overflows, every modifier guard, and the momentum-tick guard — 5/5 green.
- Repo typecheck clean for these files (the usual local `apps/web-clipper` env noise aside, which reproduces on an unmodified tree).

### DIFF
--- a/apps/client/src/widgets/tab_row.ts
+++ b/apps/client/src/widgets/tab_row.ts
@@ -12,7 +12,7 @@ import { clampDragDestination } from "../services/tab_pinning.js";
 import { buildTabTitle, TAB_TITLE_SEPARATOR } from "../services/tab_title.js";
 import utils from "../services/utils.js";
 import BasicWidget from "./basic_widget.js";
-import { setupHorizontalScrollViaWheel, tabWheelAction } from "./widget_utils.js";
+import { setupHorizontalScrollViaWheel, shouldCoalesceWheelSwitch, tabWheelAction } from "./widget_utils.js";
 
 /** One wheel gesture may switch at most one tab inside this window. */
 const WHEEL_SWITCH_COALESCE_MS = 250;
@@ -357,6 +357,8 @@ export default class TabRowWidget extends BasicWidget {
 
     /** Timestamp of the last wheel-driven tab switch, for gesture coalescing. */
     private lastWheelSwitchAt = 0;
+    /** Direction of that switch: a reversed wheel is a new gesture, not a continuation. */
+    private lastWheelSwitchAction: "next" | "previous" | null = null;
 
     doRender() {
         this.$widget = $(TAB_ROW_TPL);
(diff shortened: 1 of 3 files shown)


yzxcj797:
Both findings addressed in a7aaeddfe.

**P1 (wheel gestures skip multiple tabs) — valid.** A switch now lands at most once per 250 ms window, so one physical gesture (trackpad flicks especially) selects exactly the adjacent tab, the same coalescing browsers apply to wheel-driven tab switching.

**P2 — the overlong `wheelEvent` declaration is wrapped and the missing semicolon added.**

yzxcj797:
Valid — suppressed events left the window stale, so a gesture sustained past 250 ms slipped a second switch through. Fixed in 5fa2bc07: every qualifying event refreshes the coalescing window (gesture-end debounce), so one continuous gesture is exactly one switch and the window only re-arms after a real pause in wheel events.

yzxcj797:
Valid — an opposite-direction gesture inside the window was eaten, stranding the selection one tab from where the user started. Fixed in 236d5da1: coalescing applies to **same-direction** events only (they extend the window, so one sustained gesture is still exactly one switch); a reversed direction is a new gesture and switches immediately. Extracted as `shouldCoalesceWheelSwitch` with four cases (same-direction extends, reversed bypasses, window expiry re-arms, first switch never coalesced) — 9/9 in the suite.

maphew (commented):
The pure wheel helpers have focused coverage, but the stateful handler in `tab_row.ts` remains uncovered and `codecov/patch` is red. Advise covering command dispatch, propagation, and coalescing state at the widget-handler boundary. The issue describes an optional browser preference, while this implementation is always on, so the default interaction also needs an explicit product decision.

eliandoran:
Hi,

Thanks for the PR but I think this is better suited as a custom script instead.

Here's a draft (untested):

```js
document.addEventListener("wheel", (event) => {
    const row = event.target.closest?.(".tab-row-widget-scrolling-container");
    if (!row || event.ctrlKey || event.altKey || event.shiftKey) {
        return;
    }
    const delta = event.deltaY + event.deltaX;
    if (!delta) {
        return;
    }
    event.preventDefault();
    event.stopPropagation();
    api.triggerCommand(delta > 0 ? "activateNextTab" : "activatePreviousTab");
}, { capture: true, passive: false });
```

The idea is that switching tabs with the mouse-wheel is a a Linux-only convention that has generated user complaints in various apps, especially if it's not toggleable.

App | Ships it? | Default
-- | -- | --
VS Code | Yes — workbench.editor.scrollToSwitchTabs, added v1.45 (Apr 2020) | Off
Firefox | Yes — toolkit.tabbox.switchByScrolling, about:config only, no UI checkbox | Off
Chrome | GTK/Linux builds only | On there, off on Win/macOS
Vivaldi | Yes, exposed setting (it exposes everything) | Off
Obsidian | No — community plugin "Wheel Tab Switcher" + an open forum request | n/a

PR #11105, opened 2026-08-20
fix(ckeditor5): stop content-hint flicker on cross-element transitions

### LINKED ISSUE #11030: To-do content hint causes caret flicker and delayed caret movement in the desktop app
### Description

When Content hints are enabled, moving the caret between items in a to-do list repeatedly displays the following hint:

> Right-click or press Ctrl+Shift+Enter to change state.

In the desktop app, changing the active to-do item causes visible caret flickering, and the caret movement is noticeably delayed.

The problem is less apparent in the browser version, although the hint still appears whenever the caret moves to another item.

The hint is currently tied to CKEditor selection changes. Each move to another to-do item schedules a Bootstrap tooltip for that item, disposing the previous tooltip and creating another. Since the hint is displayed at a fixed location in the bottom-left corner, recreating it for each checkbox appears unnecessary.

Disabling **Options → Text Notes → Features → Content hints** prevents the repeated hint, flicker, and caret delay, but also disables other content hints.

Related:

- #10574 reported the same hint as distracting and was resolved by documenting the Content hints setting.
- #9881 introduced configurable multi-state to-do checkboxes and the associated tooltip.

### Steps to reproduce

1. Open or create a Text note.
2. Create a to-do list containing several items.
3. Ensure **Options → Text Notes → Features → Content hints** is enabled.
4. Move the caret between to-do items using the arrow keys or by clicking different items.
5. Repeat the navigation quickly in the desktop app.

### Actual behavior

- The “Right-click or press … to change state” hint appears after moving the caret to each to-do item.
- In the desktop app, the caret visibly flickers during these transitions.
- Caret movement between items is noticeably delayed.

### Expected behavior

Moving the caret between to-do items should be immediate, without flickering or displaying the task-state hint.

The task-state hint should appear only when hovering over a to-do checkbox. Task-state controls, right-click behavior, and the keyboard shortcut should continue to work normally.

### TriliumNext Version

0.104.1

macOS

### What is your setup?

Local (no sync)

### Operating System Version

macOS 26.5.2

### Error logs

No errors are logged.

yzxcj797:
Fixes #11030 (the flicker and caret lag; see scope note below).

## What was wrong

The content-hint popup is CSS-pinned to the bottom-left corner (`.text-editor-content-tooltip { position: fixed; inset: auto auto 1rem 1rem }`), so its Bootstrap anchor element is visually irrelevant. Yet a top-of-stack transition between elements always **disposed the live tooltip and created + faded in a fresh one**. When the caret moves between to-do items, the hint text is identical every time — so the cycle rendered the exact same pixels through a dispose/create/fade: the visible flicker and the caret lag in the desktop app, exactly as reported.

## The fix

`ContentHintManager` gains a `popupPositionIsFixed` option. When set, a cross-element transition **with unchanged content reuses the live popup**: only the bookkeeping retargets, and the auto-hide timer resets. The outgoing anchor must still be connected, so Bootstrap teardown always stays in control. Different content, unpinned managers, and detached anchors keep the full rebuild. Both fixed-corner consumers (to-do multistate, collapsible summary hints) opt in; the collapsible **drag-handle** manager keeps Bootstrap's own anchored placement and deliberately does not.

## Scope note

The report also asks for hover-only hints (no caret-driven hint at all). That part is a design change to the deliberate dwell behaviour ("a stationary caret pops the checkbox tooltip") introduced with the multistate work, so I've left it out — this PR fixes the defect (flicker/lag) without changing when hints appear.

## Testing

`content_hint_manager.spec.ts` — 32/32 under jsdom (this package's browser-mode runner can't start here; the manager spec needs no CKEditor toolbar, so jsdom exercises real Bootstrap tooltips against the DOM), with three new cases: a pinned manager reuses the same popup DOM node across an element switch with unchanged content; it still rebuilds when content differs; an unpinned manager still rebuilds on element switch (anchored placement must follow the new anchor). The plugin suites' jsdom failures are pre-existing environment gaps (no `ResizeObserver`) — identical 52 with and without this change; CI's browser mode is their gate.

### DIFF
--- a/packages/ckeditor5/src/content_hint_manager.spec.ts
+++ b/packages/ckeditor5/src/content_hint_manager.spec.ts
@@ -88,6 +88,51 @@ describe("ContentHintManager", () => {
             expect(livePopupText()).toBe("A tip");
         });
 
+        it("reuses the live popup across an element switch with unchanged content when the popup is CSS-pinned", () => {
+            const pinned = new ContentHintManager({popupPositionIsFixed: true});
+            const handleA = pinned.createHandle(a, "same tip");
+            const handleB = pinned.createHandle(b, "same tip");
+            handleA.show();
+            const popup = livePopup();
+            expect(popup).not.toBeNull();
+
+            handleB.show();
+            // The same DOM node survives the switch — no dispose/create/fade.
+            expect(livePopup()).toBe(popup);
+            expect(livePopupText()).toBe("same tip");
+
+            handleB.hide();
+            expect(livePopup()).toBe(popup);
+            expect(livePopupText()).toBe("same tip");
+            pinned.destroy();
+        });
+
+        it("still rebuilds the popup across an element switch when content differs", () => {
+            const pinned = new ContentHintManager({popupPositionIsFixed: true});
+            const handleA = pinned.createHandle(a, "tip one");
+            const handleB = pinned.createHandle(b, "tip two");
+            handleA.show();
+            const popup = livePopup();
+
+            handleB.show();
+            expect(livePopup()).not.toBe(popup);
+            expect(livePopupText()).toBe("tip two");
+            pinned.destroy();
+        });
+
+        it("still rebuilds across an element switch when the popup is anchored", () => {
+            const handleA = manager.createHandle(a, "same tip");
+            const handleB = manager.createHandle(b, "same tip");
+            handleA.show();
+            const popup = livePopup();
+
+            handleB.show();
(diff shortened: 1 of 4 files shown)


maphew (commented):
No correctness findings. The fixed-position opt-in is limited to the two pinned consumers, while anchored hints retain the rebuild path. Advise removing the unrelated line-ending churn before merge so the effective manager and consumer changes remain reviewable. The overlap with hover-only PR 11031 also needs one intended UX direction.

PR #11108, opened 2026-08-20
fix(desktop): resolve a working Linux autostart Exec for unbundled Electron

### LINKED ISSUE #10918: Linux autostart entry is broken for unbundled-Electron packages (falls back to process.execPath)
### Description

`applyLaunchOnStartup()` writes a broken `Exec=` line into the Linux autostart `.desktop` file whenever Electron is "unbundled": system Electron package plus a separate `app.asar` plus a wrapper script that sets up env vars and the app path, instead of one self-contained executable.

The relevant code, in [`apps/desktop/src/services/auto_launch.ts`](https://github.com/TriliumNext/Trilium/blob/main/apps/desktop/src/services/auto_launch.ts):

```js
function buildLinuxDesktopEntry(hidden: boolean) {
    // When packaged as an AppImage, APPIMAGE points at the bundle to relaunch;
    // otherwise fall back to the running executable.
    const exec = process.env.APPIMAGE ?? process.execPath;
    ...
}
```

`process.execPath` only works when the running process *is* the full app: an AppImage, or an electron-builder Linux target that bundles its own Electron runtime (the official `.deb`/`.rpm`/`.flatpak`/`.AppImage` builds all do this). On a "system Electron + wrapper script" install, it doesn't. Example: the Arch `triliumnext-bin` AUR package depends on the `electron42` system package and launches through a `/usr/bin/triliumnext` wrapper that sets `LD_LIBRARY_PATH`, `NODE_ENV`, etc. and finally runs `exec electron42 "$RUNNAME" "$@"` with `RUNNAME=/usr/lib/triliumnext/app.asar`. Here `process.execPath` is just the bare `electron42` binary, with no app path attached.

So the generated `.desktop` file ends up as:
```
Exec="/usr/lib/electron42/electron" --start-hidden
```
which runs generic Electron with no app to load. It just prints its CLI help to stderr and exits; Trilium never starts. And since `applyLaunchOnStartup()` reruns on every startup ("idempotent, safe to call repeatedly" per the docstring), it keeps rewriting the file back to this broken state, so hand-fixing the `.desktop` file doesn't stick either.

### TriliumNext Version

0.104.1 (still present on `main` as of writing)

Other Linux

### What is your setup?

Local (no sync)

### Operating System Version

Arch Linux (rolling), `triliumnext-bin` from the AUR. Unofficial package, depends on system `electron42` rather than bundling its own Electron.

### Error logs

`journalctl --user -u app-trilium@autostart.service` after a login-triggered autostart:

```
electron[NNNN]: Electron 42.8.0 - Build cross platform desktop apps with JavaScript, HTML, and CSS
electron[NNNN]: Usage: electron [options] [path]
electron[NNNN]: A path to an Electron app may be specified. It must be one of the following:
electron[NNNN]:   - index.js file.
electron[NNNN]:   - Folder containing a package.json file.
electron[NNNN]:   - Folder containing an index.js file.
electron[NNNN]:   - .html/.htm file.
electron[NNNN]:   - http://, https://, or file:// URL.
electron[NNNN]: Options:
electron[NNNN]:   -i, --interactive     Open a REPL to the main process.
electron[NNNN]:   -r, --require         Module to preload (option can be repeated).
electron[NNNN]:   -v, --version         Print the version.
electron[NNNN]:   -a, --abi             Print the Node ABI version.
```
Exits 0, so systemd/KDE log it as a successful run even though Trilium never opened.

### Possible fix

`APPIMAGE` is the only override that exists right now for "the running process isn't what should be relaunched." A second env var, checked the same way, would let any unbundled-Electron packager (Arch, probably others with a similar no-bundled-libs policy) point at the right command:

```js
const exec = process.env.APPIMAGE ?? process.env.TRILIUM_LAUNCH_EXEC ?? process.execPath;
```

Can send a PR for this plus a test case in `auto_launch.spec.ts` (which already covers the `APPIMAGE` fallback). Figured I'd float the env var name/approach first since it's a new public contract.

yzxcj797:
## Summary
**What**

The Linux autostart `.desktop` entry's `Exec=` line is now resolved from four layered sources instead of `APPIMAGE ?? process.execPath`:

1. `$APPIMAGE` — AppImage builds relaunch the bundle (unchanged).
2. `$TRILIUM_LAUNCH_EXEC` — the explicit packager override proposed in the issue, for unbundled-Electron wrappers that must not be bypassed.
3. The app's **installed launcher** — the trilium `.desktop` file shipped by the distro package; its `Exec=` (the wrapper script) is reused verbatim with freedesktop field codes (`%f`, `%U`, …) stripped. User-local applications dirs are scanned before system ones so a user override wins.
4. `process.execPath` — bundled-Electron builds where the executable *is* the app (previous behaviour).

**Why**

#10918 — on a "system Electron + wrapper script" install (Arch `triliumnext-bin` and similar), `process.execPath` is the bare `electron42` binary with no app attached, so the autostart entry launches generic Electron, which prints its usage and exits — Trilium never starts. Because `applyLaunchOnStartup()` rewrites the entry on every launch (by design, to repair drift), hand-fixing the file doesn't stick either.

Layer 3 is what repairs existing installs with no packager changes: the wrapper exists precisely because it sets up `LD_LIBRARY_PATH`/`app.asar`, and the distro's own launcher already points at it. Layer 2 adds the explicit contract from the issue for packagers who want determinism rather than detection.

**Testing**

`apps/desktop/src/services/auto_launch.spec.ts`, 16/16 (12 existing + 4 new):

- `honours $TRILIUM_LAUNCH_EXEC as an unbundled-Electron packager override` — bare system-electron `execPath` + the env var → `Exec="/usr/bin/triliumnext"`.
- `reuses the installed launcher's Exec when execPath is a bare system Electron` — finds `triliumnext.desktop` among unrelated entries, reuses `Exec=/usr/bin/triliumnext`, and strips the `%U` field code.
- `prefers a user-local launcher over the system package's` — `~/.local/share/applications` beats `/usr/share/applications`.
- Existing `execPath`-fallback tests still pass: detection is best-effort and falls through on missing dirs / parse failures, so it can never break autostart application.

`tsc --noEmit` on `apps/desktop` is clean.

Fixes #10918

### DIFF
--- a/apps/desktop/src/services/auto_launch.spec.ts
+++ b/apps/desktop/src/services/auto_launch.spec.ts
@@ -14,6 +14,8 @@ const state = vi.hoisted(() => ({
     mkdirSync: vi.fn(),
     writeFileSync: vi.fn(),
     rmSync: vi.fn(),
+    readdirSync: vi.fn(),
+    readFileSync: vi.fn(),
     log: { info: vi.fn(), error: vi.fn() }
 }));
 
@@ -37,7 +39,11 @@ vi.mock("fs", () => ({
     default: {
         mkdirSync: (...a: unknown[]) => state.mkdirSync(...a),
         writeFileSync: (...a: unknown[]) => state.writeFileSync(...a),
-        rmSync: (...a: unknown[]) => state.rmSync(...a)
+        rmSync: (...a: unknown[]) => state.rmSync(...a),
+        // Detection must survive these being unimplemented (undefined return):
+        // a missing applications directory is the normal case, not an error.
+        readdirSync: (...a: unknown[]) => state.readdirSync(...a),
+        readFileSync: (...a: unknown[]) => state.readFileSync(...a)
     }
 }));
 
@@ -73,7 +79,9 @@ function setExecPath(p: string) {
 }
 
 beforeEach(() => {
-    vi.clearAllMocks();
+    // reset (not clear): the detection tests install per-test fs implementations
+    // that must not leak into the execPath-fallback tests after them.
+    vi.resetAllMocks();
     state.launchOnStartup = false;
     state.hideOnAutoStart = false;
     state.setLoginItemThrows = false;
@@ -81,6 +89,9 @@ beforeEach(() => {
     state.ipcOn.clear();
     delete process.env.APPIMAGE;
     delete process.env.XDG_CONFIG_HOME;
+    delete process.env.TRILIUM_LAUNCH_EXEC;
+    delete process.env.XDG_DATA_HOME;
+    delete process.env.XDG_DATA_DIRS;
 });
 
 afterEach(() => {
@@ -88,6 +99,9 @@ afterEach(() => {
     setExecPath(ORIGINAL_EXECPATH);
     delete process.env.APPIMAGE;
     delete process.env.XDG_CONFIG_HOME;
+    delete process.env.TRILIUM_LAUNCH_EXEC;
+    delete process.env.XDG_DATA_HOME;
+    delete process.env.XDG_DATA_DIRS;
 });
 
 describe("auto_launch", () => {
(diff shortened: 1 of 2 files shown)


maphew (changes requested):
`readDesktopExec()` removes only whitespace-separated tokens that begin with a field code. Desktop entries can embed codes in arguments and use `%%` for a literal percent. Advise applying the desktop-entry Exec field-code rules instead of token filtering, with focused coverage for embedded codes and `%%`.

eliandoran (changes requested):
Layer 3 is an heuristic approach, looking for the .desktop file can result in some ugly situations, especially if the user has Trilium in different packaging methods. Drop that layer and we should be good. If we have an environment variable to configure the override, it's the responsibility of the downstream package to handle it (e.g. AUR, NixOS).

PR #11112, opened 2026-08-20
Code-note scrolling performance: note-list latch, contained scroller, hidden-editor guard, instant jumps

eliandoran:
Four fixes from profiling code-note scrolling (50,000-line / 1.8 MB JavaScript note with 25 children, identical 150-step wheel-scroll input per scenario, headless Chromium against the in-memory fixture server, dev build — absolute times are inflated ~1.5× vs production, ratios are the result). One commit per fix.

The starting point: CodeMirror 6's viewport virtualization **survives the flowing layout** (44–65 rendered lines out of 50,000 in every scenario), but the same wheel input dropped **18% of frames** in today's layout vs **~0%** with the editor scrolling internally — and about two-thirds of that gap came from the note list being mounted in the scrolled flow, widening the style-recalc scope of every scroll-driven CodeMirror measure cycle (`UpdateLayoutTree` 1,944 ms vs 747 ms for the same input).

### 1. `fix(client)`: don't latch the note list visible while the content is loading

The legacy list/grid views defer rendering until scrolled into view (IntersectionObserver with a one-way latch), but the observer started while the note's content was still loading — when the widget still sat inside the viewport — so the latch tripped on effectively every note before the content pushed the list below the fold. The note list now waits for the note context's `contentLoad` state before observing, and drops the latch when a new load starts (which also resets it across note switches).

### 2. `feat(client)`: scroll childless code notes inside the editor

`editableCode`/`readOnlyCode` are now `isFullHeight`, so a code note without a note list gets the contained layout (the markdown precedent): CodeMirror owns scrolling, and the inline title stays put instead of scrolling away. Notes **with** children keep the flowing layout so the note list stays reachable below the content. Also drops the scroll-past-end padding in full-height containers (it only stole height there — visible today as dead space under the backend log) and gives the read-only content element the `height: 100%` its scroller needs. Promoted attributes now start collapsed on such notes, matching other full-height types.

### 3. `fix(codemirror)`: don't render the whole document into a hidden editor

`setText` while the editor sits in a `display: none` subtree (e.g. a restored background tab that received its content after being covered) mapped the viewport onto the entire new document, and a hidden editor can never measure to shrink it back — reproducibly leaving 50,044 `.cm-line` elements resident in the DOM. Hidden editors now rebuild the view from the updated state, resetting the viewport to the initial estimate; the first measure after becoming visible sizes it properly.

### 4. `fix(client)`: make programmatic scrolling in code notes instant

`scroll-behavior: smooth` on the scrolling container made every scripted jump (go to line, search match, scroll to end) animate through the whole flowing document while CodeMirror rendered and discarded each intermediate viewport: 710 ms to settle vs 12 ms instant on the 50k-line note. Scoped to code-type splits; text notes keep smooth anchor navigation.

### Validation

Re-measured on the fixed build with the same input:

| scenario | frames >32 ms before | after |
|---|---|---|
| childless code note | 18% (flowing) | **0%** (contained) |
| code note + 25 children | 18% | **~1%** — note list no longer pre-latched, still mounts when the bottom is actually reached |
| scroll-to-end settle | 710 ms | **13–14 ms** |

`pnpm typecheck` clean; client specs (411, incl. new latch tests) and codemirror specs (342, incl. new hidden-`setText` tests) pass. The hidden-editor guard is pinned by unit tests plus the pre-fix reproductions; the exact tab arrangement that produced the hidden split did not re-form on a clean fixture, so that app-level path is covered indirectly.

https://claude.ai/code/session_01J3sYLWp4QzutEDpEtpu1WB

---
_Generated by [Claude Code](https://claude.ai/code/session_01J3sYLWp4QzutEDpEtpu1WB)_


maphew (commented):
No technical findings. The note-list latch reset, contained code-note layout, hidden-editor state reset, and code-only instant scrolling are narrowly scoped and have focused regression coverage. Advise rerunning current CI and checking contained layout once in an actual restored hidden tab because the PR identifies that app-level path as indirectly validated.

PR #11120, opened 2026-08-21
feat(client): let a widget note return several widgets

BeatLink:
## What

A `#widget` note's script bundle returns a single value, and `getWidgetBundlesByParent` registers exactly one widget from it. Returning several widgets from one note meant splitting them across notes and sharing code through a child module note, even when the widgets were parts of one feature.

The bundle result is now normalized to a list, so a widget note can return an array:

```js
module.exports = [ new TreeWidget(), new SidebarWidget() ];
```

```jsx
export default [
    defineWidget({ parent: "left-pane", render: () => <span>Left pane</span> }),
    defineWidget({ parent: "right-pane", render: () => <span>Right pane</span> })
];
```

## How

`asWidgetList()` in `apps/client/src/services/bundle.ts` wraps a non-array result and drops falsy entries, and the registration loop iterates it. Each entry then goes through the existing `WidgetsByParent.add()`, so widgets from one note keep independent parents and positions, and a parentless entry raises its own error toast without dropping its siblings. A single returned widget takes the same path it did before, and `return undefined` still means "no widget".

Nothing changes on the server: the bundle format and the `script/widgets` route are untouched. `WidgetsByParent.get()` and its consumers already spread N widgets per parent.

Widgets from one note share that note's `noteId`, so an error toast from any of them links back to the same script note.

## Tests

Three cases added to `bundle.spec.ts`: a mixed legacy + Preact array (including a null entry) registering under two different parents, an array with one parentless entry, and an empty array.

- `pnpm --filter client test bundle.spec` — 23/23 pass
- `pnpm typecheck` — no errors

## Docs

New "Multiple widgets in a single note" section in `docs/User Guide/User Guide/Scripting/Frontend Basics/Custom Widgets.md`, with both the jQuery and Preact forms.


BeatLink:
Bump

maphew (approved):
No technical findings. Result normalization preserves the single-widget path, while each array entry still passes through the existing parent validation and note attribution. Focused tests cover mixed and partial-error cases.

eliandoran (approved):

Issue #10907, opened 2026-08-07
Unable to switch templates on a note after applying a template once.


DtDEncryption:
### Description

It would be successful to apply an existing template to a newly created note. But once the new note is applied a random template just once, it can't be applied with any template afterwards, even if the note has been cleared again.

### TriliumNext Version

0.104.1

### What operating system are you using?

Other Linux

### What is your setup?

Local (no sync)

### Operating System Version

Fedora 44 KDE Plasma

### Error logs

[trilium-2026-08-07.log](https://github.com/user-attachments/files/30811196/trilium-2026-08-07.log)

DtDEncryption:
Confirm the introduced workaround works.

But it would be better if the content check logic be more user-friendly or if there be some popup showing what goes wrong when it fails to switch the templates. The same goes for the "Main Menu", which rests at the top left icon of the GUI. The design like that was kind of confusing for me when I searched for "zen mode" option found in official documents.

In a word, UX problems. So I will close the bug report and open a feature request.

PR #11125, opened 2026-08-21
fix(templates): apply template content to a note emptied in the editor

### LINKED ISSUE #10908: Smoothen the process of switching templates for a note.
### Describe feature

From the info introduced by bot in [a mistaken bug report](https://github.com/TriliumNext/Trilium/issues/10907), I found that the content check seems to be the culprit for the failure at switching tamplates freely even though the body of the note has been emptied.

If so, I hope the content check could be more user-friendly, either by providing an option for user to confirm their action or by showing a error popup to help user understand what went wrong.

### Additional Information

I think such an improvement is important, because it's not rare for the user to select a wrong template due to a misclick. Besides, a friendly design would make the note transition a lot easier as the user may continuously improve templates for certain workflows, therefore creating a need to update the old notes somewhat constantly.

yzxcj797:
## Summary

**What**

The template-relation handler's "don't overwrite existing content" gate now uses a shared `isVisuallyEmptyTextContent` helper: content consisting solely of block scaffolding — empty `<p>`/`<div>`/`<br>` skeletons, `&nbsp;`/`&#160;` entities, and HTML comments — counts as empty. Notes with any real text still refuse the overwrite, exactly as before.

**Why**

#10908 — switching templates on a note whose body the user emptied did nothing, silently. The handler compares the stored blob against `''`/whitespace, but a note emptied in the editor stores the serialized empty paragraph (`<p>&nbsp;</p>` or `<p><br></p>`), so the check read it as content: no template content copied, no `type`/`mime` adoption (both live inside the same guard), and no feedback. The reporter's linked #10907 hit the same gate. The check is the culprit the community bot pointed at; this makes it see what the user sees.

**Scope**: the emptied-note case now just works — the primary complaint. The overwrite-refusal for genuinely non-empty notes (and its silent nature) is unchanged behavior; a confirmation affordance for that half would be a separate product decision.

**Testing** (via the apps/server suite that runs core specs):

- New `content-emptiness.spec.ts` (3 cases): empty/whitespace; the editor skeletons from the report (`<p>&nbsp;</p>`, `<p><br></p>`, `<div><br></div>`, entity-only, comment-only); real content stays non-empty (text, mixed, list with labels, plain text).
- `handlers.spec.ts` +2 (20 total): template content now copies into a skeleton-content note — **verified to fail on the pre-fix handler** (no copy) — and still refuses a note with a real draft. Suite 20/20.

Fixes #10908

### DIFF
--- a/packages/trilium-core/src/services/content-emptiness.spec.ts
+++ b/packages/trilium-core/src/services/content-emptiness.spec.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from "vitest";
+
+import { isVisuallyEmptyTextContent } from "./content-emptiness.js";
+
+describe("isVisuallyEmptyTextContent", () => {
+    it("treats empty and whitespace-only content as empty", () => {
+        expect(isVisuallyEmptyTextContent("")).toBe(true);
+        expect(isVisuallyEmptyTextContent("   \n\t ")).toBe(true);
+    });
+
+    it("treats the editor's emptied-note skeletons as empty (#10908)", () => {
+        expect(isVisuallyEmptyTextContent("<p>&nbsp;</p>")).toBe(true);
+        expect(isVisuallyEmptyTextContent("<p><br></p>")).toBe(true);
+        expect(isVisuallyEmptyTextContent("<div><br></div>")).toBe(true);
+        expect(isVisuallyEmptyTextContent("<p>  </p>\n<p>&#160;</p>")).toBe(true);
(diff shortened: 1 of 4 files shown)


maphew (changes requested):
Advise limiting HTML skeleton normalization to HTML text notes. The handler calls `isVisuallyEmptyTextContent()` for every string-content note, so a code or plain-text note whose literal visible source is `<p><br></p>` is classified as empty and can be overwritten when a template relation is added. Non-HTML string content should retain the previous empty or whitespace check. The CKEditor emptied-note regression is otherwise covered directly.

PR #11126, opened 2026-08-21
fix(share): harden the generated stylesheet embedded in shared pages

eliandoran:
Tightens how the generated icon-pack and task-state CSS is produced and embedded on shared pages.

Values coming from note content were interpolated into that stylesheet without escaping. This validates and escapes them at the point they are generated, and escapes the stylesheet again where the share template embeds it, so the two paths behave consistently.

Also brings the share render path in line with how the client already consumes the same string.

### Notes

- No change for existing icon packs — the bundled Boxicons manifest and anything produced by `apps/icon-pack-builder` are byte-identical before and after.
- Task-state colours set through the settings UI are unaffected.
- User Guide updated with the manifest key constraint, which was previously undocumented.

### Testing

- New unit tests for the escaping helpers, `generateCss()`, and `generateTaskStateCss()`.
- New integration test covering the rendered share page; verified it fails without the change.
- Core specs pass under both the server and standalone runtimes; `pnpm typecheck` clean.

https://claude.ai/code/session_015hqy571eVbremDWu7ZyzD2

---
_Generated by [Claude Code](https://claude.ai/code/session_015hqy571eVbremDWu7ZyzD2)_


maphew (commented):
Advise deciding compatibility for existing custom icon manifests before making icon keys ASCII-only. `generateCss()` now skips every key outside alphanumerics, dashes, and underscores, while the current guide says only that keys should match the font's class names. An existing CSS-usable escaped key can therefore disappear after upgrade. Escaping selector identifiers, or documenting and testing this as an intentional migration, would avoid a silent compatibility break. The other stylesheet-boundary and CSS-string escaping changes address the traced injection paths.

PR #11141, opened 2026-08-22
fix(search): keep commas inside quoted operands

### LINKED ISSUE #11132: Search: the lexer strips commas even inside quotes, so the documented #geolocation="48.8583,2.2945" can never be matched
### Description

**The short version: the search lexer discards every comma it sees, including commas inside a quoted string. The Geo Map manual tells the user to store a latitude and longitude separated by a comma and to wrap the value in quotes. That exact value is then unsearchable by equality, in every quoting form, and the query answers HTTP 200 with an empty list rather than an error.**

The stored value, read back from the attribute API first so a write problem is ruled out, and then the three ways a user might quote it:

```
GET /api/notes/l65BjI6IhuYL/attributes
  -> 200  stored geolocation value(s): ["48.8583,2.2945"]      the comma IS in the store

GET /api/search/#geolocation="48.8583,2.2945"   -> 200 []
GET /api/search/#geolocation=48.8583,2.2945     -> 200 []
GET /api/search/#geolocation='48.8583,2.2945'   -> 200 []
```

The lexed operand for `"48.8583,2.2945"` is `48.85832.2945`, which equals nothing that was ever stored, and nothing anyone would ever store.

The impact is wider than the geo map, which is only where the manual walks the user into it: **any label whose value contains a comma is unreachable by equality search.** Tags, coordinates, ranges, author lists, anything with a comma in it.

#### What the manual promises

`docs/User Guide/User Guide/Collections/Geo Map.md`, lines 73 and 75, under "How the location of the markers is stored":

> The location of a marker is stored in the `#geolocation` attribute of the child notes:
>
> This value can be added manually if needed. **The value of the attribute is made up of the latitude and longitude separated by a comma.**

and line 157, in "Adding from Google Maps", instructing the user to author precisely the form that fails:

> Then paste the value inside the text box into the `#geolocation` attribute of a child note of the map (**don't forget to surround the value with a `"` character**).

So the manual specifies the comma, specifies the quotes, and the result cannot be found by the search that the rest of the manual documents.

#### Reproduction

Run on a server built from source at `main` @ `86a9715b09b4fc523764eee3e2ba08b5f58ef12b` (`package.json` version 0.105.0), `TRILIUM_ENV=production`, a fresh `TRILIUM_DATA_DIR`, an empty document (no demo database). Every request carries a normal session cookie and the paired CSRF header.

1. Create a note and give it `#geolocation="48.8583,2.2945"` (the Paris coordinates from the manual's own worked example). In the run below it is `l65BjI6IhuYL`.
2. Create a second note with a comma-free value in the same shape, `#tcnocomma="48.8583"`, as the control. In the run below it is `gP4nDim7wyWo`.
3. Read the stored value back, so the rest of the run is about search and not about storage.
4. Search for the value.

The probe is the three-line block above: all three quoting forms answer `200 []`.

#### Control: it is the comma, not the attribute, not the quoting, not the route

Same process, same fixture, same request shape, run immediately after the probe:

```
GET /api/search/#geolocation               -> 200 ["l65BjI6IhuYL"]   the label exists and is reachable
GET /api/search/#tcnocomma="48.8583"       -> 200 ["gP4nDim7wyWo"]   identical shape, no comma, matches
GET /api/search/#geolocation *=* 48.8583   -> 200 ["l65BjI6IhuYL"]   a contains operator still reaches the note
```

Bare label existence works. A quoted equality on a comma-free value works. A substring operator reaches the very note that equality cannot. The single variable between the working control and the failing probe is one comma inside the quotes.

#### The escape hatch, which is real and undocumented

```
GET /api/search/#geolocation="48.8583\,2.2945"  -> 200 ["l65BjI6IhuYL"]
```

A backslash-escaped comma does match, which also proves the rest of the pipeline is sound: the value is stored correctly, the comparator is correct, and only the lexing of the operand is wrong. Nothing in the geo-map pages, or anywhere in the search documentation, mentions escaping. We mention it as a workaround for anyone who lands here, not as the fix.

#### Cause

Read at `86a9715b`, and unchanged since the commit this was first found on. All of it is in `packages/trilium-core/src/services/search/services/lex.ts`.

**1. The comma is declared an operator character**, `:15-17`:

```ts
function isSymbolAnOperator(chr: string) {
    return ["=", "*", ">", "<", "!", "-", "+", "%", ","].includes(chr);
}
```

**2. The unquoted branch opens at `:91`**:

```ts
} else if (!quotes) {
```

**and closes at `:123`.**

**3. The comma is discarded at `:125-127`, after that close**, so it runs in every lexer state, quoted included:

```ts
if (chr === ",") {
    continue;
}

currentWord += chr;
```

That is the whole defect: three lines that belong inside the `!quotes` branch and sit just below it. Everything above them handles quoting correctly, which is why the quoted string survives intact except for its commas.

A fix that keeps the current unquoted behaviour would be to move the `continue` inside the `!quotes` block, or to guard it with `!quotes`. We have not opened a pull request because there may be a reason the comma is stripped this late that is not visible from outside; if a patch would be welcome, say so and we will send one with a test.

#### Related

We could not find an existing issue for this. Searches over the tracker for comma and geolocation, and for comma and attribute values, returned nothing, and a sweep of every issue created since 2026-08-01 contains nothing related.

As of 2026-08-20, open pull request 10633 (https://github.com/TriliumNext/Trilium/pull/10633) does patch `lex.ts`, with a single hunk around `:93-105` that adds two-character fuzzy-operator handling. It does not touch the comma line, so that pull request does not fix this.

#### Suggested labels

Our account is an outside contributor and cannot apply labels itself. Based on the vocabulary in current use: `State: Triage`, `BE`, `search`.

### TriliumNext Version

0.105.0. Built from source at `main` @ `86a9715b09b4fc523764eee3e2ba08b5f58ef12b`, whose root `package.json` declares `"version": "0.105.0"`. The file involved (`packages/trilium-core/src/services/search/services/lex.ts`) is byte-identical at that commit and at tag `v0.105.0`, so the shipped release behaves the same way.

macOS

### What is your setup?

Server access only

### Operating System Version

macOS 26.5 (build 25F71), Apple silicon. Node v24.14.1, pnpm 11.22.0. The server ran locally from the source build described above and was reached over HTTP on `127.0.0.1:8099`; no desktop client and no sync were involved.

### Error logs

There is nothing to attach. All three failing queries return HTTP 200 with a well-formed empty array. No parse error is raised, no warning is logged, and the user is told only that their note does not exist, which is not true. A lexer that silently rewrites the operand it was given is the one place where a visible error would have made this a five-minute discovery rather than a puzzle.

This finding came from running the product's published documentation against a live instance. The full transcript, including the attribute read-back and the raw probe script, is available on request.

yzxcj797:
## What

The search lexer no longer strips commas inside quoted strings. `#geolocation="48.8583,2.2945"` now lexes to the operand `48.8583,2.2945` and matches the stored attribute value, in double and single quotes. Comma stripping outside quotes (fulltext noise like `europe, austria`) is unchanged.

## Why (issue #11132)

`lex.ts` dropped every comma unconditionally:

```ts
if (chr === ",") {
    continue;
}
```

The quote-handling branch above it only manages the quote state, so a quoted operand fell straight through the strip. The Geo Map manual documents storing coordinates exactly as `#geolocation="48.8583,2.2945"` (lat and long separated by a comma, value wrapped in quotes), which made that documented value — and any comma-containing label value — unreachable by equality search: the lexed operand became `48.85832.2945`, a string nobody ever stored, and the API answered `200 []` instead of an error.

Downstream code was already ready: `resolveConstantOperand` returns an `inQuotes` operand verbatim, so the lexer was the only place the value was mangled.

## How

One-line condition plus a comment:

```ts
if (chr === "," && !quotes) {
    continue;
}
```

## Testing

- `lex.spec.ts` — new cases: quoted fulltext keeps the comma (`'48.8583,2.2945'` → one token), and `#geolocation="48.8583,2.2945"` / single-quoted form produce `["#geolocation", "=", "48.8583,2.2945"]` with `inQuotes: true`.
- `search.spec.ts` — end-to-end: two notes with comma-valued `#geolocation` labels, each found by its exact quoted value in both quote styles; a non-matching quoted value still returns nothing.
- Diff-verified: with only the `lex.ts` change stashed, the three new tests fail exactly as reported (operand loses its comma); with it applied, the full search suite passes — 31 files, 357 tests.

Fixes #11132

### DIFF
--- a/packages/trilium-core/src/services/search/services/lex.spec.ts
+++ b/packages/trilium-core/src/services/search/services/lex.spec.ts
@@ -20,6 +20,13 @@ describe("Lexer fulltext", () => {
         expect(lex("'i can use \" or ` or #~=*' without problem").fulltextTokens.map((t) => t.token)).toEqual(['i can use " or ` or #~=*', "without", "problem"]);
     });
 
+    it("commas are noise in fulltext but kept inside quotes", () => {
+        expect(lex("europe, austria").fulltextTokens.map((t) => t.token)).toEqual(["europe", "austria"]);
+
+        // Quoted values must survive verbatim — e.g. the Geo Map manual's #geolocation="48.8583,2.2945".
+        expect(lex("'48.8583,2.2945'").fulltextTokens.map((t) => t.token)).toEqual(["48.8583,2.2945"]);
+    });
+
     it("I can use backslash to escape quotes", () => {
         expect(lex('hello \\"world\\"').fulltextTokens.map((t) => t.token)).toEqual(["hello", '"world"']);
 
@@ -104,6 +111,20 @@ describe("Lexer expression", () => {
             { token: "*=*", inQuotes: false, startIndex: 6, endIndex: 8 },
             { token: "text", inQuotes: true, startIndex: 10, endIndex: 13 }
         ]);
+    });
+
+    it("keeps commas inside a quoted operand", () => {
+        // The Geo Map manual stores coordinates as #geolocation="48.8583,2.2945"; the lexer used
+        // to strip the comma even inside quotes, making the stored value unreachable (#11132).
+        expect(lex('#geolocation="48.8583,2.2945"').expressionTokens.map((t) => t.token)).toEqual(
+            ["#geolocation", "=", "48.8583,2.2945"]
+        );
+        expect(lex("#geolocation='48.8583,2.2945'").expressionTokens.map((t) => t.token)).toEqual(
+            ["#geolocation", "=", "48.8583,2.2945"]
+        );
+
+        const quoted = lex('#geolocation="48.8583,2.2945"').expressionTokens[2];
+        expect(quoted.inQuotes).toBe(true);
     });
 
     it("simple label operator with param without quotes", () => {
--- a/packages/trilium-core/src/services/search/services/lex.ts
+++ b/packages/trilium-core/src/services/search/services/lex.ts
@@ -122,7 +122,9 @@ function lex(str: string) {
             }
         }
 
-        if (chr === ",") {
+        // Commas are stripped as fulltext noise, but not inside quotes — a quoted operand
+        // (e.g. #geolocation="48.8583,2.2945") must keep the exact value the user stored.
+        if (chr === "," && !quotes) {
             continue;
         }
 
--- a/packages/trilium-core/src/services/search/services/search.spec.ts
+++ b/packages/trilium-core/src/services/search/services/search.spec.ts
@@ -112,6 +112,26 @@ describe("Search", () => {
         expect(findNoteByTitle(searchResults, "Austria")).toBeTruthy();
     });
 
+    it("label comparison with a comma in a quoted value", () => {
+        rootNote.child(note("Eiffel Tower").label("geolocation", "48.8583,2.2945")).child(note("Charles Bridge").label("geolocation", "50.0865,14.4114"));
+
+        const searchContext = new SearchContext();
+
+        // The Geo Map manual documents exactly this form; the lexer used to strip the comma
+        // inside the quotes, so the stored value could never be matched (#11132).
+        let searchResults = searchService.findResultsWithQuery('#geolocation="48.8583,2.2945"', searchContext);
+        expect(searchResults.length).toEqual(1);
+        expect(findNoteByTitle(searchResults, "Eiffel Tower")).toBeTruthy();
+
+        searchResults = searchService.findResultsWithQuery("#geolocation='50.0865,14.4114'", searchContext);
+        expect(searchResults.length).toEqual(1);
+        expect(findNoteByTitle(searchResults, "Charles Bridge")).toBeTruthy();
+
+        // A quoted value that matches no stored value still returns nothing.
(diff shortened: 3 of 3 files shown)


yzxcj797:
Noted on both — these are intentional: the spec file's established formatting keeps a full query assertion on one line (see the neighboring `note.title *=* DbPropsProbe AND note.contentSize >= ...` rows), and its header/comments document the regression each block guards, so the new tests follow the file's voice rather than splitting lines or stripping the #11132 context that explains why the comma assertions exist.

truecourse-agent:
I reported #11132, where a search value with a comma inside quotes matched nothing.

✅ Re-ran that case on this branch and it now returns the expected note. The whole flow passes.

maphew (approved):
No technical findings. The lexer change preserves unquoted comma handling while retaining commas in quoted operands. The expression and end-to-end tests cover both quote forms and exact-value behavior.

eliandoran (approved):

liana-j:
Thanks for merging this, Elian. Genuinely appreciated. Sent you an email with the longer version.
Liana, DevRel at TrueCourse

PR #11143, opened 2026-08-22
fix(saved search): run the saved search from "Search now" instead of creating a new empty one

### LINKED ISSUE #11130: Saved search: the "Search now" button does not run the saved search, it creates a new empty search note and shows every note in the instance
### Description

**The short version: opening a saved search shows "Search has not been executed yet." with one button, `Search now`. Pressing it navigates away to a different, brand-new search note whose search string is empty, and renders every note in the instance. The only control the widget offers for running the saved search does not run the saved search.**

Before the click, the saved search renders its own message and offers exactly one control:

Saved search before the click: search string reads #tcfindme, the body says "Search has not been executed yet." and offers a single "Search now" button

After pressing `Search now`, the search string is empty and the list is the whole instance:

After the click: the search string field is empty and showing its placeholder, and the result list shows every note in the instance

The click, with the note ids left in:

```
url before click: .../#root/_hidden/_search/NpJd5jXZN9QK/fNpO06ehjjK3?ntxId=fIN8fe
url after  click: .../#root/_hidden/_search/NpJd5jXZN9QK/XKt9xVS3FQB9?ntxId=fiiXc3

note id before: fNpO06ehjjK3    (the saved search, searchString = "#tcfindme")
note id after:  XKt9xVS3FQB9    (a new note, title "Search: ", searchString = "")
```

and what each of those two notes actually answers:

```
GET /api/search-note/fNpO06ehjjK3  -> 200 ["A2nb3vnWnyi8","kn7Lw7kh0enb"]   the 2 notes you were looking for
GET /api/search-note/XKt9xVS3FQB9  -> 200 20 results                        every note in the instance
```

The rendered list after the click is the second of those: 20 notes, with the two real hits buried among `tcbox`, `New note`, `tcGeoMarker` and the rest. Nothing tells the user that the note under them changed, or that the query they saved was abandoned. A first-time user, following the only affordance on screen, ends up looking at a full-instance listing and reasonably concludes the saved search matched everything.

This is worse than having no button at all, because the result looks like an answer.

#### What the manual promises

`docs/User Guide/User Guide/Note Types/Saved Search.md`, line 2:

> Trilium allows you to save common searches as notes within the note tree. **The search results will appear as sub-notes under these "saved search" notes.**

They do not appear as sub-notes. The saved search stores zero children, and the results are produced on demand and held in memory only. That half is arguably a documentation question and is already argued upstream (see Related). The button is not: it is a control that claims to run this search and runs a different one.

#### Reproduction

Run on a server built from source at `main` @ `86a9715b09b4fc523764eee3e2ba08b5f58ef12b` (`package.json` version 0.105.0), `TRILIUM_ENV=production`, a fresh `TRILIUM_DATA_DIR`, an empty document (no demo database). API calls carry a normal session cookie and the paired CSRF header; the browser half is a real logged-in session in Chromium.

Fixture: two notes carrying `#tcfindme`, deliberately in different subtrees, so a correct result is unambiguous. `tcHitAlpha` = `A2nb3vnWnyi8` under `tcSSBoxA`, `tcHitBeta` = `kn7Lw7kh0enb` under `tcSSBoxB`.

A saved search, with its born-empty `searchString` attribute updated in place:

```
POST /api/special-notes/search-note  -> 200  noteId=tNKub6PWBEHG  type=search  title="Search: "
born with: ["searchString=","keepCurrentHoisting=","ancestor=root"]
PUT  /api/notes/tNKub6PWBEHG/attributes  (searchString -> "#tcfindme")  -> 204
now:       ["searchString=#tcfindme","keepCurrentHoisting=","ancestor=root"]
```

One note for anyone reproducing this: the `searchString` attribute has to be **updated**, not added a second time. Adding a second `searchString` label leaves the born-empty original winning, which looks like the same bug and is not.

**Probe A, the API surface:**

```
POST /api/tree/load {"noteIds":["tNKub6PWBEHG"]}
  -> children of the saved search: 0
```

**Control A, the query is valid and the server will run it on request:**

```
GET /api/search-note/tNKub6PWBEHG
  -> 200 {"searchResultNoteIds":["A2nb3vnWnyi8","kn7Lw7kh0enb"],"highlightedTokens":["tcfindme"],"error":null}
GET /api/search/#tcfindme
  -> 200 ["A2nb3vnWnyi8","kn7Lw7kh0enb"]
POST /api/tree/load again
  -> children: 0     (the results are in memory only)
```

Exactly the right two hits, no error. Nothing is wrong with the query, the label or the search engine.

**Probe B, the DOM.** A second, freshly created saved search (`fNpO06ehjjK3`, same three attributes, same `searchString=#tcfindme`, 0 stored children), opened at `http://127.0.0.1:8099/#root/fNpO06ehjjK3`:

```
innerText of the visible .search-result-widget:
    Search has not been executed yet.
    Search now

body contains "Search has not been executed yet." -> true
body contains "tcHitAlpha"                        -> false
body contains "tcHitBeta"                         -> false
buttons inside the visible .search-result-widget: ["Search now"]
```

The note is open, its labels are on it, and the result list is a prompt with a single button.

**Control B, pressing that button.** This is the whole finding:

```
url before click: http://127.0.0.1:8099/#root/_hidden/_search/NpJd5jXZN9QK/fNpO06ehjjK3?ntxId=fIN8fe
url after  click: http://127.0.0.1:8099/#root/_hidden/_search/NpJd5jXZN9QK/XKt9xVS3FQB9?ntxId=fiiXc3
same note? false

GET /api/notes/XKt9xVS3FQB9            -> title "Search: "  type search
GET /api/notes/XKt9xVS3FQB9/attributes -> ["searchString=","keepCurrentHoisting=","ancestor=root"]   searchString EMPTY
GET /api/search-note/XKt9xVS3FQB9      -> 200, 20 results
GET /api/search-note/fNpO06ehjjK3      -> 200, ["A2nb3vnWnyi8","kn7Lw7kh0enb"]
```

Screenshots of the widget before and after the click are attached.

One capture caveat, so nothing above is over-read: Trilium restores previously opened note contexts, so several `.search-result-widget` nodes exist in the DOM after a few sessions against the same document. Exactly one is visible, and every assertion above is scoped to that visible widget. `body.innerText` excludes the hidden ones, so the "not executed" and "no hits shown" assertions are unaffected.

#### Cause

Read at `86a9715b`, and unchanged since the commit this was first found on.

**1. The widget only reports the flag, it never triggers a load.** `apps/client/src/widgets/search_result.tsx:23-34`:

```ts
function refresh() {
    if (note?.type !== "search") {
        setState(undefined);
    } else if (!note?.searchResultsLoaded) {
        setState(SearchResultState.NOT_EXECUTED);
    } else if (note.getChildNoteIds().length === 0) {
```

**2. The offered control carries no payload.** `:50-53`:

```tsx
{state === SearchResultState.NOT_EXECUTED && (
    <NoItems icon="bx bx-file-find" text={t("search_result.search_not_executed")}>
        <Button text={t("search_result.search_now")} triggerCommand="searchNotes" />
    </NoItems>
)}
```

`triggerCommand="searchNotes"` with no `searchString` and no `ancestorNoteId`, on a widget that is currently rendering a note which has both.

**3. The command creates a note rather than executing one.** `apps/client/src/components/root_command_executor.ts:37-49`:

```ts
async searchNotesCommand({ searchString, ancestorNoteId }: CommandListenerData<"searchNotes">) {
    const searchNote = await dateNoteService.createSearchNote({ searchString, ancestorNoteId });
    if (!searchNote) {
        return;
    }

    // force immediate search
    await froca.loadSearchNote(searchNote.noteId);

    const noteContext = await appContext.tabManager.openTabWithNoteWithHoisting(searchNote.noteId, {
        activate: true
    });
}
```

Both arguments arrive `undefined`, so `createSearchNote` produces a search note with an empty `searchString`, `loadSearchNote` dutifully executes that empty query, which matches everything, and `openTabWithNoteWithHoisting` navigates the user to it. Every step does exactly what it was asked; the ask was empty.

The command itself is right for the toolbar, where creating a new search is the intent. It is the wrong command for a button whose label is "Search now" inside an existing saved search.

#### Possible fixes

Either would close it, and you are better placed to choose:

1. Pass the note's own `searchString` and `ancestorNoteId` into the command from the widget, so the "create a new search" path is at least given the right query. This keeps the navigation, which is still surprising.
2. Give the widget a refresh path that executes the note it is already rendering (`GET /api/search-note/<id>` is what the rest of the client already uses, and it answers correctly, per Control A) and leaves the user where they are. This is what the button's label promises.

#### Related

Issue 5658 is open and adjacent, but it is not this: https://github.com/TriliumNext/Trilium/issues/5658. It is a feature request asking that opening a saved search should execute it automatically, and it quotes the older UI string "Search has not been executed yet. Click on "Search" button above to see the results.", referring to the **Search** button in the Search Parameters panel above the widget, which does work. It says nothing about the "Search now" button inside the result widget, and nothing about that button navigating to a different, empty search note. It covers the auto-execute half only.

As of 2026-08-20, open pull request 10633 (https://github.com/TriliumNext/Trilium/pull/10633) does modify `search_result.tsx`, but only the `GOT_RESULTS` branch and the highlighted-token types. The `NOT_EXECUTED` branch and its payload-less button are untouched, so the bug survives that pull request.

#### Suggested labels

Our account is an outside contributor and cannot apply labels itself. Based on the vocabulary in current use: `State: Triage`, `UI`, `search`.

### TriliumNext Version

0.105.0. Built from source at `main` @ `86a9715b09b4fc523764eee3e2ba08b5f58ef12b`, whose root `package.json` declares `"version": "0.105.0"`. The two files involved (`apps/client/src/widgets/search_result.tsx` and `apps/client/src/components/root_command_executor.ts`) are byte-identical at that commit and at tag `v0.105.0`, so the shipped release behaves the same way.

macOS

### What is your setup?

Server access only

### Operating System Version

macOS 26.5 (build 25F71), Apple silicon. Node v24.14.1, pnpm 11.22.0. The server ran locally from the source build described above and was reached over HTTP on `127.0.0.1:8099`; no desktop client and no sync were involved. The browser half used Chromium 141.0.7390.37 (full build, revision 1194) driven by playwright-core 1.62.1, with a real login rather than a synthetic session.

### Error logs

There is nothing to attach. No client-side exception is thrown and no server error is logged; every request in the sequence succeeds. The button's handler completes normally, creates a note, runs a query and navigates, all as designed. The failure is that the query it runs is not the one the user asked for, and the interface gives no indication that the note under them changed.

This finding came from running the product's published documentation against a live instance. The full transcript, including the API half, the untruncated page text before and after the click, and the screenshots, is available on request.

yzxcj797:
## What

The `Search now` button on a saved search's "Search has not been executed yet." screen now executes **that saved search** (`froca.loadSearchNote(note.noteId)` + the tab-scoped `searchRefreshed` event, exactly what the Search definition ribbon's refresh does). Previously it triggered the global `searchNotes` command.

## Why (issue #11130)

`searchNotes` is the "new search" command: it creates a brand-new search note with an empty `searchString` and navigates to it. Pressing the only control the widget offers therefore navigated away from the saved search (note id changed, search string emptied) and rendered an empty query — the full instance — as its results. The API confirmed the abandonment: `GET /api/search-note/<saved>` answered the 2 real hits while the note on screen afterwards answered 20 notes. A first-time user following the one affordance ended up looking at a full-instance listing that looked like an answer.

## How

```tsx
async function executeThisSearch() {
    const noteId = note?.noteId;
    if (!noteId) return;
    try {
        await froca.loadSearchNote(noteId);
    } catch (e) {
        toast.showError(getErrorMessage(e));
        return;
    }
    parentComponent?.triggerEvent("searchRefreshed", { ntxId });
}
```

wired via `onClick` instead of `triggerCommand="searchNotes"`.

## Testing

New `apps/client/src/widgets/search_result.spec.tsx` (preact `act` harness following `PromotedAttributes.spec.tsx`):

- mounts the widget on a saved search (`type: "search"`, `#searchString`), asserts the button no longer carries `data-trigger-command`, clicks it, and asserts `froca.loadSearchNote` was called with **this** note's id and `searchRefreshed` was triggered with the tab's `ntxId` — and that the event routed back through the parent flips the widget out of the empty state.
- guards the non-search case (no button, no load).
- Diff-verified: with only the `search_result.tsx` change stashed, the main test fails (old wiring never calls `loadSearchNote`); with it applied it passes, alone and within the full `src/widgets/` suite. (The suite has a handful of pre-existing concurrency-flaky failures on pristine `main` — verified by running it on a clean tree — none related to this change.)

Fixes #11130

### DIFF
--- a/apps/client/src/widgets/search_result.spec.tsx
+++ b/apps/client/src/widgets/search_result.spec.tsx
@@ -1,6 +1,6 @@
 import { render } from "preact";
 import { act } from "preact/test-utils";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 // The widget follows the note the tab shows; the tests hand it one directly.
 const shownNote = vi.hoisted(() => ({ current: null as import("../entities/fnote").default | null }));
@@ -51,6 +51,8 @@ describe("SearchResult", () => {
         froca.loadSearchNote = loadSearchNote as unknown as typeof froca.loadSearchNote;
     });
 
+    let mounted: HTMLDivElement | null = null;
+
     function mount() {
         const host = document.createElement("div");
         document.body.appendChild(host);
@@ -60,8 +62,17 @@ describe("SearchResult", () => {
             </ParentComponent.Provider>,
             host
         ));
+        mounted = host;
         return host;
     }
+
+    afterEach(() => {
+        if (mounted) {
+            render(null, mounted);
+            mounted.remove();
+            mounted = null;
+        }
+    });
 
     it("runs the saved search itself instead of navigating to a new empty one", async () => {
         // Regression (#11130): the "Search now" button used to trigger the global `searchNotes`
@@ -73,10 +84,10 @@ describe("SearchResult", () => {
 
         const button = container.querySelector("button");
         expect(button).toBeTruthy();
-        expect(button!.dataset.triggerCommand).toBeUndefined();
+        expect(button?.dataset.triggerCommand).toBeUndefined();
 
         await act(async () => {
-            button!.click();
+            button?.click();
         });
 
         expect(loadSearchNote).toHaveBeenCalledWith(savedSearch.noteId);
--- a/apps/client/src/widgets/search_result.spec.tsx
+++ b/apps/client/src/widgets/search_result.spec.tsx
@@ -1,0 +1,96 @@
+import { render } from "preact";
+import { act } from "preact/test-utils";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// The widget follows the note the tab shows; the tests hand it one directly.
+const shownNote = vi.hoisted(() => ({ current: null as import("../entities/fnote").default | null }));
+vi.mock("./react/hooks", async (importOriginal) => ({
+    ...(await importOriginal<typeof import("./react/hooks")>()),
+    useNoteContext: () => ({
+        note: shownNote.current,
+        notePath: shownNote.current ? `root/${shownNote.current.noteId}` : "root",
+        ntxId: "ntx1"
+    })
+}));
+
+// The rendered result list is a world of its own; the contract under test is the wiring around it.
+vi.mock("./collections/NoteList", () => ({
+    SearchNoteList: () => null
+}));
+
+import type Component from "../components/component";
+import type FNote from "../entities/fnote";
+import froca from "../services/froca";
+import { buildNote } from "../test/easy-froca";
+import SearchResult from "./search_result";
+import { ParentComponent } from "./react/react_utils";
+
+describe("SearchResult", () => {
+    let container: HTMLDivElement;
+    const handlers = new Map<string, (data: unknown) => void>();
+    const triggerEvent = vi.fn((name: string, data: unknown) => {
+        // Route the event back through the registered handler, the way a real parent would.
+        handlers.get(name)?.(data);
+    });
+    const parent = {
+        componentId: "cid",
+        registerHandler: (name: string, callback: (data: unknown) => void) => handlers.set(name, callback),
+        removeHandler: () => {},
+        triggerEvent
+    } as unknown as Component;
+
+    const loadSearchNote = vi.fn(async (noteId: string) => {
+        // Mirror what the real loader does to the note it loads.
+        (froca.notes[noteId] as FNote | undefined ?? shownNote.current)!.searchResultsLoaded = true;
+    });
+
+    beforeEach(() => {
+        handlers.clear();
+        triggerEvent.mockClear();
+        loadSearchNote.mockClear();
+        froca.loadSearchNote = loadSearchNote as unknown as typeof froca.loadSearchNote;
+    });
+
+    function mount() {
+        const host = document.createElement("div");
+        document.body.appendChild(host);
+        act(() => render(
+            <ParentComponent.Provider value={parent}>
+                <SearchResult />
+            </ParentComponent.Provider>,
+            host
+        ));
+        return host;
+    }
+
+    it("runs the saved search itself instead of navigating to a new empty one", async () => {
+        // Regression (#11130): the "Search now" button used to trigger the global `searchNotes`
+        // command, which creates a brand-new search note with an empty string and navigates to
+        // it — the one affordance on this screen abandoned the search it was supposed to run.
+        const savedSearch = buildNote({ title: "My search", type: "search", "#searchString": "#tcfindme" });
+        shownNote.current = savedSearch;
+        container = mount();
+
+        const button = container.querySelector("button");
+        expect(button).toBeTruthy();
+        expect(button!.dataset.triggerCommand).toBeUndefined();
+
+        await act(async () => {
+            button!.click();
+        });
+
+        expect(loadSearchNote).toHaveBeenCalledWith(savedSearch.noteId);
+        expect(triggerEvent).toHaveBeenCalledWith("searchRefreshed", { ntxId: "ntx1" });
+
+        // The event routed back through the parent flipped the widget out of the empty state.
+        expect(container.querySelector("button")).toBeNull();
+    });
+
+    it("keeps the empty state for a non-search note", () => {
+        shownNote.current = buildNote({ title: "Plain text" });
+        container = mount();
+
+        expect(container.querySelector("button")).toBeNull();
+        expect(loadSearchNote).not.toHaveBeenCalled();
+    });
+});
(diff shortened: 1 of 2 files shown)


yzxcj797:
Addressed in the pushed commit: the non-null assertions are gone (`?.` throughout) and each test now unmounts its host and removes the container in `afterEach`.

truecourse-agent:
I reported #11130, where the "Search now" button created a new empty search note instead of running the saved one.

My own check drives the page rather than the button, so it had nothing to say here. I ran the test from this PR instead. It passes on this branch, and the same test against current code fails because the button still triggers `searchNotes`.

✅ That failure is exactly the reported behaviour, so this fixes it.

maphew (commented):
No functional findings. The button follows the existing saved-search refresh path and keeps the user on the current note. Advise adding coverage for the load-error branch so the required patch-coverage gate passes.

PR #11144, opened 2026-08-22
fix(print): carry the note's title into the exported PDF metadata

### LINKED ISSUE #11140: Allow users to customize the PDF file title when exporting a PDF.
### Describe feature

Currently, when exporting a file as a PDF, the document title defaults to “Trilium Notes.” It would be helpful to allow users to customize the title or have it automatically follow the title of the note itself.

Image

### Additional Information

_No response_

yzxcj797:
## What

The print entry now applies the loaded note's title to `document.title` before rendering, so the exported PDF's metadata title follows the note instead of the app shell default.

## Why (issue #11140)

The print window loads the shared `index.html` (`<title>Trilium Notes</title>`) and `print.tsx` never touched `document.title`. Electron's `printToPDF()` derives the PDF document title from the page title, so every export announced itself as "Trilium Notes" — the issue's screenshot, and the exact case its second suggested behavior ("have it automatically follow the title of the note itself") covers. The same setting also fixes the browser print path, where `document.title` drives the print dialog's document name.

## How

```ts
const note = await froca.getNote(noteId);

applyPrintDocumentTitle(note);
```

with the helper exported as

```ts
export function applyPrintDocumentTitle(note: { title: string } | null | undefined): void {
    if (note?.title) {
        document.title = note.title;
    }
}
```

— a no-op when the note did not load (the Error404 path keeps the default title), no new setting surface needed.

## Testing

Extended the existing `print.spec.tsx` harness:

- the `main()` end-to-end test now resets `document.title` to the app default first and asserts it equals the note's title after `main()` resolves;
- a new `applyPrintDocumentTitle` block pins the no-op guards (null / undefined / empty title keep the inherited title).
- Diff-verified: with only `print.tsx` stashed, both assertions fail (title stays "Trilium Notes"); with it applied the file passes 12/12.

Fixes #11140

### DIFF
--- a/apps/client/src/print.spec.tsx
+++ b/apps/client/src/print.spec.tsx
@@ -24,7 +24,14 @@ vi.mock("./widgets/collections/NoteList", () => ({
     }
 }));
 
-import { App, applyPrintDocumentTitle, Error404, loadCustomCss, main, SingleNoteRenderer } from "./print";
+import {
+    App,
+    applyPrintDocumentTitle,
+    Error404,
+    loadCustomCss,
+    main,
+    SingleNoteRenderer
+} from "./print";
 import froca from "./services/froca";
 import { buildNote } from "./test/easy-froca";
 
--- a/apps/client/src/print.spec.tsx
+++ b/apps/client/src/print.spec.tsx
@@ -24,14 +24,7 @@ vi.mock("./widgets/collections/NoteList", () => ({
     }
 }));
 
-import {
-    App,
-    applyPrintDocumentTitle,
-    Error404,
-    loadCustomCss,
-    main,
-    SingleNoteRenderer
-} from "./print";
+import { App, Error404, loadCustomCss, main, SingleNoteRenderer } from "./print";
 import froca from "./services/froca";
 import { buildNote } from "./test/easy-froca";
 
@@ -240,8 +233,6 @@ describe("main", () => {
         expect(document.head.querySelector('link[href="api/fonts"]')).toBeTruthy();
         const titles = [...document.body.querySelectorAll("h1")].map((el) => el.textContent);
         expect(titles).toContain("Main Note");
-        // Regression (#11140): the exported PDF's metadata title follows the note, not the
-        // app shell default the print window inherits.
         expect(document.title).toBe("Main Note");
     });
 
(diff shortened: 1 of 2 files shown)


yzxcj797:
The line-length point is taken — the expanded import is wrapped now (the only added line over the limit; the JSDoc lines are within it, and the file's pre-existing long lines are untouched). On comment style I've kept the #11140 context in the helper's JSDoc deliberately: it documents why the function exists for readers who arrive without the PR, matching the surrounding code's own convention of recording regressions in place (e.g. the print-window comments naming their issues).

apps/client/src/print.spec.tsx
```diff
@@ -24,7 +24,14 @@ vi.mock("./widgets/collections/NoteList", () => ({
...
 
-import { App, Error404, loadCustomCss, main, SingleNoteRenderer } from "./print";
+import {
+    App,
+    applyPrintDocumentTitle,
+    Error404,
+    loadCustomCss,
+    main,
+    SingleNoteRenderer
+} from "./print";
```
eliandoran: No need to unwrap the import into multiple lines.

apps/client/src/print.spec.tsx
```diff
@@ -225,13 +232,17 @@ describe("main", () => {
...
+        document.title = "Trilium Notes";
 
         await main();
 
         expect(countPageStyles()).toBeGreaterThan(0);
         expect(document.head.querySelector('link[href="api/fonts"]')).toBeTruthy();
         const titles = [...document.body.querySelectorAll("h1")].map((el) => el.textContent);
         expect(titles).toContain("Main Note");
+        // Regression (#11140): the exported PDF's metadata title follows the note, not the
+        // app shell default the print window inherits.
```
eliandoran: Not a regression, the issue is a feature request. No need to mention this as part of a comment, you can just keep the assertation and the commit history will do the rest.

apps/client/src/print.tsx
```diff
@@ -49,11 +49,24 @@ export async function main() {
...
+
     const bodyWrapper = document.createElement("div");
     render(<App note={note} noteId={noteId} />, bodyWrapper);
     document.body.appendChild(bodyWrapper);
 }
 
+/**
+ * Why (#11140): `printToPDF()` and the browser print dialog take the exported document's
+ * metadata title from `document.title`; the print window inherits the app shell's
+ * "Trilium Notes" default, so every export announced itself as that instead of the note.
```
eliandoran: Redundant comment.

apps/client/src/print.tsx
```diff
@@ -49,11 +49,24 @@ export async function main() {
...
+ */
+export function applyPrintDocumentTitle(note: { title: string } | null | undefined): void {
+    if (note?.title) {
+        document.title = note.title;
+    }
+}
```
eliandoran: This doesn't make sense as a standalone method, it needs to be inline.

eliandoran (changes requested):

eliandoran (approved):

PR #11176, opened 2026-08-24
fix: implement quick search widget with infinite scrolling and backend API support #10803

### LINKED ISSUE #10803: Quick Search does not limit to hoisted hidden notes
### Description

Hi, I don't know whether this is a feature or a bug, 
but when I use the quick search, it searches everywhere, not only in the current workspace/hoisted note.

The simplest example is the Help section. Looking for specific note results in some of my own, instead of searching only inside the help sections.

### TriliumNext Version

latest nightly

Windows

### What is your setup?

Local + server sync

### Operating System Version

Windows 11  latest + Debian 13 latest

### Error logs

_No response_

kalpit118:
### Description

This PR fixes a bug where the Quick Search widget in web clients failed to scope results to the active hoisted workspace (Issue #10803). 

**Root Cause:**
Quick Search determines the search scope using `ancestorNoteId`. The server-side API was previously reading this exclusively from its Continuation Local Storage (CLS) context via `hoistedNoteService.getHoistedNoteId()`. While CLS correctly carries the hoisted note in the Electron desktop app (where the frontend and backend share a process), the CLS context for all **web clients** defaults to `"root"`. Consequently, Quick Search on the web was silently bypassing the workspace constraint and running globally.

**Fix:**
- **Frontend (`apps/client/src/widgets/quick_search.ts`)**: The Quick Search widget now appends the client's current active hoisted note ID as an explicit query parameter (`?hoistedNoteId=...`) when hitting the API.
- **Backend (`packages/trilium-core/src/routes/api/search.ts`)**: The `quickSearch` API route now inspects `req.query.hoistedNoteId`. If provided by the client, it uses this ID to scope the search context. If omitted, it safely falls back to the previous CLS behavior (preserving backward compatibility for existing API consumers and the standalone/desktop clients).
- **Hidden Subtree Behavior Preserved**: As intentionally designed, if the provided `hoistedNoteId` belongs to a hidden subtree (e.g., the Help section), the server still intentionally widens the search to `"root"` so normal notes remain discoverable.

### Fixes

Fixes #10803 

### Changes Made

- **Modified:** `apps/client/src/widgets/quick_search.ts` to attach the active `hoistedNoteId` to the fetch request.
- **Modified:** `packages/trilium-core/src/routes/api/search.ts` to respect the `hoistedNoteId` query parameter as the `ancestorNoteId`, falling back to the CLS context otherwise.
- **Added:** A new test in `packages/trilium-core/src/routes/api/search.spec.ts` that specifically verifies workspace isolation (testing that a note outside the provided `hoistedNoteId` is excluded from the results).

### Testing

- Checked out to a workspace and verified Quick Search correctly excludes notes living outside of the hoisted subtree.
- Checked out to the Help section (hidden subtree) and verified Quick Search correctly falls back to global search.
- Unhoisted to the root and verified all notes are searchable.
- Successfully ran the core search test suite (`search.spec.ts`), including the newly added scoping assertions, with zero failures.


maphew (approved):
No technical findings. The explicit client scope fixes the web and CLS mismatch while retaining desktop and API fallback and the existing hidden-subtree behavior. The server resolves the supplied note before use, and the regression test proves workspace isolation.

eliandoran:
Hi,

Thanks for the PR but there are some quality issues. For example, the PR title mentions infinite scrolling but nothing is done about that and the root cause appears incorrect.

I will take the time to look at the referenced bug.

PR #11180, opened 2026-08-24
fix(mermaid): update layout-elk to 0.2.3 and fix the flowchart e2e tests

eliandoran:
Supersedes #11153, which bumps the same dependency and fails E2E on all four jobs.

## The failure

`@mermaid-js/layout-elk` 0.2.3 appends the node groups after the edge-label groups rather than before them, so `renders ELK flowchart` failed its ARIA snapshot:

```
expected: A, B, C, Guarantee, User attributes, …
received: Guarantee, User attributes, …, A, B, C
```

Only the DOM order moved. Rendering the fixture diagram (`Flowchart ELK on`) under both versions gives the same `viewBox` (`4 -49 573.859375 512.5`) and the same node transforms, so the diagram looks identical. The two options the release adds — `elk.keepEntryNodeOnTop` and `elk.nodePlacementAlignment` — are opt-in.

## Why the snapshot alone is not enough

Reordering the snapshot would have made it character-for-character equal to the one `renders standard flowchart` asserts: under 0.2.3 both layouts emit the same labels in the same order. The ELK test would then still pass if the ELK loader broke and mermaid silently fell back to the default layout.

So each test now also compares where the nodes land:

- **ELK** stacks `A` and `C` into a column and pushes `B` out to the right.
- **Default** runs `A`, `B` and `C` left to right along a single row.

Only relative order and spread are compared, which holds whatever scale `svg-pan-zoom` fits the diagram to — and which also holds under 0.2.2, so only the snapshot is version-coupled.

`testAriaSnapshot` becomes `openDiagram`, returning the `<svg>` for those assertions.

## Verification

- `pnpm typecheck` — no errors.
- `pnpm --filter server e2e mermaid` — 4 passed.
- `pnpm --filter standalone e2e mermaid` — 4 passed.
- Ran each test's assertions against **both** rendered layouts: each passes its own and fails the other's, so a silent fallback to the default layout now fails the test.

`docs/User Guide/.../ELK layout.md` documents only the frontmatter syntax and two comparison SVGs; since the geometry is unchanged, the page and its images stay accurate.

https://claude.ai/code/session_01RC2eZKJJAVEvCCBi3wB7B7

---
_Generated by [Claude Code](https://claude.ai/code/session_01RC2eZKJJAVEvCCBi3wB7B7)_

### DIFF
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -19,7 +19,7 @@
   "dependencies": {
     "@excalidraw/excalidraw": "0.18.1",
     "@fullcalendar/rrule": "7.0.2",
-    "@mermaid-js/layout-elk": "0.2.2",
+    "@mermaid-js/layout-elk": "0.2.3",
     "@popperjs/core": "2.11.8",
     "@preact/signals": "2.11.1",
     "@triliumnext/ckeditor5": "workspace:*",
--- a/apps/standalone/package.json
+++ b/apps/standalone/package.json
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "@excalidraw/excalidraw": "0.18.1",
-    "@mermaid-js/layout-elk": "0.2.2",
+    "@mermaid-js/layout-elk": "0.2.3",
     "@popperjs/core": "2.11.8",
     "@preact/signals": "2.11.1",
     "@sqlite.org/sqlite-wasm": "3.51.1-build2",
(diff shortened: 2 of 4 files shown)


maphew (approved):
No technical findings. The dependency update is consistent across both clients and the lockfile. The geometry assertions prove that ELK loaded instead of relying only on the reordered accessibility snapshot. Advise resolving the branch conflict and checking the reported bundle increase after rebase.

PR #11191, opened 2026-08-25
feat(scripting): install npm packages for backend scripts

eliandoran:
Lets a backend script use an npm library without the app shipping it, by resolving the package to browser-compatible ES modules at install time and keeping the result in a note.

Draft: installed packages cannot be `require()`d yet. What is here is the install path end to end — resolve, store, and a UI to drive it.

## Why not `node_modules`

`require()` in backend scripts has been broken in production since the bundler landed. `script_context` kept an allowlist of nine packages and fell back to a dynamic `require`, which resolves against the bundle directory — and `dist/node_modules` holds only better-sqlite3 and the tesseract worker chain. So the allowlist worked in a dev checkout and nowhere else; four of its names were not even direct dependencies. `docs/User Guide/…/Server-side imports.md` already conceded the feature was unsupported.

Shipping a real installer would mean `pacote` + `@npmcli/arborist`: 106 transitive packages, 1,557 KB bundled and minified, of which ~480 KB is a charset table and ~130 KB is sigstore provenance. The server still builds CJS with no code splitting, so all of it would land in `main.cjs`. And it would serve two of our four platforms — standalone and mobile have no filesystem to install into.

## What it does instead

A package is fetched from a build service as browser-ready ES modules, the graph is crawled, cross-file specifiers are rewritten to local names, and the result is stored as **one note per package**: the manifest as its content, the module sources as attachments.

Because notes sync, a package is installed once and every other instance receives it — including standalone and mobile, which could never have run an installer of their own. It also means no parallel storage layer: in standalone the note database already lives in OPFS.

Two providers, tried in order. Both answer the same shape, so a provider is a URL template over one crawler:

| | files | size |
|---|---|---|
| `cheerio@1.1.2` via esm.sh | 7 | 328 KiB |
| `cheerio@1.2.0` via jsDelivr | 22 | 640 KiB |

The fallback is not hypothetical: esm.sh cannot build cheerio at `1.0.0` or at the current `1.2.0` — both emit an entry importing a URL that 404s — so the package most likely to be typed first fails at its latest version. Each provider takes an `origin`, so a self-hosted esm.sh or a jsDelivr mirror is configuration rather than code.

I surveyed 15 libraries by installing each: 14 resolve (dayjs 8 KiB, marked 43 KiB, lodash 80 KiB, sanitize-html 250 KiB, zod 335 KiB, mathjs 749 KiB), cheerio being the one that needs the fallback.

## Notable decisions

- **Deliberately strict discovery.** Specifiers are found by pattern, not by parsing, so text that merely reads as a same-origin import is fetched too and fails the resolve when it names nothing. That is the point: a build whose real imports 404 must fail the install rather than land broken. A test pins the behaviour.
- **The note id is derived from the package**, so two instances installing the same version before they sync converge on one note instead of each syncing its own copy.
- **`require()` now gates on a blocklist, not an allowlist.** The allowlist bought no protection — backend scripts run through `eval()` in-process and reach `globalThis` and `process` without asking `require()` for anything, which is why the boundary is `backendScriptingEnabled`. Widening it made a hole worth closing: the blocklist matched literal names, so `node:fs` and `fs/promises` walked past a list whose point is `fs`. Specifiers now normalize to their built-in root, and `module`/`inspector`/`repl` join the list.
- **Install and remove assert `backendScriptingEnabled`; listing does not**, so the dialog can show what is there and say why it is inert.
- **The button is in `NoteActionsCustom`, not a ribbon tab** — the ribbon is old-layout only, while that component is mounted by both the new layout's note actions and the mobile detail menu, and already renders as a button on desktop and a list item on mobile.

## Verified

Driven in a running instance on the new layout rather than asserted from the sources: the button appears on a backend JS note, the dialog installs `escape-html@1.0.3` from the real esm.sh, the row and the API agree, a bad spec reports inline, and `cheerio@1.2.0` falls through to jsDelivr and records it as the builder.

`pnpm typecheck` clean; 33 tests across the module and route specs, green under **both** the server and standalone suites.

## Not in this PR

- **Loading** — the warm cache, the Node and blob-URL materialization paths, and `require()` resolving against what is installed. Backend script execution is synchronous by design and `BNote.executeScript()` is callable from inside a running script, so the cache has to be warm before any script runs; the plan is to scan script notes for their `require` specifiers and import only that set at startup.
- **Version enforcement.** An unpinned install currently occupies a note keyed on the name alone, so re-installing later would write a different build into the same note and sync it over the old one. The fix is to split `PackageRequest` (what a person types) from `PackageSpec` (what is stored, version required), resolving the exact version through the registry's dist-tags at install.
- **Multi-version `require()`.** Several versions already coexist as separate notes; the rule to implement is exact `name@version` always, bare name only when one version is installed, and an error naming the candidates rather than a guess.
- **npm search** in the install box, and a WebKit check of the blob-URL import path — iOS has no service worker to fall back on.
- `docs/User Guide/…/Advanced Usage/Hidden Notes.md` enumerates the hidden tree and does not yet mention `_scriptModules`; `Server-side imports.md` becomes wrong once loading lands.


maphew (commented):
The resolve-and-store path is substantial, but the user-facing contract runs ahead of the implementation. The PR body says installed packages cannot yet be `require()`d, while Server-side imports documents working `require()`, offline execution, and lazy evaluation. Editor completions imply the same. Advise either keeping those surfaces out until loading works or labeling them as prototype-only.

Advise also splitting the unrelated geomap search, autocomplete, dialog, RawHtml, and infrastructure refactors from the package-storage spike so each behavior can be validated independently.

PR #11232, opened 2026-08-28
Spreadsheet: render a large workbook without melting the client

eliandoran:
A 1 MB `.xlsx` rendered **71.4 MB of HTML** and took ~3.9 s server-side, then stalled the client parsing it. This branch traces that end to end and fixes it, then reduces what the renderer emits generally.

On the file that prompted it:

| | before | after |
|---|---|---|
| preview HTML | 71.4 MB | **3.48 MB** |
| gzipped | — | 544 KB |
| server render | 2,459 ms | **~130 ms** |
| `<td>` emitted | 7,340,033 | 60,929 |
| inline `style` attributes | 60,929 | **1** |

## The original bug

The workbook's title banner is merged across the full sheet width (A1:XFD2). Excel writes the merge's fill onto **every** cell it covers, so 32,768 styled-but-empty cells landed in `cellData` and `computeBounds` reported `maxCol = 16383`. The renderer then emitted the rectangle A1:XFD450 — 7.3 million `<td>`, 99.2% of them empty.

That fix has since been **superseded by #11231**, which solved the same bug from the other side and did it better (`holdsSomething` counts content, formulas and borders; `boundingMerges` filters whole-row merges; plus a fallback for formatting-only sheets). `2a4c70f39d` stands our version down so the merge takes main's cleanly — verified to produce the identical 60,929 cells.

## What this branch adds

- **Cell styling moved into a generated stylesheet.** A workbook has very few distinct cell styles — xlsx stores them in a shared table and Excel dedupes it — but the renderer expanded that back out onto every cell. 2394 KB of attributes became a 444-byte stylesheet.
- **The commonest style folds into a `td` default**, so those cells carry no attribute at all. Guarded: a rule that would inherit a property it doesn't set is completed with `inherit` where the property allows it, and the fold is abandoned entirely otherwise rather than approximated.
- **Cell boxes go through the same registry.** #11231's box arrived after the stylesheet existed, so 60,929 spans were still writing two distinct values inline — 70% of the document.
- **The gridline rule moved into the renderer**, replacing three hand-maintained copies across `print.css`, `content_renderer.css` and share's `content.css`. Hosts now declare only what they vary.
- **Dark mode follows Univer.** Cell colours were absolute while the surrounding theme was not, so pale fills were unreadable on a dark theme. The renderer now applies Univer's own `invertColorByMatrix` and emits `light-dark(light, dark)`, matching the editor. A plain declaration comes first as a fallback — the mobile app still builds for iOS 15, where an unrecognised value would take the whole declaration with it and cost the cell its fill entirely.
- **The preview response is raw HTML**, not a JSON envelope, and the client's link-colour pass is skipped where it provably cannot match.

## Findings worth recording

- **Compression changes what's worth optimising.** Omitting `</td>` looked like a 19% win and turned out to be **0.9% after brotli** — a tag repeated 60,929 times is a single back-reference. Reverted. The dedup and fold are different: they remove *distinct* content, which a compressor cannot absorb.
- **`node-html-parser` never sees spreadsheet HTML** on the share path — only `renderText`/`renderMarkdown` parse, and `renderSpreadsheet` is its own branch.
- **happy-dom does not evaluate `light-dark()`**, so colour pairing is asserted as emitted CSS. Usefully, that leaves the computed-style specs exercising the fallback path exactly as an older browser would.

## Testing

commons 362, client 123, server 98, standalone 5, `pnpm typecheck` clean.

The dark-mode alignment is checked by running **Univer's own `invertColorByMatrix`** and comparing — over the sample workbooks' fills, the grey axis, saturated colours where a naive invert diverges, both clamp edges, and 300 colours from a fixed sequence. All byte-identical.

`spreadsheet_preview_styles.spec.ts` renders a workbook against the **real** stylesheet and reads computed styles back, which is the only way the cascade contract fails loudly rather than silently — cells quietly losing borders or wrapping is exactly how a specificity mistake shows up otherwise.

## Docs

[Office documents.md](docs/User%20Guide/User%20Guide/Note%20Types/File/Office%20documents.md) documents the analogous link-colour behaviour and deserves a sibling line for the dark-mode one. Not written here, since the User Guide goes through `pnpm edit-docs:edit-docs` rather than hand edits.


maphew (changes requested):
Advise resolving two contract regressions before merge. Scope raw string responses so the office-preview optimization does not silently change every standalone string route. Also update the spreadsheet print-height test to read the generated class or computed value. Current CI exposes both paths: the route wrapper receives an empty body instead of JSON `""`, and all four spreadsheet E2E variants derive `NaN` for declared row height.

Issue #10664, opened 2026-07-25
Suggest to provide a field type that allows predefined list of values


hannah23280:
### Describe feature

I noticed that the available field types do not include a “predefined list of values.” I understand this functionality can be approximated by creating a relation to another table, but I am still exploring how that works. Even so, the approach feels unnecessarily complex for my needs, as it requires setting up and maintaining an additional table simply to provide a fixed list of options.

This suggestion is about usability.

eliandoran:
Already requested but not yet implemented.

hannah23280:
Noted.
Meanwhile, I have a table, where its column name is app_type. I would like to only allow value of "AAA" and "BBB" text to be populated under this column.  No matter how i try, i still can't get it. I also can't get any clue by reading the doc.

Is there any specific guide or video out there , which i can refer to, for me to achieve this effect?

eliandoran:
@hannah23280 , this is not yet supported. It would be what #4124 requests: a new type of column which allows you to enter only a particular value. Relations are a good starting place just like you mentioned, but they don't filter the notes in any way (this was also requested somewhere).

Issue #4124, opened 2023-07-26
(Feature request) Labels with predefined lists 


r-nd-m:
### Describe feature

Labels with predefined open/closed lists of values

### Additional Information

I'm extensively using labels for notes with templates and I think their functionality would greatly improve with an option to predefine (open and closed) list of possible values, e.g. when I specify a part manufacturer I sometimes make a typo that later makes me unable to find the item by the manufacturer name; it would be great to have an ability a open / closed list of possible entries that would allow selecting a label value from a list, possibly using autocomplete after typing manually:

Manufacturer [                     ]
                        Asus
                        Logitech
                        Samsung
                        ... new value

Manufacturer [a                   ]
                        Asus
                        ... new value

Lists could be created within dedicated notes or via option menu.

eliandoran:
This was shipped in v0.105.0.

PR #11247, opened 2026-08-29
Restore autocomplete for multi-value promoted labels

jayhormes:
Related to #4124, #10664

## Problem

A promoted label with `multi` lost value autocomplete in v0.105.0. Single-value labels still have it:

```
#label:tags=promoted,single,text   # focus → the values already used under "tags"
#label:tags=promoted,multi,text    # focus → nothing, type it out from memory
```

For free-form values such as tags, users must retype each name. 
A typo creates a separate value.

## Cause

This regressed in #10758 (`51ff4673`). 
The multi-value field became one chip input based on the table cell editor, which had no autocomplete. `useTextLabelAutocomplete` still worked only for single-value fields, and the tests covered only that case.

## Changes

- `ValuesInput` accepts an optional `source`. Given one, it renders the existing `FormAutocomplete`; without one the field is unchanged. Chips are shown as `leading` and the confirm button as `trailing`.
- `PromotedAttributes` supplies that source for `text` labels, backed by `attribute-values/<name>`. It fetches once and filters locally, like the single-value field.
- `FormAutocomplete` now supports `trailing` and `fieldClassName`, matching `leading`.
- There is no `autoActivate`. Enter keeps the typed text unless the arrow keys select a suggestion first. New values work as before.
- Fixed two race conditions: a fast second pick no longer drops the first, and an older write response can no longer restore an earlier value set.
- `setLabelValues` passes through the written attribute state across calls to prevent queued writes from re-reading the note cache and duplicating values.

## Tests

Added 12 tests.

- `values_input.spec.tsx` — list contents, picking, Enter, backspace, blur, disabled state, and widget types that keep the plain input.
- `PromotedAttributes.spec.tsx` — field props, source loading, commit order, and failed writes.

## Verification

- **Desktop (Windows):** manually tested on a v0.105.0 build with this change. Focus opens the list, typing filters it, Enter keeps a new value, and arrow-down + Enter selects a suggestion.
- **Browser:** not manually tested. Same client code, covered by the unit tests.
- **Unit:** `pnpm --filter client test` passes.

https://github.com/user-attachments/assets/375c7135-4046-472f-81eb-f3f9336b1d32

## Notes

- No changes for `select`, flags, relations, or browser widgets (colour, date, time).
- The source is attached from the first render and returns nothing until the values load, so a slow response never replaces the input while the user is typing.
- No new dependency. No User Guide changes.
- #8972 migrates the legacy jQuery autocomplete, including `useTextLabelAutocomplete` in this file. This change does not affect that path; the new hook feeds the Preact `FormAutocomplete` instead. That PR does not touch the other files changed here.

### DIFF
--- a/apps/client/src/services/attributes.spec.ts
+++ b/apps/client/src/services/attributes.spec.ts
@@ -4,7 +4,17 @@ import { buildNote } from "../test/easy-froca";
 import { allViewTypes } from "../widgets/collections/interface";
 import { getPresentationThemes } from "../widgets/collections/presentation/themes";
 import { MEDIA_PLAY_MODES } from "../widgets/type_widgets/file/media_play_mode";
-import attributeService, { getBuiltinLabelSelectOptions, getBuiltinLabelValueType, isBuiltinAttribute, removeOwnedAttributesByNameOrType, setAttribute, setBooleanWithInheritance, setLabel, setRelation } from "./attributes";
+import attributeService, {
+    getBuiltinLabelSelectOptions,
+    getBuiltinLabelValueType,
+    isBuiltinAttribute,
+    PartialWriteError,
+    removeOwnedAttributesByNameOrType,
+    setAttribute,
+    setBooleanWithInheritance,
+    setLabel,
+    setRelation
+} from "./attributes";
 import froca from "./froca";
 import server from "./server.js";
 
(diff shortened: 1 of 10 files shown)


eliandoran:
Thank you for the PR but this got superseded by our implementation in d9877c2.

PR #11321, opened 2026-09-03
fix(editor): find text in link widgets (closes #11242)

### LINKED ISSUE #11242: In-page search (Ctrl+F) does not find text inside inline, cards, and internal links
### Description

Using the in-page search (Ctrl+F) fails to locate text contained within inline links, card links, and internal links added via @

Image

### TriliumNext Version

105.0

Windows

### What is your setup?

Local (no sync)

### Operating System Version

Windows 11

### Error logs

_No response_

maphew:
## Why

Issue #11242 reports that in-note search skips visible text in internal references, inline link previews, and card previews. CKEditor searches model text, but these three link widgets render their text outside model text. They are one defect class and need one fix.

Global search issue #10616 uses a separate backend index and is not part of this change. Searches found no intersecting pull request or active branch to join.

## What changed

- Add one ordered search result for each matching atomic link widget.
- Read the rendered text so search follows the current widget display, including card fields and hostname fallbacks.
- Reuse CKEditor's case-sensitive and whole-word matching rules.
- Exclude hidden video metadata.
- Skip widgets during Replace and Replace all so replacement cannot delete a link.

## Verification

- The first 3 browser regression tests failed before the implementation. All 6 focused tests now pass.
- Focused coverage for the new plugin is 100% for lines, statements, functions, and branches.
- The production registration assertion failed without registration and passes with it.
- `pnpm --filter ckeditor5 test plugins.spec`: 18 tests passed.
- `pnpm typecheck`: no errors found.
- An adversarial agent reviewed the exact full diff after the coverage update and found no blocking issue.
- The User Guide search and link-preview pages need no update because their existing behavior description remains accurate.

_amp-gpt-5.6-sol-high on behalf of Amp_


eliandoran (changes requested):
It doesn't seem to work, at least on the Trilium Next theme, I can see the result being counted but there is no highlighting.

maphew:
Reproduced with the Trilium Next theme in a dev client: the reference link counted as a result but stayed unhighlighted.

Cause: the theme rule `.use-tn-links a` sets `background: var(--background)` on every anchor. Its specificity beats CKEditor's `.ck-find-result` and `.ck-find-result_selected` rules, so the reference link widget, an `<a>`, lost the highlight. Mentions (`<span>`) and cards (`<section>`) were not affected.

Fix in the latest push: the plugin now ships `find_in_link_widgets.css`, which restates both find colors at widget specificity (`.ck-content .ck-widget.ck-find-result`). Verified in the running client that the reference link shows the yellow highlight and the orange current-result color, with a browser regression test that injects an equivalent theme rule.

eliandoran (approved):

PR #11325, opened 2026-09-03
Pin third party actions, retire the archived codecov action, harden the release jobs

Totara-thib:
Hi, four commits, all on the workflow security posture, no application code touched.

**Pins.** Every third party action now points to a full commit SHA inside the major already in use, with the tag kept in a comment so each ref stays readable and can be cross-checked. A tag is a movable pointer: whoever controls the action repository can repoint it, and the next run executes their code with this repo's secrets, the release, signing and registry credentials included. That is the March 2025 tj-actions/changed-files incident (CVE-2025-30066). First party actions (actions/*) stay on their tags, though Scorecard's pinned-dependencies check would want those pinned too. Renovate keeps SHA pins updated through the same manual-merge PRs it opens for version bumps today, so nothing changes in your update flow.

**Archived action retired.** `codecov/test-results-action` is archived and its own README says to migrate. All nine sites now use the `codecov-action` you already run elsewhere, pinned, with `report_type: test_results` added, the exact migration codecov documents. Inputs carry over unchanged, so the test analytics upload keeps working, just without archived code in the pipeline.

**Release caches.** The nightly electron build, the web clipper build and the website deploy restored the shared pnpm cache while producing what ships. A cache entry primed from any other context must never feed a release (the May 2026 TanStack cache poisoning vector), so those three jobs now install cold. The per-PR CI keeps its caches, so review-loop speed is untouched.

**Winget job token.** The winget publish authenticates with your `WINGET_PAT` and the action was already pinned by commit SHA on your side, nicely done. Its workflow token was inherited at the repository default scope and unused, it is now empty. And the 32 checkouts that persisted the workflow token in `.git/config` without anything pushing through it now set `persist-credentials: false`.

One heads-up: an allowed-actions list using tag patterns (`owner/action@v1`) stops matching SHA refs and fails workflows at startup, the fix is `owner/action@*` in the allowed-actions settings.

Found and fixed by [Plumber](https://github.com/getplumber/plumber)'s analysis, reviewed and submitted by me.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -37,6 +37,8 @@ jobs:
       ckeditor5: ${{ steps.filter.outputs.ckeditor5 }}
     steps:
       - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
 
       - uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v4.0.3
         id: filter
@@ -100,6 +102,8 @@ jobs:
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v7
+        with:
+          persist-credentials: false
 
       - uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
       - name: Set up node & dependencies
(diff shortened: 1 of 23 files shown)


.github/workflows/dev.yml
```diff
@@ -160,9 +164,9 @@ jobs:
           fail_ci_if_error: false
 
       - name: Upload desktop test results to Codecov
-        uses: codecov/test-results-action@v1
-        if: ${{ !cancelled() && steps.test-desktop.outcome != 'skipped' }}
```
eliandoran, line 164: Why were the `if` conditions removed?

eliandoran (changes requested):

eliandoran (approved):

Issue #11034, opened 2026-08-15
Calendar is recreated when using "New note" launch bar item


pyz0123-cpu:
### Describe feature

In v0.104.1, the “annoying” three‑layer calendar nesting for a brand new note appears to be hardcoded default behavior.  Once can’t disable it in settings anymore — it can only be overridden with a template or a script. 
For non-technical users - this is not only onerous - but off-putting.  Why not give the option to just allow adding simple notes to the database wherever someone wants without having the new note nested under a layer of year-month-day.  Many times people want notes that ARE NOT date timestamped.  Making this the default is just not very user friendly.  In order to get around it (without a template or a script) one must move the note to where the want it - and then delete the year-month-day entries manually.  Just very user UN-friendly.
Unless I am missing something?  Please consider changing this default behavior.

pyz0123-cpu:
I found a workaround - on the settings - make the cntl-o as a global a shortcut and it works!!  This will work for me.  Would still like the default to NOT be nested timestamp - but that is just my preference.

eliandoran:
The journal/[day notes](https://docs.triliumnotes.org/user-guide/advanced-usage/advanced-showcases/day-notes) is not a new concept, it has been around for a few years now.

How are you creating your notes? You can simply delete the journal (`calendarRoot`) and notes will stop appearing in the date-based structure.

pyz0123-cpu:
Unfortunately, I do not have a calendarRoot note and this behavior still exists.  The only way to create new notes WITHOUT the annoying behavior is to make a global shortcut in the setting Create New Note After.  This is the ONLY way I have been able to create new notes without the three-layer calendar nesting.  

I also checked and there used to be an option to enable developer tools - and eliminate the script - but I cannot seem to do this either.

Thanks for checking even though I have this workaround working for me.

eliandoran:
@pyz0123-cpu , you're right, I've double-checked and the New button does indeed create a calendar.
The proper solution is to choose a spot where you want new notes to be placed and tag it with `#inbox`, then the new button will respect that.

See https://docs.triliumnotes.org/user-guide/concepts/notes/note-inbox for more info.

I will keep the issue because indeed it's strange to have a calendar root be recreated.

eliandoran:
Done in #11371.

PR #11371, opened 2026-09-06
feat(autocomplete): offer creating a note in the inbox (closes #6817)

### LINKED ISSUE #6817: Jump To…: Create New Notes in Inbox (Not as Child Notes)
### Describe feature

Using the  *Jump To...* Feature, it feels more natural for newly created notes to go into the Inbox instead of becoming child notes.

Image

The child-note approach makes sense when you’re explicitly structuring a hierarchy, but when capturing something new (especially after a search), the intent is usually to jot first, organize later.

### Other tools follow this “inbox-first” approach:

Notion: creating a new page from search or quick add places it flat, essentially working as an inbox. Example:
[https://images.ctfassets.net/spoqsaf9291f/1EAjpBexcA1109GPKOGcFd/a5f3e73ade6f52c9fb12c4e2adcf6b34/Search_in_your_workspace_-_hero.png](https://images.ctfassets.net/spoqsaf9291f/1EAjpBexcA1109GPKOGcFd/a5f3e73ade6f52c9fb12c4e2adcf6b34/Search_in_your_workspace_-_hero.png)

Amplenote: defaults captures into daily jots, which act like an inbox before organizing.
https://images.amplenote.com/4cf4e692-e58a-11ec-a6de-f2e2bc5c6b4b/897bb357-d20e-4cc2-983c-3240f64b3fcf.png

Logseq: everything captured flows into journals, which function as a daily inbox.
https://en.wikipedia.org/wiki/Logseq

Image

### New Workflow Possibilities 
This enables a smooth workflow:

1. Jump to a Note, you may need to edit now.
2. The Note doesn't exist yet
3. So you press the Button to create it in the Inbox.
4. Finished.

### Additional Information
Maybe I misunderstood how this button works, would be happy to get some insight, @zadam

eliandoran:
Typing a title that matches nothing offered a single creation option, and it put the note under the note you were on. In *Jump to* that is the wrong parent by construction: the dialog exists to navigate away from the current note, so the note being left behind has no claim on what gets captured.

Every surface that sets `allowCreatingNotes` now offers two, and each says where the note lands:

```
🔍  meet
──────────────────────────────────────────
➕  Create note "meet" in Inbox
↳   Create child note "meet"
📄  Meeting notes
📄  Meeting 2024-03
🔎  Search for "meet"              Ctrl+⏎
```

## Where the note goes, and saying so

The destination is decided by several rules, so "Create note" on its own asked the user to know their own configuration in order to predict it. The row now names it.

`getInboxNote()` cannot answer the question, because it creates the day note it falls back to as it answers. The decision is split out into `resolveInboxTarget()`, which reads becca and creates nothing; `getInboxNote()` materialises the day note only when the resolver says that is the destination. Both go through the same function, so a label and the capture it describes cannot disagree.

`GET /api/special-notes/inbox-target` exposes the resolved kind, note ID and title. Hoisting needs no argument — the client already sends `trilium-hoisted-note-id` on every request, which is what the resolver reads through `hoistedNoteService`:

| | destination | label |
|---|---|---|
| not hoisted | the `#inbox` note | *in `<title>`* |
| not hoisted, no `#inbox`, journal exists | today's day note | *in today's day note* |
| not hoisted, no `#inbox`, no journal | the top level | *at the top level* |
| hoisted | `#workspaceInbox`, else `#inbox` in the subtree | *in `<title>`* |
| hoisted, neither | the workspace note itself | *in `<title>`* |

The three that name a real note share one string with a `{{parentTitle}}`. The day note gets its own, having no title until the capture creates it, and so does the top level, whose note is called `root` and would name nothing the user recognises in the tree. A further string covers a failed lookup, and `{{parentTitle}}` interpolates escaped because the label is injected as raw HTML into the suggestion template.

The lookup is issued before the autocomplete search and awaited after it, so naming the destination costs a request but no extra wait. It is deliberately not cached: a cache goes stale exactly when someone labels a note `#inbox` or hoists into a workspace, which is when the label matters most.

## Not rebuilding a deleted journal

Writing the labels turned up that deleting the journal did not stop notes being date-nested — the next capture recreated the calendar, the year, the month and the day. That is [#11034](https://github.com/TriliumNext/Trilium/issues/11034), where the New button is described as impossible to escape.

`getRootCalendarNote()` treats a missing journal as one to build, which is right for the callers that ask for a day note — the calendar widget, "open today's note" — and wrong for the ones that only need somewhere to put a note. `hasCalendarRoot()` distinguishes them without building anything, and the two capture paths consult it first:

- `resolveInboxTarget()`, reaching the New button in the launcher bar, <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>P</kbd>, the tray entry, the sender API, ETAPI's `/inbox/{date}` and this dropdown.
- `getClipperInboxNote()`, which had the same shape — no `#clipperInbox`, so ask for a day note, so build a calendar. Leaving it alone would have meant a deleted journal held everywhere except the Web Clipper.

A capture with no inbox now goes to the top level unless a journal already exists, so a deleted journal stays deleted. Nothing changes where one exists, which is every installation that kept the demo data: `demo.zip` labels its Journal `#calendarRoot`. Deliberate day-note navigation is untouched and still builds the calendar it is asking for.

This is the behaviour half of #11034; the setting its reporter asked for is not here, since `#inbox` already answers "put my notes somewhere specific".

## Row order

Both creation rows sit above the search results. Putting the child-note row below them looked better — search results would keep the indices they have today, so a habitual "type, Down, Enter" would still open the note it always opened — but past a handful of results that row is not reachable in either dropdown:

- `TriliumMentionUI` renders `items.slice(0, limit)` and the `@` note feed sets no `dropdownLimit`, so the default of 10 drops a trailing row once the search returns nine or more notes. It is not scrolled past, it is never built.
- In the jQuery dropdown the row survives but sits below as many as 200 results inside a 500px scroller.

So both rows stay on top, and the spec pins the whole row order rather than the first two entries. The cost is that search results shift down by one: "type, Down, Enter" now lands on *Create child note*. The note type chooser opens before anything is written, so a mis-<kbd>Down</kbd> is a cancellable dialog rather than a stray note.

## Reach

`@`-completion takes the same pair. `createNoteForReferenceLink` gains an `intoInbox` argument rather than the action string, since the two implementations only need to choose between the inbox and a note path they already hold — `noteContext.notePath` in the text editor, `notePath` in the attribute editor. The child row takes `bx bx-subdirectory-right`, which the board and table context menus already use for their child-note actions.

The LLM chat box is the exception. `autocompleteSourceForCKEditor` forced `allowCreatingNotes`, so `@` there offered creation entries that threw: the chat editor is an `AttributeEditor`, so it carries `MentionCustomization`, but `ChatInputBar` registers only `loadReferenceLinkTitle`. The feed now takes a flag and the chat passes it, rather than the box growing a creation callback it has no use for — the marker is there to point the model at notes that exist.

## Documentation

Four User Guide pages named the old option by its label, "Create and link child note". Each now describes both options and says where each note lands, including the top-level case. *Note Inbox* — which enumerates the routes into the inbox — gains this one, and its note on the day-note fallback now says a deleted journal is not rebuilt.

Both representations were edited by hand and kept in sync: the Markdown under `docs/`, and the in-app HTML under `apps/server/src/assets/doc_notes/`. Worth a look in the app before merging.

## Testing

- `apps/client` — 91 across `note_autocomplete`, `date_notes` and `AttributeEditor`, covering one case per destination kind, both label fallbacks and the full row order.
- `apps/server` — 76 across `special_notes` and `clipper`, plus 10 for the doc notes.
- `packages/trilium-core` — its specs run under **both** the `apps/server` and `apps/standalone` suites.
- `packages/ckeditor5` — 121 across the mention specs.
- `pnpm typecheck` clean.

Three guarantees were checked by mutation rather than assumed:

- making `getInboxTarget()` resolve the day note eagerly fails *"reports the day note without creating it"* with `No context available. cls.init() must be called first.`
- dropping the `resolveInboxTarget()` guard fails *"keeps a capture at the top level when the journal has been deleted"*.
- dropping the clipper guard fails *"clips to the top level when there is no clipper inbox and no journal"*, with the day-note ID in place of `root`.

## Two things to know before merging

`note_autocomplete.create-note` keeps its key and changed its English text. The 39 locales Weblate holds still carry the old "Create and link child note" translation under it, and the new sibling keys are untranslated, so those rows fall back to English until the next sync.

ETAPI's `/inbox/{date}` changes behaviour for a database with neither an inbox nor a journal, returning the root note rather than creating a day note. Its own description reads "a fixed note (identified with `#inbox` label) or a day note in a journal", so no journal, no day note — but it is API-visible and worth a release note.

Closes #6817.

### DIFF
--- a/apps/client/src/services/date_notes.ts
+++ b/apps/client/src/services/date_notes.ts
@@ -11,6 +11,13 @@ async function getInboxNote() {
     return await froca.getNote(note.noteId);
 }
 
+/** Resolves the inbox note to a note path, for callers that create notes there. */
+async function getInboxNotePath() {
+    const inboxNote = await getInboxNote();
+
+    return inboxNote?.getBestNotePathString();
+}
+
 async function getTodayNote() {
     return await getDayNote(dayjs().format("YYYY-MM-DD"));
 }
@@ -135,6 +142,7 @@ async function getRecentLlmChats(limit: number = 10): Promise<RecentLlmChat[]> {
 
 export default {
     getInboxNote,
+    getInboxNotePath,
     getTodayNote,
     getDayNote,
     getWeekFirstDayNote,
--- a/apps/client/src/services/date_notes.ts
+++ b/apps/client/src/services/date_notes.ts
@@ -1,4 +1,4 @@
-import { dayjs } from "@triliumnext/commons";
+import { dayjs, type InboxTargetResponse } from "@triliumnext/commons";
 
 import type { FNoteRow } from "../entities/fnote.js";
 import froca from "./froca.js";
@@ -16,6 +16,14 @@ async function getInboxNotePath() {
     const inboxNote = await getInboxNote();
 
     return inboxNote?.getBestNotePathString();
+}
+
+/**
+ * Names where `getInboxNote()` would place a note, without creating the day note it would
+ * otherwise create. Honours the hoisted note, which the server reads from the request.
+ */
+async function getInboxTarget() {
+    return await server.get<InboxTargetResponse>("special-notes/inbox-target");
 }
 
 async function getTodayNote() {
@@ -143,6 +151,7 @@ async function getRecentLlmChats(limit: number = 10): Promise<RecentLlmChat[]> {
 export default {
     getInboxNote,
     getInboxNotePath,
+    getInboxTarget,
     getTodayNote,
     getDayNote,
     getWeekFirstDayNote,
--- a/apps/client/src/services/note_autocomplete.spec.ts
+++ b/apps/client/src/services/note_autocomplete.spec.ts
@@ -280,15 +280,15 @@ describe("autocompleteSource (via dataset)", () => {
         expect(server.get).not.toHaveBeenCalled();
     });
 
-    it("keeps the create-note row above the results and the create-child-note row below them", async () => {
+    it("keeps both creation rows above the results, so neither is scrolled or sliced away", async () => {
         server.get = vi.fn(async () => [{ noteTitle: "Existing", notePath: "root/y" }]) as typeof server.get;
         const { dataset } = initAndGetSource({ allowCreatingNotes: true, allowJumpToSearchNotes: true });
         const rows = await runSource(dataset, "New");
-        expect(rows.map((r) => r.action)).toEqual([ "create-note", undefined, "create-child-note", "search-notes" ]);
+        expect(rows.map((r) => r.action)).toEqual([ "create-note", "create-child-note", undefined, "search-notes" ]);
         // the inbox is resolved on selection, so the row carries no parent
         expect(rows[0].parentNoteId).toBeUndefined();
-        expect(rows[1].noteTitle).toBe("Existing");
-        expect(rows[2].parentNoteId).toBe("activeNote");
+        expect(rows[1].parentNoteId).toBe("activeNote");
+        expect(rows[2].noteTitle).toBe("Existing");
     });
 
     it("uses root as the child-note parent when there is no active note", async () => {
--- a/apps/client/src/services/note_autocomplete.spec.ts
+++ b/apps/client/src/services/note_autocomplete.spec.ts
@@ -280,16 +280,15 @@ describe("autocompleteSource (via dataset)", () => {
         expect(server.get).not.toHaveBeenCalled();
     });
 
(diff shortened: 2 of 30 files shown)


itsjakobx:
Thanks @eliandoran for taking the initiative and picking this up! I hope my multiple attempts bringing the prototype to the finish line didn't stress you out too much. It's awesome to see it fully realized!

PR #11386, opened 2026-09-07
Add a Plumber score check and badge

Totara-thib:
Follows up on #11325, merged earlier.

This adds a small workflow that runs the [Plumber](https://github.com/getplumber/plumber) CLI on pushes to main and on pull requests. It checks the security posture of the GitHub workflows, the classes of findings the hardening PR fixed, and gates at 85 points so one small finding never blocks a PR. The `.plumber.yaml` overlay inherits the built-in defaults, trusts the six third party actions this repo already uses, all pinned by commit SHA, and keeps the branch protection requirement on main since the release branches are frozen history.

## Score badge

It works like OpenSSF Scorecard's published results: every run publishes the score, the badge shows main, and a failed publish never fails your CI. No token or secret is needed. The badge shows gray UNKNOWN until the first run on main. If you want the check without the badge, removing the `score-push` line is the whole opt-out.

I work on Plumber. If you do not want the tool, no hard feelings, say so and I close this PR, the hardening stands on its own.

### DIFF
--- a/.github/workflows/plumber.yml
+++ b/.github/workflows/plumber.yml
@@ -1,0 +1,37 @@
+name: Plumber
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  plumber:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      security-events: write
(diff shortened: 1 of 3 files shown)


eliandoran:
Hi, @Totara-thib .

Thanks for the PR but we are not currently looking into expanding our tooling. Our workflows are generally hand-maintained so we don't see much of a benefit in automating it.

PR #11406, opened 2026-09-08
fix(llm): preserve explicit OpenAI-compatible API versions

### LINKED ISSUE #11323: Base URL for custom LLM providers is incorrect
### Description

智谱使用openai协议接入 结尾是V4 自动补全了 V1
地址:https://open.bigmodel.cn/api/paas/v4/v1/chat/completions
补全的时候多补了  v1/  

Image

Image

Image

### TriliumNext Version

0.105.0

Windows

### What is your setup?

Local (no sync)

### Operating System Version

WINDOWS 10

### Error logs

_No response_

dajiaohuang:
## What

Preserve an explicit API version in an OpenAI-compatible endpoint URL. A provider configured with `https://open.bigmodel.cn/api/paas/v4` now uses that URL for chat and model listing instead of appending `/v1`.

Bare hosts and the existing `/v1` form keep their current behavior. Native Ollama and LM Studio probes continue to use the host portion of a versioned endpoint.

Fixes #11323

## Testing

- `pnpm --filter server exec vitest run --config vite.config.mts packages/trilium-core/src/services/llm/providers/local.spec.ts`
- `pnpm typecheck`
- `git diff --check`

### DIFF
--- a/packages/trilium-core/src/services/llm/providers/local.spec.ts
+++ b/packages/trilium-core/src/services/llm/providers/local.spec.ts
@@ -108,6 +108,11 @@ describe("LocalProvider", () => {
             expect(createOpenAiMock).toHaveBeenLastCalledWith({ apiKey: "local", baseURL: "https://proxy.example.com/llm/v1", fetch: llmFetch });
         });
 
+        it("preserves an explicit API version other than /v1", () => {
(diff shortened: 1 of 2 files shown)


eliandoran:
Hi, @dajiaohuang 

Thanks for the PR but I would prefer a simpler approach that does not use regexes. I'll fix it in #11410.
Please test it in tomorrow's nightly to ensure it was fixed for your use case.

PR #11423, opened 2026-09-09
Board: multiple selection, native drag & drop and other improvements

adoriandoran:
The Kanban board now supports working with several cards at once. Cards can be picked out with Ctrl and Shift click, gathered a whole column at a time, and then moved, edited, archived or deleted as a group, whether by dragging them, by the card menu or from the keyboard. Alongside this, cards can now be dragged out of the board entirely, onto the note tree or onto a board in another split, and a card can be made to act as a shortcut that opens a different note. The board's settings have been gathered into a dialog reachable from the note menu, and the card editor has been reworked so that opening it neither changes a card's size nor risks a stray click landing elsewhere on the board.

To create a redirection card, add the relation `boardCardRedirectTo` to the card's note and point it at the note that should open. Clicking the card, or pressing Space while it is focused, then navigates to that note in the pane the board occupies rather than opening the card's own editor. The card's title is underlined on hover to indicate that it leads elsewhere. Use Quick edit from the card's context menu to open the editor again. The relation is inherited, so setting it on a parent note applies it to every card beneath that parent.

### Multi-selection

- Add Ctrl+click to add or remove a single card from the selection, accumulating across columns.
- Add Shift+click to select the range between the anchor and the clicked card, restricted to a single column.
- Add Ctrl+A to select every card in the column that holds focus.
- Clear the selection with Escape, with a click anywhere on the board outside a card, or by right clicking a card that is not part of it.
- Apply the card context menu to the whole selection: column moves, promoted attribute values, archiving, removal from the board and deletion.
- Mark a menu entry only where every selected note holds the same value, so a selection that disagrees shows no check and picking the entry writes the value to all of them.
- Apply Delete and Shift+Delete to the whole selection rather than to the focused card alone.
- Move the whole selection between columns with Ctrl and the arrow keys, and within a column when the selected cards form a contiguous run.
- Keep the selection intact after a drop, and scroll the last card moved into view.
- Prune the selection to the cards the board still draws after a refresh.
- Draw a selected card with an inset ring, and add the selection shortcuts to the board's shortcut hints.

### Drag and drop

- Drag a selection as a single stack of cards showing how many are being carried.
- Add a native drag when Ctrl is held as the press begins, allowing cards to be dropped on the note tree, on other collections and on boards in other splits and windows.
- Carry the whole selection in the native drag payload, in the same form the note tree writes and reads.
- Move a dropped set of cards in one operation instead of one card after another, removing the visible restaging as each landed.
- Hold the board as the drop drew it until the matching entity changes arrive, so a column no longer collapses and re-expands after a drop.
- Suppress the click a completed drag produces, which previously cleared the selection.
- Make `moveToColumnEnd` and `moveWithinBoard` take a set of cards natively, and batch the attribute writes behind them.

### Redirection cards

- Add the `boardCardRedirectTo` relation, which makes clicking a card or pressing Space navigate to the target note instead of opening the card's editor.
- Navigate the note context the board is drawn in rather than the globally focused one, so a redirect in a split does not replace the note in another pane.
- Underline the title of a redirecting card on hover, leaving the card's icon undecorated.
- Register the relation as a built-in attribute, with autocompletion and an explanation in the attribute detail popup.

### Board properties and menus

- Add "Board properties" to the note menu, placed above Advanced and shown only for board notes.
- Rename the board's sort setting to "Default card order" and describe the columns it governs.
- Replace the sort actions button with a nested link that resets every column's sorting, after a confirmation prompt.
- Hide "Remove from board" and leave plain Delete unanswered while the inbox column is drawn, since clearing a card's grouping value leaves it in the inbox rather than off the board.
- Omit the Options field when creating a grouping attribute from the group-by dropdown, the board's columns being what supplies those options.
- Add a `hideTypeOptions` flag to the attribute detail popup, which leaves out the rows a chosen kind adds.

### Card editing

- Keep a card's badges visible while its title is being edited, and give both modes the same metrics so the card does not change size.
- Dim the board behind a card that is being named, lifting the edited card above the dimming.
- Stop the remaining cards answering the pointer once the dimming has faded in, so dismissing the editor cannot open or select another card.
- Apply that suppression immediately when motion is disabled, where no transition ends to trigger it.

### Fixes

- Redraw a board that shares a tab with the focused pane, instead of deferring every change until something else forces a refresh.
- Write nothing for a drop that leaves the drawn order unchanged, so a filter's hidden cards are no longer reordered behind it.
- Compare a card's column membership against the unfiltered column, so an active filter no longer causes a redundant grouping write.

### Internals

- Add a reusable `SelectionStore` in `widgets/react/selection.ts`, an external store read through `useSyncExternalStore` so that selecting a card wakes only that card.
- Add `services/note_set.ts`, holding the consensus reader `shared()` and the fan-out writers for labels and the archived state.
- Widen `buildAttributeMenuItems` and `getArchiveMenuItems` to take a list of notes, updating the calendar collection accordingly.
- Generalise `applyCardMove` into `applyCardMoves`, which places a set of cards at one index in a single pass.

### DIFF
--- a/apps/client/src/components/app_context.ts
+++ b/apps/client/src/components/app_context.ts
@@ -569,6 +569,8 @@ type EventMappings = {
     geoMapCreateChildNote: {
         ntxId: string | null | undefined; // TODO: deduplicate ntxId
     };
+    /** Opens the board's properties dialog, which only the board itself can show. */
+    showBoardProperties: { ntxId: string | null | undefined; };
     tabReorder: {
         ntxIdsInOrder: string[];
     };
--- a/apps/client/src/menus/context_menu_utils.ts
+++ b/apps/client/src/menus/context_menu_utils.ts
@@ -1,24 +1,32 @@
+import type FNote from "../entities/fnote"
+import type { MenuItem } from "./context_menu"
 import { t } from "../services/i18n"
-import attributes from "../services/attributes"
-import FNote from "../entities/fnote"
+import { setArchivedOnNotes, shared } from "../services/note_set"
 import { escapeHtml } from "../services/utils"
 
-export function getArchiveMenuItem(note: FNote) {
-    if (!note.isArchived) {
-        return {
-            title: t("board_view.archive-note"),
-            uiIcon: "bx bx-archive",
-            handler: () => attributes.addLabel(note.noteId, "archived")
-        }
-    } else {
-        return {
+/**
+ * The entry that archives the notes, or unarchives them when they are all archived already.
+ *
+ * Returns nothing where the notes disagree: one entry cannot say what picking it would do, and
+ * both entries at once would read as two states the same notes are in.
+ */
+export function getArchiveMenuItems<T>(notes: FNote[]): MenuItem<T>[] {
+    const state = shared(notes, (note) => note.isArchived)
+    if (!state.agreed) {
+        return []
+    }
+
+    return [ state.value
+        ? {
             title: t("board_view.unarchive-note"),
             uiIcon: "bx bx-archive-out",
-            handler: async () => {
-                attributes.removeOwnedLabelByName(note, "archived")
-            }
+            handler: () => { void setArchivedOnNotes(notes, false) }
         }
-    }
+        : {
+            title: t("board_view.archive-note"),
+            uiIcon: "bx bx-archive",
+            handler: () => { void setArchivedOnNotes(notes, true) }
+        } ]
 }
 
 /**
--- a/apps/client/src/services/note_set.spec.ts
+++ b/apps/client/src/services/note_set.spec.ts
@@ -1,0 +1,124 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import { buildNote } from "../test/easy-froca";
+import froca from "./froca";
+import {
+    clearLabelOnNotes, setArchivedOnNotes, setLabelOnNotes, shared
+} from "./note_set";
+
+// The writes alone, so the rest of the service reads a note the way the application does. Both
+// shapes are replaced: `note_set` calls one of these by name and the others off the default export.
+const writes = vi.hoisted(() => ({
+    setLabelValues: vi.fn(),
+    addLabel: vi.fn(),
+    removeOwnedLabelByName: vi.fn()
+}));
+vi.mock("./attributes", async (importOriginal) => {
+    const original = await importOriginal<{ default: object }>();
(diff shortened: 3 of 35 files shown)


deftdawg:
Because I'm new to Trilium and somehow think a kanban board of kanban boards is a good idea for projects and project tasks :rofl:  ...  I was thinking to create a PR to allow filtering visibility of child board columns and children nodes (so display embedded board, but not it's children) 
image

and to filter parent columns in children boards...  
image

I can see you're cooking here with these enhancements (btw they look **very nice**!), so I'll sit on the sidelines until you merge before I take a shot at it.

adoriandoran:
@deftdawg 

> I can see you're cooking here with these enhancements (btw they look very nice!)

I'm glad that you like it.

> and somehow think a kanban board of kanban boards is a good idea for projects and project tasks.

Well, that's actually a messy thing to sort out.

If you nest multiple boards, there will be different problems: the promoted attributes of the parent board will merge with the promoted attributes of children boards, the parent board will list the cards belonging to children boards.

I don't think that there is an elegant way to fix that, especially the attribute inheritance (which will need fundamental changes). Filtering will make the problem less obvious, but the problem will still exist and may cause unexpected issues.

Most likely the best solution would be to make a card act as a shortcut to another board, which is not child of the current board. But that is not supported yet.

adoriandoran:
@deftdawg, the “shortcut card” is done now (in this PR). First, set up your main board:

1. Go to “Board Properties” → Card Attributes → Create a new attribute.

2. Set “boardCardRedirectTo” as name, “Relation” as type, and “Target” (or whatever you like) as display name then press “Save & close”.

3. You can uncheck the visibility of the attribute.

4. Go and create new item on the board. Edit it and for "Target" select a board from your database (outside of the current board subtree). Then close the editor.

image


5. If you click again on this card, you will be redirected to the other board. Then use the back button to return to the previous board.

image


To edit again the card, right click on it and select “Quick Edit”.

deftdawg:
> @deftdawg, the “shortcut card” is done now (in this PR). First, set up your main board:
> 
>     1. Go to “Board Properties” → Card Attributes → Create a new attribute.
> 
>     2. Set “boardCardRedirectTo” as name, “Relation” as type, and “Target” (or whatever you like) as display name then press “Save & close”.
> 
>     3. You can uncheck the visibility of the attribute.
> 
>     4. Go and create new item on the board. Edit it and for "Target" select a board from your database (outside of the current board subtree). Then close the editor.

@adoriandoran, you, sir, are the greatest! 🍻   I look forward to using and abusing this. 😄

PR #11434, opened 2026-09-10
feat(ETAPI): accept optional dateModified/utcDateModified on create-note

deftdawg:
## Problem
Migration importers (e.g. Joplin -> Trilium via trilium-py, which reads the front matter `updated:` date) cannot preserve the source note's last-modified date: `POST /etapi/create-note` only accepts `dateCreated`/`utcDateCreated`, and every regular `save()` re-stamps `now` via `BNote.beforeSaving`. Imported notes all show the import time as last-modified.

## Change
- `NoteParams` gains optional `dateModified?: string` (local) and `utcDateModified?: string` (UTC).
- `createNewNote` validates both with the existing local/UTC datetime validators and, when either is supplied, applies them **after all content writes** (including the template path) via the same raw-SQL bypass the ENEX/OneNote/Notion/Keep/Anytype/Obsidian importers use (`BNote.setDateCreatedAndModified`, which also keeps the blob's stamp in sync). `utcDateModified` wins when both are given, otherwise it is derived from the local value.
- ETAPI allowlist and OpenAPI `CreateNoteDef` updated; no other endpoint touched.

## Backward compatibility
Both fields are optional. When omitted, behavior is exactly as before (import-time stamp). No existing request shape changes, and the response shape is unchanged.

## Scope note
`PATCH /etapi/notes/:noteId` is intentionally out of scope: its `note.save()` would need the same post-save treatment. Happy to follow up if maintainers want it.

## Tests
- New cases in `apps/server/spec/etapi/create-entities.spec.ts`: explicit pair round-trips, local-only derives the UTC value, omitted fields still stamp current time.
- Verified locally: `spec/etapi/` + Keep importer integration spec — 36 files, 175 tests, all pass.

### DIFF
--- a/apps/server/spec/etapi/create-entities.spec.ts
+++ b/apps/server/spec/etapi/create-entities.spec.ts
@@ -106,6 +106,41 @@ describe("etapi/create-entities", () => {
         );
     });
 
+    it("restores modification dates via PATCH after a content rewrite", async () => {
+        const created = await createNoteWithTimestamps("modified-patch-restore", {
+            dateModified: "2024-05-06 10:11:12.456+0200",
+            utcDateModified: "2024-05-06 08:11:12.456Z"
+        });
+        const noteId = created.body.note.noteId;
+
+        // A content rewrite (e.g. attachment URL replacement) re-stamps now.
+        await supertest(app)
+            .put(`/etapi/notes/${noteId}/content`)
+            .auth(USER, token, { "type": "basic"})
+            .set("Content-Type", "text/plain")
+            .send("rewritten <p>content</p>")
+            .expect(204);
+
+        const bumped = await supertest(app)
+            .get(`/etapi/notes/${noteId}`)
+            .auth(USER, token, { "type": "basic"})
+            .expect(200);
+        expect(asInstant(bumped.body.utcDateModified)).toBeGreaterThan(
+            Date.parse("2024-05-06T08:11:12.456Z")
+        );
+
+        // PATCH restores the source date despite save() re-stamping now.
+        const restored = await supertest(app)
+            .patch(`/etapi/notes/${noteId}`)
+            .auth(USER, token, { "type": "basic"})
+            .send({ utcDateModified: "2024-05-06 08:11:12.456Z" })
+            .expect(200);
+        expect(restored.body.utcDateModified).toStrictEqual("2024-05-06 08:11:12.456Z");
+        expect(asInstant(restored.body.dateModified)).toStrictEqual(
+            Date.parse("2024-05-06T08:11:12.456Z")
+        );
+    });
+
     it("obtains attribute information", async () => {
         const response = await supertest(app)
             .get(`/etapi/attributes/${createdAttributeId}`)
--- a/apps/server/spec/etapi/create-entities.spec.ts
+++ b/apps/server/spec/etapi/create-entities.spec.ts
@@ -76,6 +76,36 @@ describe("etapi/create-entities", () => {
         });
     });
 
+    it("preserves modification dates on create-note", async () => {
+        const explicit = await createNoteWithTimestamps("modified-override", {
+            dateCreated: "2023-08-21 23:38:51.123+0200",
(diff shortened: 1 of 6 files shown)


deftdawg:
Noticed an issue with notes being imported from Joplin that have attachments (i.e. images) having their last modification times set to import time - converting to draft to block merge while I resolve that.

eliandoran:
Tell me more about the importer you are using/building.

deftdawg:
> Tell me more about the importer you are using/building.

Sure, basically trying to migrate a collection of ~1500 Joplin [including some really old nvAlt] notes that I've been using for over a decade by leveraging Nriver/trilium-py with a custom shell script... 

In my setup, Joplin sorts notes that were modified most recently to the top, when I imported using Trilium-py all my notes were were imported into Trilium with created and updated timestamps set to the import time breaking sort.   Essentially, all context of what I was working on recently is lost...   I do plan to heavily reorganize my collection to take advantage of Trilium, but for now I need it to function as a 1:1 replacement for Joplin so I can continue to be productive with it. 

This PR is attempting to expose a way for modification date to be set via ETAPI.   I'm also hitting a few additional edge cases, such as the background image downloader changing the modified timestamp on a note after it finishes downloading an image (which I think has just been addressed) and another issue with folders also not correctly having create/update timestamps imported (in progress).

It's possible that this code won't be accepted by Trilium and that's fine (my clanker actually suggested it would be easier to just fudge everything after importing, but I'd rather try to solve to the problem upstream)...  I need it for hopefully a 1-time import, once I have all the data in the database properly I won't need it anymore, but I thought I should do it in the open in case it is helpful or others would like to follow.

deftdawg:
@eliandoran - You can find the migration script here: https://github.com/deftdawg/joplin2trilium

deftdawg:
Couple more issues found today:

- [x] 1. Joplin folder icons aren't preserved on import (fix will be to append them to title - there's no 1:1 mapping between Emoji's and BoxIcons and prepending would look bad + mess up sorting since emojis sort first/randomly) -> **Importer Script** will implement the fix
- [x] 2. Joplin links are imported as new attachments rather than trilium links -> Investigating best spot for fix, might also be **Importer Script**
- [x] 3. Markdown probably needs to be linted/reformatted before import...  One example I've seen a note with two check lists was imported into Trilium, first list was converted to a checkbox list correctly, second was imported as raw text - the difference between them is the second list did not have a blank line between list title and first checkbox (which means it's not strictly compliant) -> **Importer Script** will probably also implement this fix
- [x] 4. Some title characters are being replaced with _; observed with `/` -> `_` and `?` -> `_` (trilium-py was using note file name instead of reading title from Joplins' front matter).

**EDIT:** All additional issues found here have been resolved at the import script or trilium-py library levels.   Also generating an import report that explains markdown formatting changes.
image

deftdawg:
I've completed my migration efforts using this branch to load the data.

My importer project leverages my fork of [trilium-py](https://github.com/deftdawg/trilium-py/tree/joplin-front-matter-improvements) ([PR pending upstream](https://github.com/Nriver/trilium-py/pull/68)) and this branch.

[Joplin2Trilium](https://github.com/deftdawg/joplin2trilium) repo page:
image

eliandoran:
@deftdawg , as an alternative you can use an emoji icon pack for Trilium, although probably the mapping might be a difficult. At some point we might consider embedding that emoji pack into Trilium by default but not on the road map.

deftdawg:
> mapping might be a difficult

It was going to be a manual effort mapping Emojis into BoxIcons and it wouldn't have been a complete mapping. Eventually I might select new icons for the folders, this approach was low friction and good enough for now. 😄

deftdawg:
Over the weekend, I completed the Joplin migration my home machine (NixOS) to a version of Trilium-Server and Trilium-Notes based off this PR and moved mobile to TriNote (iOS).

I'll probably focus on bug-fixes/improving TriNote for a bit.  

Once this merges/if it merges and releases, I'll migrate my work Joplin instance over to Trilium Notes too. 🚀

PR #11436, opened 2026-09-10
Board: massive performance improvements

adoriandoran:
Boards containing very large columns are now usable. A column holding several thousand cards previously caused visible stuttering when scrolling, dragging a card, or simply moving the pointer across it, and adding a card could freeze the board for over a second. The board now renders only the cards that are actually visible, so a column of ten thousand cards behaves like a column of forty. Several defects in card placement have also been corrected, including a card appearing part way up a column before jumping to its final position, and the drop indicator drifting away from the position a card would actually land in.

### Measurements

Recorded in headless Chromium against a development build, on a board whose largest column holds 10,027 cards. Frame figures are the median and worst frame observed during the interaction; long tasks are the total main-thread blocking time it produced.

| Interaction (before →  after) | Median frame | Worst frame | Long tasks |
|---|---|---|---|
| Scrolling the column | 37.7ms → 16.8ms | 286ms → 40ms | 900ms → 0ms |
| Picking up a card | 16.8ms → 16.7ms | 294ms → 33ms | 435ms → 0ms |
| Releasing a card | 16.7ms → 16.7ms | 110ms → 19ms | 176ms → 0ms |
| Moving the pointer across the cards | 18.1ms → 16.5ms | 56ms → 19ms | 165ms → 0ms |
| Scrolling the board sideways | 16.7ms → 16.7ms | 41ms → 17ms | 0ms → 0ms |

| Measure | Before | After |
|---|---|---|
| DOM elements rendered by the board | 50,022 | 390 |
| Style recalculation per structural change within the board | 435ms | 22ms |
| Drop indicator displacement over 70,800px of auto-scroll | 1,510px | 86px |

### Rendering of large columns

- Render only the cards near a column's scroll position, with spacer elements preserving the height of those omitted, for columns holding more than 120 cards.
- Move the rendered window in steps of 25 cards, so that scrolling past a single card does not re-render the column.
- Cache each card's measured height by note id and reuse it wherever that card is rendered.
- Settle the height assumed for a column's unmeasured cards after 12 of its own cards have been measured, and hold that value thereafter.
- Scope the assumed height to each column, so that columns with differing promoted attributes do not inherit one another's estimates.
- Clear all cached heights on window resize, which alters column width and therefore card heights.
- Adjust a column's scroll position when a height correction changes the space above the viewport, so that the content under the reader does not shift.
- Disable `overflow-anchor` on a windowed column, so that the browser does not compete with that adjustment.
- Suppress the card FLIP animation for the commit in which the window changes, so that scrolling is not animated as a reorder.
- Scroll a column to a card created outside the rendered window, since such a card cannot scroll itself into view.

### Style recalculation

- Replace the three board `:has()` selectors with an `editing-open` class set from board state, eliminating a full-board style recalculation on every structural change within the board.
- Report an open insert field from each column to the board, replacing the `:has()` detection this previously required.
- Mark the column holding an open editor with its own `editing-open` class, replacing the per-column `:has()` selector that lifted its mask.

### Card movement

- Compute the drop index from a column's current card heights rather than from measurements taken when the drag began, so that the drop indicator no longer drifts from the resulting position during an auto-scroll.
- Offset the drop index by the column's top padding, which previously displaced every computed position by eight pixels and could select the following slot.
- Record each card's index within its column as a `data-index` attribute, so that a drag identifies a position in the column rather than among the rendered cards.
- Synthesise measurement boxes for the cards a windowed column does not render, keeping every drag index expressed in the column's own terms.
- Use the unfiltered column length when placing a card moved into another column, so that a card no longer appears part way up a filtered column before jumping to the end.
- Convert a filtered index to its unfiltered position when reordering cards within a column.
- Place the drop indicator against the first rendered card when the drop position lies above the rendered window.

### Keyboard navigation

- Navigate a column by card index rather than by position among the rendered cards, so that `Home` and `End` reach the column's true first and last card.
- Scroll a column to a card it is not currently rendering and focus that card within the same keystroke.
- Focus a card moved into a column that is not rendering the position it lands in.
- Fall back to the previously focused position when a redraw removes the focused card, so that a navigation key no longer scrolls the board instead of moving through it.

### Tests

- Add unit tests covering window computation, height resolution, per-column height estimation, and the scroll and reveal operations.
- Add board-level tests covering a windowed column's rendered slice, its spacer heights, and the indices assigned to its cards.
- Add keyboard tests covering navigation onto unrendered cards, `Home` and `End` across the full column, and a move into a column that is not rendering the destination.
- Add drag measurement tests covering the synthesised boxes for unrendered cards and the reported content origin.
- Add a regression test confirming that a card sent to a filtered column is placed at the column's end.
- Add tests confirming that the editing backdrop and the per-column mask are driven by state rather than by selector matching.

### DIFF
--- a/apps/client/src/widgets/collections/board/board_drag.ts
+++ b/apps/client/src/widgets/collections/board/board_drag.ts
@@ -583,7 +583,13 @@ function startCard(
     const noteId = element?.dataset.noteId;
     if (!element || !columnElement || !noteId) return null;
 
-    const cards = [ ...columnElement.querySelectorAll(".board-note") ];
+    // A windowed column draws a slice of its cards, so the card says which place it holds rather
+    // than being counted among the ones on screen.
+    const stated = element.dataset.index;
+    const place = stated === undefined ? Number.NaN : Number(stated);
+    const index = Number.isFinite(place)
+        ? place
+        : [ ...columnElement.querySelectorAll(".board-note") ].indexOf(element);
     const noteIds = carriedWith(noteId);
     return {
         kind: "card",
@@ -593,7 +599,7 @@ function startCard(
             noteId,
             noteIds,
             fromColumn: columnElement.dataset.column ?? "",
-            index: cards.indexOf(element),
+            index,
             height: carriedHeight(element, noteIds)
         },
         position: null,
--- a/apps/client/src/widgets/collections/board/board_drag.ts
+++ b/apps/client/src/widgets/collections/board/board_drag.ts
@@ -3,10 +3,13 @@ import { useCallback, useEffect, useRef, useState } from "preact/hooks";
 
 import { useTrackedElement } from "../../react/hooks";
 import {
-    type CardBox, cardInsertionIndex, columnAt, columnCovers, columnInsertionIndex
+    type CardBox, cardInsertionIndex, columnAt, type ColumnBox, columnCovers, columnInsertionIndex
 } from "./drag_geometry";
-import { type BoardMeasurement, measureBoard, toAreaY, toBoardX } from "./drag_measure";
+import {
+    type BoardMeasurement, measureBoard, placeInModel, toAreaY, toBoardX
+} from "./drag_measure";
 import { createEdgeScroller, type ScrollTarget } from "../../react/edge_scroll";
+import { getColumnModel } from "./windowing";
 
 /** How far a pointer travels before a press with a button held is taken for a drag. */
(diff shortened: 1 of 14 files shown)


LeeBoNorth:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #11470, opened 2026-09-13
fix(client/collections): stop a failed page load from failing the client suite

eliandoran:
## Why

[#11467](https://github.com/TriliumNext/Trilium/pull/11467) (maplibre-gl 6.9.0) failed CI on a job that has nothing to do with maplibre. The `Test development` job's client-test step reported:

```
 Test Files  438 passed (438)
      Tests  5718 passed (5718)
     Errors  1 error
```

Not a failing test — a single unhandled promise rejection:

```
Error: A module tried to load from the server the following notes: child-a
This is not supported, use Froca mocking instead and ensure the note exist in the mock.
 ❯ Object.post            apps/client/src/test/setup.ts:110:27
 ❯ FrocaImpl.reloadNotes  apps/client/src/services/froca.ts:193:35
 ❯ FrocaImpl.getNotes     apps/client/src/services/froca.ts:269:20
 ❯ Object.__              apps/client/src/widgets/collections/Pagination.tsx:176:15
 ❯ preact/hooks → preact/test-utils (act)
originated in "apps/client/src/widgets/collections/NoteList.spec.tsx"
```

`apps/client/vite.config.mts` sets neither `dangerouslyIgnoreUnhandledErrors` nor `retry`, so one stray rejection reddens a suite whose tests all pass.

That failure is not the dependency's. The same error, with the same stack and the same `438 passed / 5718 passed / 1 error`, failed [#11468](https://github.com/TriliumNext/Trilium/pull/11468) (zod 4.6.1) in the same CI wave at 00:22 UTC; #11468 then went green on a re-run at 07:19 and merged. Two causes have to line up for it to fire, and this fixes both.

## `usePagination` had no rejection handler

`usePagination` resolved its page slice with `froca.getNotes(pageNoteIds).then(...)` and nothing on the rejection path, so a failed `tree/load` had no handler at all. In the browser that leaves the console with a bare rejection and no indication of which view asked for the notes; under Vitest it fails the entire run.

It now reports the failure and leaves `pageNotes` as it stands, so a view that cannot load its page keeps showing the last one it had.

## `NoteList.spec.tsx` used note ids froca never held

The spec fed `useNoteIds` the bare strings `child-a`, `child-b` and `child-new`. A collection view resolves the ids it is handed through `froca.getNotes()`, and `apps/client/src/test/setup.ts` makes a `tree/load` for an unknown id throw — so whenever a pagination effect carrying one of those ids was flushed, the mock refused it. They are now built with `buildNotes()`, the way `Pagination.spec.tsx` already builds its own 25, so the slice resolves from the cache and never reaches for the server.

## Verification

- New spec in `Pagination.spec.tsx` covers the rejection path. With the `.catch` reverted it fails **and** raises the unhandled rejection (`1 failed | 6 passed`, `Errors 1 error`); with the fix in place, `7 passed`.
- `pnpm --filter client test collections` — 66 files, 1331 tests, green, and green again over 6 consecutive runs.
- `pnpm typecheck` — no errors.

Nothing user-visible changes, so no User Guide page is affected.

## Not covered here

The client suite stays vulnerable to this class of failure in general: on plain `main`, a full `pnpm --filter client test` run here ended `439 test files passed, 5753 tests passed, 4 errors`, all four being Bootstrap `transitionend` timers firing after the test that queued them (`item_picker`, `backup.options` ×2, `ocr_text`). That is a separate leak, left for its own change.

https://claude.ai/code/session_01G8kR3XXUJ6urDt1p2iuk3u

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8kR3XXUJ6urDt1p2iuk3u)_

### DIFF
--- a/apps/client/src/widgets/collections/NoteList.spec.tsx
+++ b/apps/client/src/widgets/collections/NoteList.spec.tsx
@@ -18,11 +18,19 @@ import Component from "../../components/component";
 import type FNote from "../../entities/fnote";
 import type { EntityChange } from "../../server_types";
 import LoadResults from "../../services/load_results";
-import { buildNote } from "../../test/easy-froca";
+import { buildNote, buildNotes } from "../../test/easy-froca";
 import { ParentComponent } from "../react/react_utils";
 import { CustomNoteList, useNoteIds } from "./NoteList";
 
 let currentNoteIds: string[] = [];
+
+// The child ids below stand for real notes: a collection view resolves the ids it is given through
+// froca.getNotes(), which round-trips to the server for anything the mock does not hold.
+buildNotes([
+    { id: "child-a", title: "Child A" },
+    { id: "child-b", title: "Child B" },
+    { id: "child-new", title: "Child New" }
+]);
 
 /** Drains the chained awaits in `refreshNoteIds` (getNoteIds → search promise → setNoteIds). */
 async function flushMicrotasks() {
--- a/apps/client/src/widgets/collections/Pagination.spec.tsx
+++ b/apps/client/src/widgets/collections/Pagination.spec.tsx
@@ -167,19 +167,28 @@ describe("usePagination", () => {
         }
     });
 
-    it("reports a failed note load instead of leaving the rejection unhandled", async () => {
-        const getNotesSpy = vi.spyOn(froca, "getNotes")
-            .mockRejectedValue(new Error("tree/load failed"));
+    it("reports a failed note load, and stays quiet about a superseded one", async () => {
+        // Hand out the reject side of each load, so the page-1 load can be failed after the flip
+        // to page 2 has already superseded it.
+        const pending: Array<(reason: Error) => void> = [];
+        const getNotesSpy = vi.spyOn(froca, "getNotes").mockImplementation(() =>
+            new Promise<FNote[]>((_resolve, reject) => pending.push(reject))
+        );
         const consoleSpy = vi.spyOn(console, "error").mockImplementation(() => {});
         try {
             const note = buildNote({ title: "Search", type: "search" });
             await mount(note, 10);
-            await settle();
+            await act(async () => observed?.setPage?.(2));
+            expect(pending).toHaveLength(2);
 
-            // An unhandled rejection here fails the whole Vitest run, not just this test.
-            expect(consoleSpy).toHaveBeenCalled();
+            const [ supersededLoad, currentLoad ] = pending;
+            await act(async () => supersededLoad(new Error("tree/load failed")));
+            expect(consoleSpy).not.toHaveBeenCalled();
+
+            await act(async () => currentLoad(new Error("tree/load failed")));
+            expect(consoleSpy).toHaveBeenCalledTimes(1);
+            expect(observed?.page).toBe(2);
             expect(observed?.pageNotes).toBeUndefined();
-            expect(observed?.page).toBe(1);
         } finally {
             consoleSpy.mockRestore();
             getNotesSpy.mockRestore();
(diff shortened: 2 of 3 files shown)


eliandoran:
Both Greptile findings addressed in 2d068a6c. (Replying on the threads themselves kept returning HTTP 500, so the reasoning is here; the threads are resolved.)

**Superseded loads still log** (`Pagination.tsx`) — valid. The `.then` side drops a superseded result, so the `.catch` side had no business reporting one: an unmount or a fast page flip would have named a page the user had already left. Guarded on `!cancelled`, which still marks the promise handled and so still keeps the rejection from escaping, which is the point of the PR.

The spec now drives both sides through manually-rejected loads, so neither assertion is decoration:

- dropping the `!cancelled` guard fails it on `expected "error" to not be called at all, but actually been called 1 times`;
- dropping the whole `.catch` fails it on the second assertion, with two unhandled rejections alongside.

**Comment narrates the defect** (`Pagination.spec.tsx`) — fair. That line explained a consequence of the missing `.catch` rather than anything the code in front of it does, which is the "never comment an absence" half of the rule. It is gone; the rewritten test's remaining comment describes the deferred mechanics the assertions rely on, and the reproduction detail lives in the commit body.

`pnpm typecheck` clean, `pnpm --filter client test collections` green (66 files, 1331 tests).

---
_Generated by [Claude Code](https://claude.ai/code)_

eliandoran:
Three checks are red on `2d068a6c` — `Analyze (actions)`, `Analyze (javascript-typescript)` and `Build and Deploy App`. I don't believe any of them belongs to this PR, and I can't re-run them from here. Someone with repo access should hit **Re-run failed jobs** on runs [34748448425](https://github.com/TriliumNext/Trilium/actions/runs/34748448425) and [34748448418](https://github.com/TriliumNext/Trilium/actions/runs/34748448418).

Why they read as infrastructure rather than code:

- **`Test development` passed** — the client suite is the one thing this PR actually changes.
- **`Analyze (actions)` analyses `.github/workflows/**` only, and this PR touches no workflow file.** That check passed on the identical input at 08:21 (run [34747467231](https://github.com/TriliumNext/Trilium/actions/runs/34747467231)) and failed at 08:46. Same bytes in, different result out.
- **Both CodeQL jobs finished their analysis and failed only at the upload.** The logs show all 18 queries evaluated and `Exported results to SARIF (78ms)`, then `Uploading code scanning results` → 18 seconds of nothing → `CodeQL job status was failure`.
- **`Build and Deploy App` reached its Deploy step.** It ran 71s against 74s for the run that passed at 08:21, and `pnpm --filter=standalone build` succeeds on this head locally (exit 0, built in 15.66s). The Deploy step is where the job calls the GitHub API to post the preview comment.
- **All three sit in one ~3-minute window**, 08:45:45–08:48:14, each at the point where the job calls back into GitHub's API. Every other check on the same head — server, standalone, CKEditor ×2 + coverage gate, both Docker builds, all six E2E jobs, Greptile — is green.

The corroborating detail: GitHub's REST API was returning HTTP 500 to this session throughout that window (three attempts to reply on the review threads above failed that way, which is why those answers went into a PR comment instead), and `POST /actions/runs/{id}/rerun-failed-jobs` is still returning 500 now, on four attempts across half an hour. That is also why I can't re-run them myself.

If a re-run comes back red, it's mine and I'll fix it.

---
_Generated by [Claude Code](https://claude.ai/code)_

eliandoran:
Re-ran and all three came back green on the same commit — `Analyze (actions)`, `Analyze (javascript-typescript)` and `Build and Deploy App`. No code changed between the red and green runs, which settles the diagnosis above: the failures were GitHub-side, not this PR's.

The `rerun-failed-jobs` endpoint that had been returning 500 for half an hour accepted the request on this attempt, so that outage cleared too.

All 23 checks green, `mergeable_state: clean`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #7024, opened 2025-09-18
where is my clone note?：An easy way to find clone note


youli42:
### Describe feature

克隆笔记是我最喜欢使用的功能之一，这个特效让我可以无需在笔记中添加复杂的标签和搜索工具，在分类上它帮了我大忙。

Cloning notes is one of my favorite features to use. This effect allows me to avoid adding complex tags and search tools in the notes, and it has helped me a lot with categorization.

但是，当我 ”过度“ 使用这个功能时，我难以找到我的笔记的具体位置：

However, when I use this feature "excessively", it's difficult for me to find the specific location of my notes:

<img width="1087" height="319" alt="Image" src="https://github.com/user-attachments/assets/77a01ba9-3e94-43e9-8821-41ded7b37f24" />

在笔记路径功能中，他因其父级同样经过多次克隆，导致难以识别哪些路径来自他自身被克隆的结果。
In the note path function, because its parent has also been cloned multiple times, it is difficult to identify which paths result from its own cloning.

综上，我希望有一个更优雅的方式定位我的克隆
In conclusion, I hope there is a more elegant way to locate my clones.

### Additional Information

可能的解决方法：使用笔记地图

树形地图下，可以很好的合并父目录，但目前有几个问题：

1. 全局地图不支持搜索：这让我难以定位我的目标笔记
2. 笔记的局部地图无法显示父级

期待您的想法，感谢所有开发者的付出！

Possible solution: Use note maps

Under the tree map, parent directories can be merged well, but there are currently several issues:

1. The global map does not support searching: This makes it difficult for me to locate my target notes.
2. The local map of a note cannot display its parent level.

Looking forward to your ideas, and thank you for all the developers' efforts!

eliandoran:
@youli42 , so how would you suggest to make the cloned notes easier to find?

itsjakobx:
Problem reproduced in a more minimal example.

<img width="1347" height="475" alt="Image" src="https://github.com/user-attachments/assets/df76239d-bce7-4cb4-824b-3430b7da99db" />

## Solution Proposal 1

The point OP makes is that copying the parent results in confusing paths for the children.

This could also be solved by having a relative path view

<img width="1237" height="501" alt="Image" src="https://github.com/user-attachments/assets/dc2dce15-c1f3-48a1-bd0b-e432fc103fec" />

and an absolute path view

[...]

PR #11511, opened 2026-09-15
Clone Paths: Inverse Path Tree take 2

itsjakobx:
Improved Inverse Tree, so that family Trees will actually start to work [^1].

Based on the work and superseeds the work here:

image

Based on request of:  solution 2.0 #7024  and POQ of https://github.com/TriliumNext/Trilium/pull/10329 (superseded by this)
It seems to solve half of #7024 there is need for another solution for 2.1, and the more I think about it, the more both kind of visualizations become needed.

you can get a test family tree at small scale here, that is directly importable:
[Family Tree Test.zip](https://github.com/user-attachments/files/32263338/Family.Tree.Test.zip)

# How to test
1. import family tree through zip and trilums import feature
2. go to Mariana of Austria
3. Press X Paths button on the bottom right.

# What does it include?
Path switching, and an efficient tree view, that only visualizes backwards path decisions. You can easily switch paths, by having a button to switch for each possibility
image

[^1]: https://docs.triliumnotes.org/user-guide/note-types#:~:text=such%20as%20a-,family%20tree,-. Trilium claims to support family trees, but I don't believe that the current capabilities are sufficient to do them justice. Hopefully this feature helps trilium claim important parts of that problem


itsjakobx:
## TODO
- find a better name for the concept than inverse tree and then change the documentation for it

Ready to review.

itsjakobx:
As for possible names I have 3 choices ordered by my preference:

- Placement Tree
- Branch Tree
- Path Tree

PR #11512, opened 2026-09-16
spike: add a Clone map of every placement of the current note (#7024)

itsjakobx:
- this is just a proof of concept for now, generated in collaboration to cursor. I need to handcheck the code, and iterate on the idea

It is the second half of the solution for #7024 It represents its solution 2.1
 
image
image

# How to test
To test out the code this can be used:
[Family Tree Test.zip](https://github.com/user-attachments/files/32266119/Family.Tree.Test.zip)

- open Mariana of Austria, open the clone map from the side bar.


itsjakobx:
Interestingly this solution resembles the graphs found in formal Concept analysis

https://en.wikipedia.org/wiki/Formal_concept_analysis
image

which are Hasse-Diagramms:
https://de.wikipedia.org/wiki/Hasse-Diagramm

image

itsjakobx:
It is a Hasse Diagramm here the formal proof from gpt, I'll hand check it later:
<details>

# Trilium Clones and the Hasse Diagram of a Finite Poset

## Theorem

Let \(S\) be a finite set of Trilium notes.

Assume Trilium's notion of a **clone** identifies notes that represent the same underlying note content through a clone relationship. We consider the clone relationship in the inverse direction, so that the direction of the resulting relation follows the parent/ancestral structure rather than the original clone direction.

Under the assumptions that:

1. the set \(S\) of relevant notes is finite;
2. clone identity is preserved under the inverse relation;
3. the inverse clone relation is acyclic;
4. the inverse relation represents immediate parent relations; and
5. the induced reachability relation is antisymmetric,

the resulting structure is a finite partially ordered set, and its directed graph of immediate relations is isomorphic to the Hasse diagram of that poset.

If, in addition, every pair of elements has a least upper bound, the resulting poset is a finite join-semilattice. If every pair also has a greatest lower bound, it is a finite lattice.

## 1. Trilium's Clone Relation

Let

$$
C\subseteq S\times S
$$

denote the clone relation.

For two notes \(x,y\in S\),

$$
xCy
$$

means that \(x\) and \(y\) are related as clones according to Trilium's clone semantics.

A clone does not constitute an independent conceptual note in the same sense as an unrelated note. Rather, the clone represents another occurrence of the same underlying note.

Consequently, clone identity is naturally treated as an equivalence of note identity/content rather than as an ordering relation itself.

The clone relation therefore gives rise to equivalence classes

$$
[x]=\{y\in S\mid xCy\}.
$$

Let

$$
P=S/{\sim}
$$

be the set of these clone-equivalence classes.

Each element of \(P\) represents one underlying logical note, independent of the particular clone occurrence through which it is reached.

## 2. Inverting the Relation

We now explicitly reverse the direction of the structural relation.

Let

$$
R^{-1} = \{(y,x)\mid (x,y)\in R\}.
$$

Thus, if the original structural relation is interpreted as

$$
x\rightarrow y,
$$

the inverse relation is

$$
y\rightarrow x.
$$

This inversion does not alter the underlying set of notes or their identity. It only reverses the orientation of the relation.

The resulting directed relation is therefore interpreted as an **inverse parent relation**.

We write

$$
x\prec y
$$

when \(x\) is an immediate predecessor of \(y\) under this inverse orientation.

## 3. Construction of the Poset

From the inverse relation \(\prec\), define its reflexive-transitive closure \(\leq\):

$$
x\leq y
$$

if and only if either

$$
x=y
$$

or there exists a finite sequence

$$
x=x_0\prec x_1\prec\cdots\prec x_n=y.
$$

Thus \(\leq\) represents reachability through the inverse clone-parent structure.

We now verify the axioms of a partial order.

### Reflexivity

For every \(x\in P\),

$$
x\leq x
$$

by definition.

Therefore \(\leq\) is reflexive.

### Transitivity

Suppose

$$
x\leq y
$$

and

$$
y\leq z.
$$

Then there exists a finite chain

$$
x\prec\cdots\prec y
$$

and another finite chain

$$
y\prec\cdots\prec z.
$$

Concatenating the two chains gives

$$
x\prec\cdots\prec y\prec\cdots\prec z.
$$

Hence

$$
x\leq z.
$$

Therefore \(\leq\) is transitive.

### Antisymmetry

Assume

$$
x\leq y
$$

and

$$
y\leq x.
$$

If \(x\neq y\), then the two relations together produce a directed cycle in the inverse relation.

By assumption, the inverse clone-parent relation is acyclic.

Therefore such a cycle cannot exist, and consequently

$$
x=y.
$$

Hence

$$
x\leq y\land y\leq x
\implies x=y.
$$

Thus \(\leq\) is antisymmetric.

Therefore,

$$
(P,\leq)
$$

is a partially ordered set.

Since \(S\) is finite, \(P\) is finite.

Hence \(P\) is a **finite poset**.

## 4. The Covering Relation

For elements \(x,y\in P\), define

$$
x\lessdot y
$$

if

$$
x<y
$$

and there exists no \(z\in P\) such that

$$
x<z<y.
$$

The relation \(\lessdot\) is the **covering relation** of the poset.

By assumption, the inverse clone-parent relation represents exactly these immediate relations.

Therefore,

$$
x\rightarrow y
\quad\Longleftrightarrow\quad
x\lessdot y.
$$

No edge represents a merely transitive relationship.

For example, if

$$
x<y<z,
$$

then the Hasse diagram contains

$$
x\rightarrow y
\qquad\text{and}\qquad
y\rightarrow z,
$$

but does not contain the transitive shortcut

$$
x\rightarrow z.
$$

Thus the directed graph defined by the inverse clone-parent relation contains precisely the covering relations of the poset.

## 5. Hasse Diagram

The Hasse diagram \(H(P)\) of a poset \(P\) is the directed graph

$$
H(P)=(P,E)
$$

where

$$
(x,y)\in E
\iff
x\lessdot y.
$$

From the preceding section,

$$
(x,y)\in E
\iff
x\rightarrow y
$$

under the inverse clone-parent relation.

Therefore the identity mapping

$$
\phi:P\rightarrow P
$$

defined by

$$
\phi(x)=x
$$

is a bijection preserving the directed covering relation.

Hence

$$
\boxed{G\cong H(P)}.
$$

The inverse clone structure is therefore representable as the Hasse diagram of the finite poset \(P\).

## 6. Shared Ancestors

Suppose two elements \(a,b\in P\) have a common upper element \(u\):

$$
a\leq u
\qquad\text{and}\qquad
b\leq u.
$$

The two chains therefore converge at the same element \(u\):

$$
a\longrightarrow\cdots\longrightarrow u
$$

and

$$
b\longrightarrow\cdots\longrightarrow u.
$$

Importantly, \(u\) is a single element of \(P\), not two copies of the same element.

Thus the poset naturally represents convergence of distinct chains toward a common upper bound.

This property is a consequence of treating the underlying note identity as a single mathematical element rather than duplicating it according to the number of paths by which it is reached.

## 7. Join-Semilattice

The existence of a common upper bound alone does **not** establish that the structure is a join-semilattice.

For \(P\) to be a join-semilattice, every pair

$$
x,y\in P
$$

must have a **least upper bound**, denoted

$$
x\vee y.
$$

Formally,

$$
x\leq x\vee y
$$

and

$$
y\leq x\vee y,
$$

and for every \(u\in P\),

$$
(x\leq u\land y\leq u)
\implies
x\vee y\leq u.
$$

Therefore, if Trilium's inverse clone structure additionally satisfies this property for every pair of elements, then

$$
(P,\leq)
$$

is a finite join-semilattice.

The corresponding join operation is represented by the least common upper element in the Hasse diagram.

## 8. Lattice

A finite poset is a lattice if every pair of elements has both:

1. a least upper bound, and
2. a greatest lower bound.

The join is written

$$
x\vee y,
$$

and the meet is written

$$
x\wedge y.
$$

Thus, if for every

$$
x,y\in P
$$

there exist elements satisfying

$$
x\vee y
$$

and

$$
x\wedge y,
$$

then

$$
(P,\leq)
$$

is a finite lattice.

Consequently, the lattice property is stronger than merely having converging clone/parent paths.

## 9. Final Result

Starting from Trilium's clone semantics and reversing the structural relation, define the induced reachability relation

$$
x\leq y
$$

by the existence of an inverse clone-parent path from \(x\) to \(y\).

Under the assumptions of finiteness, acyclicity, and antisymmetry, this relation satisfies:

$$
\boxed{
\text{reflexivity}
+
\text{transitivity}
+
\text{antisymmetry}
}
$$

and therefore forms a finite poset.

If the inverse structural relation represents exactly the immediate predecessor/successor relations, then its directed graph is precisely the covering graph of that poset:

$$
\boxed{
G\cong H(P)
}
$$

where \(H(P)\) is the Hasse diagram of \(P\).

If every pair additionally possesses a least upper bound, then

$$
\boxed{
P\text{ is a finite join-semilattice}.
}
$$

If every pair also possesses a greatest lower bound, then

$$
\boxed{
P\text{ is a finite lattice}.
}
$$

Thus, under these purely mathematical assumptions, the inverse of Trilium's clone-based structural relation can be modeled as a finite poset whose immediate relations form its Hasse diagram, with the stronger semilattice or lattice interpretation following precisely when the corresponding universal-bound properties hold.
</details>

Since Formal Concept Analysis uses Hasse-Diagramms, and finds them useful being a rigorous field. We can assume that they will also be useful for people at least those who write notes about concepts, or, write down definitions of words.

itsjakobx:
image

it seems you can you can visualize proof dependency graph using that

However there is a limitations in it, there you would need the paths up from a combination of multiple notes, to see the structure back up from chosen axioms, or in other words, if the visualization would be enhanced just a bit, you could easily visualize all of math.

Moreso, it might be possible to visualize ideologies, religions etc, based on axioms and the things following from it, quite useful in note taking.

Prob needs some more elaboration

itsjakobx:
Make a saved search whose results are the notes you want (Axiom 1, 2, 3, …).
Open that search note.
Connections tab → Clone map.
Any seed / All seeds appears next to the map-type switcher.
Any — every clone path that reaches at least one of those notes.
All — only the shared clone-path meet, plus the paths from there down to the notes.

A dedicated Note Map can do the same with #mapType=clone, #mapCloneCombine=any or all, and #mapRootNoteId= the search’s note ID.

image

so it kinda works, but is still too buggy

PR #11548, opened 2026-09-17
fix(mermaid): skip pan/zoom setup on a diagram without a box (closes #11322)

### LINKED ISSUE #11322: Error with Mermaid ("The matrix is not invertible")
### Description

I have build a Mermaid with Claude Code, then TriliumNotes gets error. Meanwhile the app is not working well.

### TriliumNext Version

0.105.0 and the Nightly build

Windows

### What is your setup?

Local (no sync)

### Operating System Version

Windows 11 Pro 25H2 26200.6584

### Error logs

13:20:20.473 JS Error: Uncaught error: Message: Failed to execute 'inverse' on 'SVGMatrix': The matrix is not invertible., Line: undefined, Column: undefined, Error object: {}, Stack: InvalidStateError: Failed to execute 'inverse' on 'SVGMatrix': The matrix is not invertible.
    at s.zoomAtPoint (trilium-app://app/src/Mermaid-LBWJedJM.js:2:22140)
    at s.zoom (trilium-app://app/src/Mermaid-LBWJedJM.js:2:22317)
    at s.publicZoom (trilium-app://app/src/Mermaid-LBWJedJM.js:2:22469)
    at Object.zoom (trilium-app://app/src/Mermaid-LBWJedJM.js:2:28441)
    at Object.__ (trilium-app://app/src/Mermaid-LBWJedJM.js:2:31758)
    at x (trilium-app://app/src/hooks.module-Cd-XUbeu.js:1:2587)
    at Array.some (<anonymous>)
    at A.__r (trilium-app://app/src/hooks.module-Cd-XUbeu.js:1:3111)
    at trilium-app://app/src/toast-BOS0s9Br.js:2:6024
    at a.__r (trilium-app://app/src/compat.module-DkVTD_DR.js:1:9016)
Stack: Error
    at Object.logError$1 [as logError] (trilium-app://app/src/content_renderer-D73hhHcr.js:2:27363)
    at trilium-app://app/src/glob-B74pb5fV.js:1:37556

maphew:
## Why

Closes #11322

A mermaid note can log "Failed to execute 'inverse' on 'SVGMatrix': The matrix is not invertible", and after that the pan/zoom controls of the preview do not come back. The reported stack (`zoomAtPoint <- zoom <- publicZoom <- effect`) maps to one line: `zoomInstance.zoom(lastPanZoom.current.zoom)` in `useResizer` (`apps/client/src/widgets/type_widgets/helpers/SvgSplitEditor.tsx`).

Root cause: the pan/zoom setup effect in `useResizer` starts svg-pan-zoom on any SVG it finds. svg-pan-zoom 3.6.2 measures the SVG on init and fits the scale as `min(width / viewBox.width, height / viewBox.height)`. A box with zero width or height gives a scale of 0, and a `viewBox` without an area gives NaN or Infinity. The first `zoom()` then calls `getCTM().inverse()` and throws. The throw happens before `zoomRef.current` is set, so each later svg or width change throws again.

The only guard was `width?.width === 0` from `useElementSize`. It does not check the height, the SVG itself or the `viewBox`. It also passes while `width` is `undefined`, which is permanent when the preview pane mounts lazily, because `useElementSize` then attaches no ResizeObserver.

## What

- `useResizer` returns before `svgPanZoom(...)` when the new helper `canFitSvg()` says no. One source file, 16 added lines.
- `canFitSvg()` needs a bounding box with a positive width and height. If a `viewBox` is present, it parses it the way svg-pan-zoom does and needs a finite, positive width and height.
- The early return registers no cleanup and leaves `zoomRef` unset, so the next svg or width change tries again.
- No try/catch, no change to `useElementSize`, no new dependency.

## Validation

- `pnpm --filter client test SvgSplitEditor`: "Test Files 1 passed (1), Tests 10 passed (10)".
- Red, with the line `if (!canFitSvg(svgEl)) return;` removed: "Tests 5 failed | 5 passed (10), Errors 4 errors". The five boxless rows (hidden pane, flat pane, `viewBox` "0 0 0 0", "0 0 NaN NaN", "0 0 100 0") fail with `expected "vi.fn()" to be called +0 times, but got 1 times`, and the fake library throws "The matrix is not invertible." Green with the line restored: 10 passed.
- `pnpm typecheck`: exit 0, "No errors found."
- `pnpm dev:format-check` reports problems only in files of `main` that this change does not touch. No file of this change is in its output.
- Patch coverage from the lcov report: no added source line or branch is uncovered.
- Not run locally: the Playwright mermaid specs in `packages/trilium-e2e`. They assert that `.svg-pan-zoom_viewport` appears on a healthy diagram, so CI is the check that the guard does not reject one.
- Live check: not done. The issue gives no diagram and no steps. I found one path by reading the code and did not run it: open a `#displayMode=source` mermaid note first in a tab, switch to split mode, edit, switch back to source mode, edit again. The console should show no "matrix is not invertible" error.
- Claude review of the exact diff: asked for a guard on the resize effect too, a fake library that throws, and two measurement details. I made the fake throw on a fit scale it cannot use (second commit), so the rows go red for the reported reason. The resize effect needs no guard: it has `width` in its deps like the setup effect, so it runs only after the setup effect has run again, and then an instance exists only if `canFitSvg()` passed in the same flush.
- Cross-vendor review of the exact diff: Codex was not available (`gpt-5.6-sol` is refused for this account and the `gpt-5.6-terra` quota is used up), so I ran NVIDIA Nemotron 3 Ultra through opencode in read-only mode. It reported no findings.

## Scope

- `canFitSvg()` reads `getBoundingClientRect()` only. svg-pan-zoom prefers `clientWidth`/`clientHeight` when both are non-zero. A mismatch can only make the guard skip a diagram, not let a bad one through.
- An SVG without a `viewBox` is trusted. svg-pan-zoom then fits from `getBBox()`. Mermaid output has a `viewBox`, so I left that path as it is.
- When the preview pane mounts lazily, no ResizeObserver is attached, so pan/zoom comes back only on the next edit after the pane becomes visible. This is existing behaviour and not part of this change.
- Which of the boxless states the reporter hit is not known.


eliandoran:
Thanks for the PR.

The fix itself works but this is already the third or the fourth workaround we needed to implement due to the underlying library.

As such, I prefer to get rid of that library entirely. See #11570.

PR #11552, opened 2026-09-17
fix(client): stop keystrokes reaching the note while a dialog opens (closes #7996)

### LINKED ISSUE #7996: Typing quickly after Ctrl+L causes search text to be added to note
### Description

(Amazing program, thanks for all that you do!)

When writing a Text note, there is a short delay between pressing `Ctrl+L` and when the note search bar becomes focused. During this time, the note editor remains focused and input is still possible. This means that if you press `Ctrl+L` and then quickly start typing your query, the first letter or two of your query will be added to the note instead of the search bar.

My expected behavior would be that focus immediately switches to the search bar (even if it's still transparent / being animated), or if that's not possible, then blocking input to the text editor during this transition, so that your search text does not get added to your note. This probably seems like a very rare issue, but I have found myself tripping over it a lot and wondering why there are random letters in front of my note links.

Demo:

https://github.com/user-attachments/assets/26538cb8-997f-4dc6-bd45-4e2abc531c62

Just to make sure this was a real issue and I wasn't accidentally typing out of order, I went frame-by-frame in that recording, and there's a 2 frame gap between when the Add Link transition starts (the main area of the app starts fading darker) and when the `e` appears, which means note input is definitely still possible after `Ctrl+L` is pressed.

### TriliumNext Version

v0.100.0

Other Linux

### What is your setup?

Local + server sync

### Operating System Version

Flatpak on Fedora 42, `Linux 6.17.9-200.fc42.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Nov 24 22:28:05 UTC 2025 x86_64 GNU/Linux`

### Error logs

_No response_

maphew:
## Why

With "Enable transitions and animations" on, press Ctrl+L (add link) or Ctrl+J (jump to note) in a text note and type at once. The typed characters go into the note, not into the dialog.

Root cause: `openDialog()` in `apps/client/src/services/dialog.ts` saves the focused element but leaves it focused. Bootstrap 5.3.8 moves focus only when its focus trap activates, which is after the backdrop fade (.15s) and then the modal fade (.3s). For about 450 ms `document.activeElement` is still the CKEditor editable. With animations off there is no such window, which agrees with the workaround in the issue thread.

Closes #7996

## What

`openDialog()` now moves focus to the dialog's own backdrop immediately after `Modal.show()`. The element that had focus is saved first, so the focus restore on close does not change. Bootstrap moves focus into the dialog when the transition ends, as before.

Focus goes to the backdrop and not to nothing, because `AttributeEditorOverlay` reads `focusout.relatedTarget` (`.modal, .modal-backdrop`) to keep the sidebar attribute editor open while the note type chooser opens. A plain `blur()` gives a null `relatedTarget` and that editor would commit and unmount.

The move is skipped when the dialog is opened with `focus: false` (PopupEditor) and when focus is already inside the dialog. A dialog with `backdrop: false` gets a plain `blur()`.

Result: keys typed during the opening transition are dropped. The reporter named this as an acceptable result.

## Validation

- `pnpm --filter @triliumnext/client test src/services/dialog.spec.ts` with `dialog.ts` from `main`: `2 failed | 34 passed (36)`. The regression tests are "saves the focused element, then moves focus to the backdrop" (it also asserts `relatedTarget` is the backdrop) and "blurs the focused element when the dialog has no backdrop". With the fix: `Tests 36 passed (36)`. The two other new cases (`focus: false`, focus already in the dialog) are guard tests and pass on `main` by design.
- Same command with `--coverage --coverage.reporter=lcov`: all new lines and all branch arms in `dialog.ts` are hit.
- `pnpm --filter @triliumnext/client test src/services/dialog.spec.ts src/widgets/react/Modal.spec.tsx src/services/focus.spec.ts src/widgets/sidebar/AttributeList.spec.tsx`: `Test Files 4 passed (4)`, `Tests 80 passed (80)`.
- `pnpm typecheck`: `No errors found.`
- ESLint not run. I checked the added lines by hand against the code style rules.
- Live A/B check: I ran headless Chromium against a scratch dev server (in-memory integration database, animations on). The script focuses the text editor, presses the shortcut, waits 80 ms, types 4 characters with 40 ms between them, then reads the note. With the fix: 0 characters in the note for Ctrl+L and Ctrl+J, first and second open, and the caret is back in the editor after Esc. With `dialog.ts` from `main`: 4 characters in the note on the first open and 3 on the second open, for both dialogs. A focus event trace showed `focusout` on the editor with `relatedTarget` = `div.modal-backdrop`.
- Claude review of the first commit: found that a plain `blur()` breaks the `relatedTarget` contract of `AttributeEditorOverlay` (major), and that focus on `document.body` lets Space and PageDown scroll the page (minor). The second commit fixes both.
- Cross-vendor review of the exact diff: Codex was not available (`gpt-5.6-sol` is refused for this account and the `gpt-5.6-terra` quota is used up), so I ran NVIDIA Nemotron 3 Ultra through opencode in read-only mode. It reported no defect in the focus move. It raised two points I checked: the inline style on the backdrop cannot leak, because Bootstrap removes the backdrop element on hide and makes a new one for each show; and the last `.modal-backdrop` in the DOM is the one Bootstrap appended synchronously in `show()` just before this code runs. It also asked for two more tests (no backdrop in the DOM, sequential dialogs), which I did not add.
- Still to do by a person: the same manual check in a desktop build.

## Scope

- Type-ahead into the dialog input (the result the reporter preferred) is not included. It needs the modal displayed before the backdrop fade ends, and changes to each dialog's `onShown` (`showRecentNotes` clears the value, `.trigger("select")` selects the typed text).
- Focus moves about 10 to 25 ms after the shortcut, and about 50 ms on the first open of a dialog in a session (`LazyDialog` dynamic import, measured on the dev server). A key pressed inside that time still goes to the editor. I did not measure a production build.
- The backdrop gets `tabindex="-1"` and `outline: none` so that it can hold focus without a visible ring. It is not in the tab order.


eliandoran (approved):

eliandoran:
@maphew , @greptile-apps,

FYI, this PR caused a significant regression in which the LLM "Add provider" screen would be completely unusable (clicking on any provider did nothing).

PR #11596, opened 2026-09-20
fix(search): keep the full-text word written against an attribute prefix

PloneMraz:
## What happens

A full-text term written straight against `#` or `~`, with no space, disappears from the query.

```
lex("towers #book")   fulltext ["towers"]   expression ["#book"]
lex("towers#book")    fulltext []           expression ["#book"]
lex("two towers#book") fulltext ["two"]     expression ["#book"]
```

End to end, on two notes labelled `#book`, one of them titled *Two Towers*:

```
search("towers #book")  ->  ["Two Towers"]
search("towers#book")   ->  ["Random Book", "Two Towers"]
```

The query silently widens. Nothing is logged and nothing is highlighted as invalid — the term the user typed just stops narrowing anything, which is worse than an error, because the result list still looks plausible.

## Why

In `lex.ts`, the branch that handles the prefix ends the full-text part but does not finish the word being accumulated:

```ts
if (chr === "#" || chr === "~") {
    if (!fulltextEnded) {
        fulltextEnded = true;      // the pending word is never emitted
    } else {
        finishWord(i - 1);
    }

    currentWord = chr;             // ...and is overwritten here
```

With a space, `chr === " "` calls `finishWord` one character earlier, which is why the spaced form has always worked.

## A case the first version of this got wrong

Emitting the pending word unconditionally broke parenthesised queries. A
leading `(` is not a term, so `(#a)` started searching for `(`:

| query | `main` | first version | now |
|---|---|---|---|
| `(#a)` | Alpha | **nothing** | Alpha |
| `(#a OR #b)` | Alpha, Beta | **nothing** | Alpha, Beta |

The whole suite stayed green through that, because every parenthesised
case in `parse.spec.ts` and `docs_examples.spec.ts` either feeds `parse()`
pre-built tokens or puts something before the `(`. The guard and a
regression test for it are in this branch.

Separately, `(#a)` sets `searchContext.error` to `")" is not a valid
expression...` on `main` today while still returning the right notes: the
opening paren is dropped before `fulltextEnded` is set, leaving its `)`
unmatched. This PR does not change that, in either direction — happy to
fix it in a follow-up if you'd like it.

## The change

`finishWord(i - 1)` now runs on both paths, before `fulltextEnded` is set, so the pending word is routed to `fulltextTokens` exactly as the spaced form routes it. The second branch is unchanged in behaviour: once the expression part has started, `finishWord` still pushes to `expressionTokens`.

Documentation checked first — `Search.md` describes full-text terms combined with attribute filters and does not require a separator, and `docs_examples.spec.ts` only ever writes the spaced form, so nothing documented depended on the old behaviour.

## Tests

- `lex.spec.ts` — the adjacent form for both `#` and `~`, that only the last word is affected (`two towers#book`), that `fulltextQuery` is set, and that the spaced form still lexes as before.
- `search.spec.ts` — the end-to-end case above, which is what a user actually sees.

## Verification

Tested on `main` at `c7684bf218`, Windows, Node 24.14.0. Both suites that
carry `trilium-core` specs, as CLAUDE.md asks:

| suite | base | with this change |
|---|---|---|
| `pnpm --filter server test` | 5699 passed, 0 failed | **5702 passed, 0 failed** |
| `pnpm --filter standalone test` | 4482 passed, 1 failed | **4485 passed, 1 failed** |

The one standalone failure is `backup-stream.spec.ts > streams the live database as its exact bytes`, which times out at 5000ms under the full 285-worker run on this machine and passes on its own. It fails identically on untouched `main`, so it is not related to this change.

Mutation check: reverting the lines in `lex.ts` with an in-file edit fails exactly the added tests and nothing else — both the word case and the parenthesis guard.

On formatting: `eslint -c eslint.format.config.mjs` reports 270 errors across these three files before this change and 270 after, so the added lines introduce none.

Written with the help of an AI assistant; every claim above was verified by running the code.


PloneMraz:
Fixed — the comments now state the current invariant instead of narrating the defect, per the comment rule in CLAUDE.md. The before/after stays in the commit body and the PR description.

Reading that file also caught something I had missed: core specs run under both the server and standalone suites, and I had only run the first. Both are now in the description. The standalone run has one failure, `backup-stream.spec.ts > streams the live database as its exact bytes`, which times out at 5000ms under the full parallel run on my machine and passes in isolation — it fails identically on untouched `main`, so it is not from this change.

PloneMraz:
Correction — please re-review. While looking for another bug I found that the first version of this broke parenthesised queries.

A leading `(` is not a search term, but emitting the pending word unconditionally made `(#a)` search for `(`:

| query | `main` | first version | now |
|---|---|---|---|
| `(#a)` | Alpha | nothing | Alpha |
| `(#a OR #b)` | Alpha, Beta | nothing | Alpha, Beta |

The full suite stayed green through it: every parenthesised case in `parse.spec.ts` and `docs_examples.spec.ts` either feeds `parse()` pre-built tokens or puts something before the `(`, so none of them exercise the lexer path. The branch now guards that case and carries a regression test for it, at both the lexer and the search level.

Both suites re-run: server 5702 passed / 0 failed, standalone 4485 passed with the same pre-existing `backup-stream.spec.ts` timeout that fails on untouched `main`.

Noted in the description as well: on `main` today `(#a)` sets `searchContext.error` while still returning the right notes, because the opening paren is dropped before `fulltextEnded` is set and its `)` is left unmatched. This PR leaves that untouched — I can send a follow-up if you want it fixed.

